PR #57, opened 2015-06-17
Reorganized and expanded HTML CCR.

Zone22:
Reorganized and expanded HTML CCR. Spoken forms and action objects need
to be vetted for format and usability.

Edit: Looks like formatting will screwed up by uploading to Github.

### DIFF
--- a/caster/bin/data/ccr/confightml.txt
+++ b/caster/bin/data/ccr/confightml.txt
@@ -1,6 +1,5 @@
-#
 # This is a heavily modified version of Dragonfly's _multiedit.py command-module
-#  meant to enable HTML  syntax with CCR
+# Meant to enable HTML syntax with CCR
 
 from dragonfly import *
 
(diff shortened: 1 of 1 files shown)


synkarius:
Thanks much, this is awesome. Just as a heads up, there is a hard limit to how many CCR commands can be in a grammar before a BadGrammar error occurs from Dragon getting overloaded. That number is around 550; it varies depending on the complexity of the commands.
http://www.knowbrainer.com/forums/forum/messageview.cfm?catid=25&threadid=20234&enterthread=y

For that reason (and for recognition accuracy), I would encourage you to move as many commands as you can (in other words, as many of them as you think you won't typically speak in sequence) into the non-CCR section of the CCR module. (See configpython.txt for an example, and/or the YouTube video.) This makes them enabled with the others, but not chainable with them.

Zone22:
Learn something new everyday!  I wonder if it differs between versions DNS versions?
I will work on moving them into the non-CCR section of the CCR module.

PR #62, opened 2015-06-27
Revised HTML CCR formatting and spacing.

Zone22:
Once documentation is completed for NodeRule the HTML CCR will be
rewritten to include attributes and non-CCR commands.


Zone22:
@synkarius
I'm having some issues with github messing with tab spacing when uploading a file. At first I tried matching my development environment to githubs default 8x tab spacing. That did not work, then I tried to match Windows github client preview tab spacing before syncing. I thought it was strange that Windows github client and the file on github itself differed in tab spacing. Ultimately I ended up comparing the file on github to the original in my editor to mangle the right tab spacing which was a pain in the butt.

There's got to be an easier way and I'm completely confused why my editor and github are matched with tab spacing that it isn't display correctly when uploaded. Any thoughts?

synkarius:
@Zone22 I have no idea why that happens. I wouldn't worry about it too much, as it's really just a display issue.

synkarius:
Also, if you rewrite HTML as a NodeRule, you don't have to worry about what I said before about moving stuff into the non-CCR section. [NodeRule](https://github.com/synkarius/caster/wiki/noderule) optimizes space in its own way.

Issue #86, opened 2015-08-11
Formatting and other dictation commands in ccr sequences


chilimangoes:
An issue that I seem to run into quite a bit when using caster, is that I'll want to use a formatting command that contains one or more words from the ccr command list. For example, yesterday afternoon, I was editing an HTML file and I had the HTML vocabulary enabled. One of the class names used on the page was "inlineModalDialog", but when I said "gerrish gum bow in-line modal dialog", what I got was "inlineModal\<dialog\>".

This doesn't happen all the time, but it's often enough to be really irritating, especially when searching through caster's source code to learn how it works, what the built-in commands are, and how I can extend it. Half the time, the word I'm searching for is a CCR command, so I end up having to resort to using the keyboard.

I was able to work around this problem by moving the text formatting commands in confignavigation.txt from the CCR section to the non-CCR section. However, there are times when it would be convenient to be able to mix text formatting commands in with CCR commands, and this got me thinking about the strategy proposed on this blog post:

http://handsfreecoding.org/?p=100

Basically, it involves having a separate category of commands that can only be invoked at the end of an utterance. This way, you can still invoke a series of CCR commands, followed by a command that includes dictation, but the dictation command must come last in the sequence.

This is something that I could probably work on, but I wanted to get your thoughts on the approach, and maybe brainstorm a little bit first.

synkarius:
I too am constantly annoyed by this problem. James's solution is extremely interesting; thank you for pointing it out. Actually, I like James's solution so much that I have implemented a version of it in the new develop branch, so that I can use it at work tomorrow. Please do not consider this to be an end to the conversation. It may be adequate, it may not; my attempt is basically my first thoughts on the subject. 

The way I have it implemented here, any active CCR command can be used as the original or terminal command. This removes the need to read in and store the original and terminal commands separately. Do you think this approach works, or do we want the original and terminal commands to be a limited subset of the overall CCR command set?

synkarius:
So for your example, now you would say, "terminal gerrish gum bow in-line modal dialog".

chilimangoes:
Wow, that was quick! I'm still trying to wrap my head around how all the pieces of caster work, especially the CCR portion.

[...]

PR #87, opened 2015-08-14
Add "set format" command.

chilimangoes:
This PR adds an explicit "set format" command. As discussed in the comments on #86, this allows you to separate the selection of a format from the command used to output formatted text, reducing the number of things you need to say for a single utterance.

Usage example: "set format laws spine bow" followed by "format some identifier".

### DIFF
--- a/caster/bin/data/ccr/confignavigation.txt
+++ b/caster/bin/data/ccr/confignavigation.txt
@@ -48,9 +48,10 @@ cmd.map = {
 	"Kraken":						R(Key("c-space"), rspec="Kraken", rdescript="Control Space"),
 	     
     # text formatting
+    "set format (<capitalization> <spacing> | <capitalization> | <spacing>) (bow|bowel)":  R(Function(navigation.set_text_format), rdescript="Text Format"), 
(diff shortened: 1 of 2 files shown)


chilimangoes:
I'm not partial to using the keyphrase "set format". In fact, I kind of like the idea of using a shorter keyword/keyphrase, like maybe "setti".

synkarius:
I wanted this in develop rather than master, so I tried checking out your branch and merging into develop, and pushing that, but then Github didn't count you as a contributor. So I reverted that, accepted to the pull request to master, reverted my revert, and merged master back into develop. It still isn't counting you as a contributor, and I'm not sure why except that maybe it has to do with your email that you registered on Github with. [See the question mark here.](https://github.com/synkarius/caster/commit/b1d5d0d886a15d457b45f207a9ecb1e3734bb726)

At any rate, my git skills need some sharpening methinks.

synkarius:
Also, thank you, this works great and is very useful.

chilimangoes:
You're welcome. I know what you mean about sharpening your git skills. I use git a lot locally for personal stuff, but working with remote repos, forking, and pull requests still trips me up. :/

I'm not sure how my email would be a problem since I just used my yahoo email to sign up on github. In any case, don't worry about it and thanks for trying. If it keeps happening, I'll look into submitting a support ticket.

PR #96, opened 2015-08-28
Javascript CCR: more consistent code block formatting.

chilimangoes:
A few of the javascript commands were a missing space between () and {} in the output. Neither Visual Studio or Sublime text correct this automatically, which means that some code blocks end up with inconsistent formatting, which I then have to correct manually.


synkarius:
Unfortunately, a lot of the CCR modules are like this. I designed them for Eclipse when Caster was just my personal macro set, and thus had much smaller aims. The Rust module is the first one I made with general usability in mind, and I think a handful of the others could be made more generic with smarter design. For instance, the "brace pan" command in the Rust module would eliminate some of the wonkiness in the Java module.

PR #97, opened 2015-08-28
Javascript CCR: Add "anon funk" command

chilimangoes:
JavaScript makes heavy use of anonymous functions passed as parameters to other functions as callbacks and event handlers. The current "function" command in the JavaScript CCR contains a NAME and a semicolon at the end, both of which then need to be deleted. This PR adds an "anon funk" command, which produces the following output, suitable for embedding in a parameter list:

```
function () {
    |<-- cursor here
}
```

### DIFF
--- a/caster/bin/data/ccr/configjavascript.txt
+++ b/caster/bin/data/ccr/configjavascript.txt
@@ -40,6 +40,9 @@ cmd.map = {
 		"function":					R(Text("function NAME(){};")+Key("left:2, enter")
 									 +SelectiveAction(Key("enter, up"), ["AptanaStudio3.exe"]), 
 									 rdescript="Javascript: Function"),
+		"anon funk":				R(Text("function () {}")+Key("left:1, enter")
+									 +SelectiveAction(Key("enter, up"), ["AptanaStudio3.exe"]), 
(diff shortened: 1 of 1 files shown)


synkarius:
This is useful. And well-named. :)

One note about the "NAME" and "VALUE" etc. stuff that you see in a bunch of places in the CCR modules: they're  meant to be used with the "fill" command from the navigation module. Their design is taken from VoiceCoder's code templates. (The original VoiceCode**r**, not voicecode.io.)

So for instance, with the "function" command, you'd do something like: "function fill name terminal format sort" to get:

    function sort () {}
                 | <--- cursor

I know this template style isn't for everyone, and I think multiple ways to approach the task like the way you've implemented here are an appropriate solution. Once I get that command editor up and running (which won't be easy and will take a while, but still...), the user will be able to just uncheck / edit the commands they don't want.

synkarius:
Uh, and whereas your `|` cursor mark indicates literally where you meant the cursor should be, mine meant "on the line above this", haha. Whoops.

chilimangoes:
I had originally named it "lambda" but several other languages actually make a distinction between anonymous functions and lambdas, so I renamed it.

Good to know about the fill command, I hadn't seen that yet.

Issue #93, opened 2015-08-27
Multi-monitor support for grids


chilimangoes:
Does anyone else use multiple monitors?

Currently, it doesn't seem like any of the grid commands (e.g. Legion, Rainbow) work on anything other than the primary monitor. How difficult would it be to modify them to display on a different monitor? I was thinking something like "Rainbow 2" or "Rainbow 3" to display the rainbow grid on monitor two or three, respectively.

Zone22:
I have two monitors and could help with testing.

chilimangoes:
From [this script](https://github.com/dictation-toolbox/dragonfly-scripts/blob/7c31b3980dca68989869d56539d8de9eca439bc0/_winctrl.py), it appears that dragonfly has some built-in support for multiple monitors. I think all we would need to do is [enumerate dragonfly.monitors](https://github.com/dictation-toolbox/dragonfly-scripts/blob/7c31b3980dca68989869d56539d8de9eca439bc0/_winctrl.py#L103-L104) and then [get the rectangle of the selected monitor](https://github.com/dictation-toolbox/dragonfly-scripts/blob/7c31b3980dca68989869d56539d8de9eca439bc0/_winctrl.py#L304-L305).

I'm looking at grids.py and I see that the constructor for TkTransparent accepts a `dimensions` parameter, which is then exposed by the concrete grid classes, such as RainbowGrid and DouglasGrid as `grid_size`. I believe that setting this parameter to the rectangle of the selected monitor is what we need to do, but I need to do some more digging to figure out how to do that.

chilimangoes:
I implemented multi-monitor support for the Rainbow and Douglas grids last night. Still need to tackle the Legion grid.

@synkarius Could you take a look at [this commit](https://github.com/chilimangoes/caster/commit/d4c919464d219a232f8fbbdeab3e6590ecbf2582) and let me know what you think of this approach?

synkarius:
Looks pretty thorough to me. My only request would have been to make the monitor number optional in the spec, which you did.

Actually I'm a little surprised that you got away with importing something from Dragonfly in an external script. A lot of Dragonfly modules import other modules which import other modules which cause Dragonfly to re-initialize (and crash), but it seems `monitors` is safe.

synkarius:
Rereading this, I realize you asked what I think of the approach. I like it.

chilimangoes:
[...]

PR #103, opened 2015-09-09
Multi-monitor support for grid commands

chilimangoes:
This PR adds support for multiple monitors when using any of the caster mouse grid commands. Usage is simple: simply append the (optional) monitor number to the grid command you want to use. If you don't specify a monitor number, the grid will display on the primary monitor (monitor 1).

For example, saying "rainbow 2" will display the rainbow grid on monitor 2. To display the rainbow grid on the primary monitor, you can say either "rainbow 1" or simply "rainbow".


chilimangoes:
References #93

synkarius:
The code is really clean, and it works. Thanks much.

PR #108, opened 2015-10-14
App grammars for Visual Studio and KDiff3.

chilimangoes:


synkarius:
I'm a bit confused. I thought Visual Studio was [this](https://github.com/synkarius/caster/blob/develop/caster/apps/msvc.py). I don't use it, so I don't know. If they're two different programs, I'll approve this. If I just named the linked file incorrectly, let's just rename and use that one and put any relevant changes in there.

chilimangoes:
They are different, yes. MSVC is MS Visual C++. It has often been referred to synonymously as Visual Studio, especially in the pre-.NET days, because it was included in a developer bundle named Visual Studio (Visual Basic, Visual C++, Visual Fox Pro, etc). But then MS created a new flagship IDE with the actual name of Visual Studio (that was really confusing for a while).

synkarius:
You learn something every day.

PR #109, opened 2015-10-14
Allow "pipe (sim | symbol)" for a more phonetically distinct option

chilimangoes:
More often than not, DNS hears "type seven" when I say "pipe sim". "Pipe symbol" is more phonetically distinct, and easy enough to remember and say, so I thought it might be good to give users that option out of the box. Feel free to reject the PR if you think this falls into the category of things I should address via filter functions.


synkarius:
These sorts of issues are the sort which should be handled via filter functions now yes, but I've had trouble with this one too. Of course, that doesn't justify the change either, but I guess until further opinions weigh in, I'm willing to say go for it. 

What should become the standard default set is a question which I'm way too lazy to worry about, especially at this stage. Ultimately, I think it'd be nice to fully adopt someone else's standard (I have Ben Meyer's explicit permission to copy his VoiceCode specs), or get Nils Klarlund in here to analyze the "core" CCR sets and the programming standard.

PR #124, opened 2015-11-04
Grammars for Foxit Reader, SSMS, and misc additions

chilimangoes:
* Added application grammars for Foxit Reader and SQL Server Management Studio.
* Added "address bar" command to Chrome.
* Added "context menu" command to nav.py.
* Added misc window movement and switching commands to nav.py.


caster/lib/ccr/core/nav.py
```diff
@@ -51,6 +53,14 @@ class NavigationNon(MappingRule):
 
         "refresh":                          R(Key("c-r"), rdescript="Refresh"),
         
+        "maxiwin":                          R(Key("w-up"), rdescript="Maximize Window"),
+        "move window":                      R(Key("a-space, r, a-space, m"), rdescript="Move Window"),
+        "window (left | lease) [<n>]":      R(Key("w-left"), rdescript="Window Left") * Repeat(extra="n"),
+        "window (right | ross) [<n>]":      R(Key("w-right"), rdescript="Window Right") * Repeat(extra="n"),
+        "monitor (left | lease) [<n>]":     R(Key("sw-left"), rdescript="Monitor Left") * Repeat(extra="n"),
+        "monitor (right | ross) [<n>]":     R(Key("sw-right"), rdescript="Monitor Right") * Repeat(extra="n"),
+        "switch (window | windows)":        R(Key("ca-tab"), rdescript="Switch Window") * Repeat(extra="n"),
```
synkarius: This command is what tie match / ex match are trying to accomplish. Oi. I wish I'd known about it. I've always just used alt-tab. :-/

chilimangoes:
Heh, yeah, same here. I just learned about it a couple days ago while searching for Windows keyboard shortcuts that might be useful as caster commands. I also just recently discovered, or rediscovered rather, the "list windows" command in DNS. Between that and this "switch windows" command, I'm finding app switching to be a bit less tedious now.

Issue #49, opened 2015-05-17
make fuzzy symbol searching smarter


synkarius:
https://github.com/synkarius/caster/blob/master/caster/lib/pita/selector.py#L11

My original shot at a fuzzy string matching algorithm sucks. It needs to be better.

synkarius:
NEEDED: test data to try out different algorithms and variations on;

Please submit symbols from any codebase you like. If you could format as follows, that'd be great too:
[spoken word(s), actual answer, choice1, choice2, choice3, etc.]

Examples:
["issue maturity date", "ismtdt", "32d22as", "i", "date", "ismat"]
["certificate cancel amount", "ccnamt", "ccd", "cancel", "certs"]
["stock ticker", "stktck", "stock", "stTick", "sstkks", "tick123"]

chilimangoes:
Per #122, added logging to the FuzzyMatchAction to facilitate data collection in the above format. Logging is disabled by default and can be opted in by setting ["pita"]["enable_logging"] to true in `settings.json`.

chilimangoes:
I just remembered I have an an algorithm I wrote a couple years ago to match YouTube video titles to theTVDB episode titles. The use case is a little different, but it might work for this. I'll dig it up and port it to python so we can test it.

synkarius:
Excellent. The more we try the better. It could turn out that some algorithms fit some codebases better than others, though in the (few admittedly limited) tests I've done so far, difflib has (unsurprisingly) done consistently better than my homemade matching algorithm.

chilimangoes:
Here are some brainstorming notes on how we might take a machine learning approach to this. We could use a recurrent neural net (RNN) to map strings to vectors and then use those vectors to determine similarity, perhaps by calculating euclidean distance between vectors or feeding the vectors into a second model that outputs a similarity score. Although not the same thing as what we're doing, I think there are some similarities to training [neural language and translation models](https://github.com/kjw0612/awesome-rnn/blob/master/README.md#machine-translation) to map words in different languages with a similar meaning to similar locations in a vector space.

One potential benefit of this, from a performance standpoint, is the process of mapping symbols to vectors could be done as part of the initial scanning process and stored in a hash table for quick retrieval. I don't know whether this would make a big difference, since we would still need to compute the vector for the spoken phrase at runtime, but it might help.

A relatively small amount of training data would be a problem, so here are some ideas for how we might be able to work around that limitation:

[...]

PR #128, opened 2015-11-08
Ported a similarity calculation algorithm

chilimangoes:
To potentially help with #49.

I ported a similarity calculation function I created to compare youtube video titles to theTVDB episode titles. For that purpose, it returned the correct episode better than 9 out of 10 times, but I don't know how well it will work for matching abbreviations. For example, if I try to match "test function", the symbol "test" has a higher score (39) than "tstfnc" (37) because of the token sequence boosting I added.

One thought I had was to change the sequence boosting to work with non-contiguous sequences. In this case, the entire sequence of "tstfnc" would count as a sequence, even though most of the letters aren't contiguous, because those letters appear in the same sequence in both.

### DIFF
--- a/caster/lib/pita/selector.py
+++ b/caster/lib/pita/selector.py
@@ -247,9 +247,82 @@ def sift4(s1, s2, max_offset, max_distance):
     
     
     
+####################################################################################
+####################################################################################
     
+def calculate_similarity(input_a, input_b):
     
+    # tokenize input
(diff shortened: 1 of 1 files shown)


synkarius:
I think your proposed change to the sequence boosting is a good idea. The present PITA algorithm does something similar, also to detect non-contiguous sequences. Another thing to try might be removing all the vowels.

PR #130, opened 2015-11-11
Mouse improvements, app grammar updates, next/prior window

chilimangoes:
* Added Windows Explorer grammar and updated Visual Studio grammar.

* Replaced the use of the Playback action in the right and left click commands with direct calls to the underlying user32 api calls to improve performance. The existing commands rely on the Dragon engine to perform mouse clicks, and this often causes a lag of a few hundred milliseconds for me. Sometimes it would also cause double-click to fail altogether.

* Added "next/prior window" command. "Next window" and "prior window" both do the same thing and are simply a shortcut for saying "switch window" followed by "shock". Convenient for repeatedly switching back and forth between two windows.

* Added mouse wheel scrolling, with the following caveat. For some windows, scrolling works regardless of whether the mouse pointer is over the scrollable area. These will simply scroll the text control or other scrollable area that has the focus. For others, in particular most IDEs that I've tried, only work when the mouse pointer is over the scrollable area that you want to scroll. I have ideas for working around this limitation in some apps, but I figured in the meantime, partially working mouse wheel support is better than none at all.

### DIFF
--- a/caster/lib/navigation.py
+++ b/caster/lib/navigation.py
@@ -290,11 +290,13 @@ def kill_grids_and_wait():
 
 def kick():
     kill_grids_and_wait()
-    Playback([(["mouse", "left", "click"], 0.0)]).execute()
+    windll.user32.mouse_event(0x00000002, 0, 0, 0, 0)
+    windll.user32.mouse_event(0x00000004, 0, 0, 0, 0)
 
 def kick_right():
     kill_grids_and_wait()
-    Playback([(["mouse", "right", "click"], 0.0)]).execute()
+    windll.user32.mouse_event(0x00000008, 0, 0, 0, 0)
+    windll.user32.mouse_event(0x00000010, 0, 0, 0, 0)
 
 def kick_middle():
     kill_grids_and_wait()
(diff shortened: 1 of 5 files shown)


caster/lib/navigation.py
```diff
@@ -290,17 +290,26 @@ def kill_grids_and_wait():
...
 def kick():
     kill_grids_and_wait()
-    Playback([(["mouse", "left", "click"], 0.0)]).execute()
+    windll.user32.mouse_event(0x00000002, 0, 0, 0, 0)
+    windll.user32.mouse_event(0x00000004, 0, 0, 0, 0)
 
 def kick_right():
     kill_grids_and_wait()
-    Playback([(["mouse", "right", "click"], 0.0)]).execute()
+    windll.user32.mouse_event(0x00000008, 0, 0, 0, 0)
```
synkarius: Hmm. Good call about this depending on Dragon. Though I hate WSR, I still want it to be as usable as possible.

chilimangoes:
I think Playback may work in WSR. For me it was more an issue of performance. Dragon occasionally lags by a couple hundred milliseconds on each Playback action, so a double click ends up registering as two single clicks.

PR #132, opened 2015-11-20
Fixes #116 - Multimonitor support for Legion grid

### LINKED ISSUE #116: Legion grid uses incorrect screen capture on secondary monitors
The Legion command always sends a screen capture the primary monitor to the TiRG DLL, regardless of what monitor your invoking the grid on (e.g. saying "Legion 2").

chilimangoes:
Fixes #116

### DIFF
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -14,7 +14,7 @@ try: # Style C -- may be imported into Caster, or externally
         sys.path.append(BASE_PATH)
 finally:
     from caster.asynch.mouse.grids import TkTransparent, Dimensions
-    from caster.lib import settings, utilities
+    from caster.lib import gdi, settings, utilities
 
 try:
     from PIL import ImageGrab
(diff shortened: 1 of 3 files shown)


chilimangoes:
@synkarius I created a new module in lib called `gdi.py` with a screen grab function. Let me know if you would prefer I move this somewhere else.

synkarius:
`/lib` seems like an appropriate place for it. 

I noticed you're converting a byte array into a PIL Image so it'll take with the LegionScanner. Did you try not converting it, and passing the gdi.py byte array directly to TiRG?

synkarius:
This change also fixes a bug that I had been meaning to write up about Legion: I have my taskbar on the left side of my screen rather than the bottom. That seemed to offset Legion by the width of the taskbar, so that the scanned boxes would show up slightly off to the side where they should be. They show up correctly now. Thanks!

chilimangoes:
Hah, glad to hear that.

It didn't even occur to me to try passing the byte array directly to TiRG. My goal was fixing it as quickly as possible since not having Legion on my other monitors at work was a real downer. I'll give it a shot, thanks!

PR #138, opened 2016-04-02
Adding support to detect dragon 13 in settings.py

mostlyjason:
This detects dragon 13 in settings and allows the reboot command to work

### DIFF
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -145,6 +145,14 @@ def init_default_values():
     if not SETTINGS["paths"]["WXPYTHON_PATH"] in sys.path and os.path.isdir(SETTINGS["paths"]["WXPYTHON_PATH"]):
         sys.path.append(SETTINGS["paths"]["WXPYTHON_PATH"])
     
+# detect dragon version 12 and 13
+if not os.path.isfile(SETTINGS["paths"]["ENGINE_PATH"]):
+    dragon13Path = "C:/Program Files (x86)/Nuance/NaturallySpeaking13/Program/natspeak.exe"
+    if os.path.isfile(dragon13Path):
+        SETTINGS["paths"]["ENGINE_PATH"] = dragon13Path
+    else:
+        print "Cannot find default dragon engine path"
+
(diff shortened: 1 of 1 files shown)


chilimangoes:
Just curious, where does this setting get used? I'm curious because I use DNS13 at work and DNS14 on one of my home computers and haven't run into any issues with this, but I'm on my phone right now so I can't search the code or look at my install path.

mostlyjason:
It gets passed to the castor/bin/reboot.bat script when saying the "reboot dragon" command

synkarius:
Since you closed the issue, I take it you realized that that section is really only meant to generate the settings.json file for the first time, not to _be_ the settings object. The user is intended to update the relevant parts of the settings file. That said... I guess we should default it to the Dragon 14 install path, eh? Probably no one installing 12 much anymore. @chilimangoes What is the 14 default full path?

mostlyjason:
Yeah I was just trying to make it smart about picking the default so new users can get going right away regardless of what version they use. If you tell me the path for 14 I can add it in and resubmit.

chilimangoes:
I'll check when I get home later today. I never actually use this command because the times when I would want to use it are almost always because Dragon is frozen, which kinda defeats the purpose. :)

mostlyjason:
Is there another way to reload the caster files? It seemed like it reloaded my user directory when I turned off and on the microphone. However, it didn't seem like it was reloading the MacroSystem directory where I installed caster.

synkarius:
@mostlyjason Oh. Hadn't thought of that. Good idea. May as well detect 13 in addition to 14 then, eh? But no, here's no other way to fully reload Caster. That's my most common use for that command actually. Reload, make sure changes worked as intended.

chilimangoes:
So, unsurprisingly, this is the path where DNS14 is installed on my home computer:

`C:/Program Files (x86)/Nuance/NaturallySpeaking13/Program/natspeak.exe`

chilimangoes:
Gah! Replace 13 with 14.

PR #139, opened 2016-04-02
Adding a setting for how long to wait for Atom palette to load

mostlyjason:
I made this improvement based on our discussion in #137. I'm defaulting to 30 since that should work for a wide variety of users. Advanced users can always lower this if they have a faster computer. Does that seem like a good approach?


synkarius:
That seems fine. Now that you mention it, a higher default _would _ be better than a faster lower default that doesn't always work.

PR #147, opened 2016-05-31
Update mouse documentation

seekM:
I worked through the [Alternate Mouse Movement Modes video](https://www.youtube.com/watch?v=UISjQBMmQ-I) and noticed that the [documentation](http://caster.readthedocs.io/en/latest/caster/doc/readthedocs/Mouse/) is missing the colors for the rainbow command, so I thought it would be nice adding them. I also changed the structure of the documentation page - please feel free to reject this pull request if you don't like it (e.g. the documentation got longer). The mouse commands were missing too, however, _curse up left 30_ did not work for me. To make it work I had to replace _curse_ by _mouse_ and use only one direction: _mouse up 30_. If this is an issue only I have, please feel free to edit this or let me know.

### DIFF
--- a/caster/doc/readthedocs/Mouse.MD
+++ b/caster/doc/readthedocs/Mouse.MD
@@ -1,16 +1,56 @@
 # Alternate Mouse Modes
-
 Demonstration [here](https://youtu.be/UISjQBMmQ-I).
 
-## Rainbow
-Creates a grid on the screen that is colored. You can select a square by calling out the color followed by the number of the tile. For example, "red 86". The colors loop when filling the screen, so you can select the second row of colors by saying "two red 86".
+## Basic mouse commands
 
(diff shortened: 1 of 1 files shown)


synkarius:
I think your documentation changes are good. Thank you. :)

Unfortunately, the video is, as you noticed, outdated. The `curse` command didn't work because you didn't use the new (more phonetically distinct) direction keywords.
`sauce [E]` = up
`dunce [E]` = down
`lease [E]` = left
`ross [E]` = right
The `[E]` there is an optional long "E" sound which can make the unvoiced "S" sound easier to speak, especially in combination with other commands. For example,
`curse lease-e-sauce ten`
is easier than
`curse lease sauce ten`
in which the S's from `lease`'s end and `sauce`'s beginning blend together.

PR #150, opened 2016-06-07
Fix mouse documentation (GitHub Flavored Markdown)

seekM:
When writing the mouse documentation for the last pull request (#147) I used the online tool http://dillinger.io/. Now I noticed that Github uses a differen syntax and therefore does not display the documentation correctly. It should work now.


synkarius:
Ah. I thought you intended for it to look that way, but this looks better.

Issue #148, opened 2016-06-07
Integer Remap causes crash for non-English DNS users


synkarius:

synkarius:
In the settings file `caster/bin/data/settings.json`, find the setting called `integer_remap_crash_fix` and set it to `true`. This should resolve the problem.

seekM:
I can confirm that it does not crash anymore.

seekM:
I just noticed that DNS does not crash if - when starting DNS - I first select my english profile and then switch to my non-englisch profile. If I directly select my non-english profile when starting DNS, it still crashes.

synkarius:
Thank you for confirming that it works now.

That's strange about the language selection order. I don't know what to make of that.

seekM:
As stated in #149, putting `"de": "dragonfly.language.de"` into [`language_map`](https://github.com/t4ngo/dragonfly/blob/94c69ad90f11e21ce2d8d208c85b9ff51ca0eb0e/dragonfly/language/loader.py#L55) avoids crashing if the german profile is loaded. Now language selection order does not matter anymore, too.

synkarius:
Thank you. I've added that to [my Dragonfly repo](https://github.com/synkarius/dragonfly/commit/dd744270e0904943550fd739d8e3f293cd8efe2a) and submitted a pull request to @t4ngo's repo.

PR #151, opened 2016-06-08
Remove fix for #148, fix number issue for german profile

seekM:
See [this comment](https://github.com/synkarius/caster/issues/149#issuecomment-224550003).

### DIFF
--- a/caster/apps/firefox.py
+++ b/caster/apps/firefox.py
@@ -23,7 +23,7 @@ from caster.lib import control
 
 
 class CommandRule(MergeRule):
-    pronunciation = "fire foxp"
+    pronunciation = "fire fox"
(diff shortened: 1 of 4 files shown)


synkarius:
I am going to redo the "Integer Remap" section such that it is language-agnostic. The correct change to fix this issue properly is the one @chilimangoes suggested and you implemented in Dragonfly.

Versatilus:
I'm having a hard time, based on flow of the conversation, figuring out whether or not this still needs addressed. If it's no longer a concern, I would like to close the issue.

Versatilus:
At this point, the code is too far out of sync and the configuration issues can be re-examined at a later date.

PR #160, opened 2016-08-23
Fix "no handlers could be found for logger" error messages

chilimangoes:
Minor update to fix the occasional "no handlers could be found for logger" messages that occasionally appear in the NatLink status window. Now if an error bubbles up to one of the Dragonfly Action classes, it should print the exception and stack trace information, which we can use for debugging.


synkarius:
Nice. That's always irked me.

PR #162, opened 2016-09-09
context fixes

Versatilus:
I added the terminal I use to access git to the list of app contexts where the git rules are active. I also tracked down an issue I was just having over the weekend where the context was getting activated when I tried to connect to Wi-Fi through the system tray. The new start menu in Windows 10 is provided through `ShellExperienceHost.exe`, which was causing a false match against `sh`.

I also added the cruft that Emacs likes to leave everywhere to `.gitignore`. If you would like me to resubmit it separately, please let me know.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@ test.py
 log.txt
 complexity_report*
 *.log
+*.*\#
+*.*~
+
 
 # Ignore Natlink
 /core/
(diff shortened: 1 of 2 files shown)


synkarius:
You don't have to submit the `.gitignore` stuff separately. The change is documented well enough through this pull request.

PR #164, opened 2016-09-19
updates to Chrome keybinds

Versatilus:
I sorted through the keybinds for Chrome, removing unnecessary or redundant ones, regrouping related ones, and adding some that weren't there. I also cleaned up the imports.

I tested most of the keybinds. A few weren't triggering their intended functions, but that may be an issue with the newly updated browser or some extension I'm running. I left them in under the assumption that they would either start working again as things got sorted out or they would work for others where they haven't for me. At the very least, the utterances should be unique enough that they shouldn't cause any false hits.

I have more I'll be submitting as a separate file at a later date which expect certain extensions to be installed and work for specific sites like Facebook, Gmail and Google Docs.

### DIFF
--- a/caster/apps/chrome.py
+++ b/caster/apps/chrome.py
@@ -9,9 +9,7 @@ Command-module for Chrome and Firefox
 """
 #---------------------------------------------------------------------------
 
-from dragonfly import (Grammar, AppContext, MappingRule,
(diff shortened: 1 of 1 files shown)


synkarius:
The specific-sites rule is a good idea, looking forward to it.

PR #178, opened 2017-03-06
History fixes

carywalk:
1. Say I had created a sequence of commands using the phonetic alphabet (e.g. "Press control Charlie"). The sequence failed to save  because of the slash  character associated with the phonetic pronunciation. This was because the code was attempting to remove the slashes. However, the type was an immutable tuple and could not be assigned to. I fixed it by changing it to a list.

2. Once created, the sequence did not load properly because the name was read in as Unicode characters from the JSON file. A downstream dependency was not expecting Unicode and failed. Fixed by immediately converting it to an ASCII string when read.


carywalk:
Great idea. I'll probably be able to take a look at it this weekend.

carywalk:
I added the requested modification.

PR #181, opened 2017-08-11
Modularize js

vivshaw:
To help prevent grammar overcomplexity, pull React & Flow into their own grammars that can be enabled & disabled separately. This way we can activate only the rules we need, when we need them.


vivshaw:
Ack! This was a mistake, please ignore all this! I was trying to do this to my own fork, not to your repo

PR #191, opened 2017-11-26
Rework and clean up setting.py

betaorbust:
This PR:
 - Removes some unused default settings values
 - Refactors settings.py with
   - More docstrings
   - More PEP-8 formatting
   - Fewer global mutations
   - Separates public and private methods
   - Moves to merging a default nested dict against the live settings. This allows separating out the defaults in the future.

On my machine, I can boot with a vanilla generated settings file and create one from scratch as needed.
The diff of generated settings files is here: https://www.diffchecker.com/KT7nL3P0

### DIFF
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -1,10 +1,10 @@
 import json
 import os
 import sys
+import collections
 
-SETTINGS = None
-BAD_LOAD = False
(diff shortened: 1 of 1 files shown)


synkarius:
This is very clean. Thank you, @betaorbust.

PR #201, opened 2018-03-12
Partial format

gerrish:
I have added a partial format command which allows the user to utter any dictation but keep the first n (1 up to 4) words. The rest of the dictation is used as context to improve the accuracy of the first words. Example usage:"single format end of time"=>"end", which minimizes the chances of Dragon actually outputting "and" while at the same time not polluting the command pool with new commands.

### DIFF
--- a/caster/lib/ccr/core/nav.py
+++ b/caster/lib/ccr/core/nav.py
@@ -167,7 +167,7 @@ class Navigation(MergeRule):
         Choice("semi",
                     {"dock": ";", "doc": ";", "sink": ""
                     }),
-        Choice("word_limit",{"single": 1,"double": 2,"triple":3}),
+        Choice("word_limit",{"single": 1,"double": 2,"triple":3,"Quadra":4}),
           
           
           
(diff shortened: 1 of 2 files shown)


Versatilus:
I'll try to make the extra effort to review this before the end of the day. From a quick glance, I don't expect any problems. I'm just trying to get out of the bad habit of merging without testing. :-)

For future reference, we do all feature enhancements to the `develop` branch first. I went ahead and changed this pull request for you.

gerrish:
Thank you I just realized and was going to fix it myself. Just a quick question however, do you have any suggestion for a testing environment where I can test the changes without messing with my main caster installation, because I have made a few other changes to my own installation and for other more complex modifications or additions it won't be so easy to drop them in and test them.

Versatilus:
Whenever I work on something new, I make a new local branch in my git
repository. I have several half implemented features that I hide away when
I'm working on the main repository.

Versatilus:
Based on the command name, I expected it to behave the same as the other `format` command, but it doesn't respect preset formatting. Was this intentional?

Versatilus:
This is awesome! Thanks!

If you wanted to figure out where to document it, it would be great if you could add a note about the command's use. I've been wanting to rebuild the quick reference with MarkDown, but I haven't gotten around to it.

PR #204, opened 2018-03-14
Prolog and vhdl rules

gerrish:
Set of rules for vhdl and prolog
The prolog set is very basic because it is very easy to type by voice. I Plan to later expand upon these both.


LexiconCode:
Thanks!

gerrish:
I seem to be doing something wrong because sublime and notepad plus plus show Proper formatting but on Git it ends up totally messed up.

PR #208, opened 2018-03-18
Unicode fixes

Versatilus:
This attempts to fix some long-standing issues with the way Caster interacts with the clipboard and reorganizes file I/O to be forwards compatible with Python 3. The files were reformatted through [YAPF](https://github.com/google/yapf) using the attached configuration file. I'm open to different styles and I will try to keep the style changes to a single commit in the future.
[style.txt](https://github.com/synkarius/caster/files/1823011/style.txt)

This patch changes Caster to prefer Unicode when interacting with the Windows clipboard. No attempts are made to explicitly handle legacy text localization. My hope is that Windows will do the right thing when it handles auto conversion. To support this, file I/O has been rewritten to assume and encode for UTF-8 in a way that is compatible with Python 3.

**I need help testing this.** The change was seamless for me, but all my configuration files were already 100% ASCII. **Backup your configuration files before testing this.**


LexiconCode:
All right I'll start testing throughout the week.

PR #213, opened 2018-03-23
Support for Gitter App

LexiconCode:
Brings voice commands to the Gitter App.
I plan to add a little more intelligence to some of the commands.

### DIFF
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -100,6 +100,7 @@ _DEFAULT_SETTINGS = {
         "flashdevelop": True,
         "foxitreader": True,
         "gitbash": True,
+        "gitter": True,
         "kdiff3": True,
         "douglas": True,
         "legion": True,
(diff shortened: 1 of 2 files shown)


caster/lib/settings.py
```diff
@@ -100,6 +100,7 @@ def _find_natspeak():
         "flashdevelop": True,
         "foxitreader": True,
         "gitbash": True,
+        "gitter": True,
```
Versatilus: This isn't really a problem for you. This is just sort of a reminder for myself. I need to make it one of my top priorities to change the configuration system so that core files aren't changed whenever a minor configuration item is added.

LexiconCode:
I appreciate the merge. I would hold off merging WIP tags pull requests until confirmed complete. At least that's what I had in mind with the WIP tags.

Versatilus:
Fair enough. The changes seemed fairly harmless. I'll look for those labels
harder next time. :-)

Issue #214, opened 2018-03-24
Roadmapish


Versatilus:
This isn't an official roadmap. I'm just using this space as a scratchpad for ideas.

*  [ ] There should be commands to store to/stage from our special clipboard history. There are many times when I'm using some program which copies strings directly to the clipboard and I would like to be able to save them for later. I also occasionally use programs which use nonstandard key bindings for copy and paste and it would be equally nice to be able to set up a paste operation from a stored string.
   I'm currently thinking the command phrases should be `store stoosh #` and `stage spark #`, but I'm open to suggestions.

*  [ ] One of my top priorities is to make Caster more robust/fault-tolerant, especially during startup. It shouldn't fail to start unless one of the core modules fails. An incorrect CCR or app grammar shouldn't cause the whole system to crumble. I think this might be most important during module imports. This is a priority to me in no small part because of how many times I've broken everything by merging some "trivial" change without testing it first.

*  [ ]  That brings us to our next area I would like to work on: unit tests. I'm flying kind of blind here. I'm not sure where to begin or what will be needed to mock up so that we can test just the Caster code.

*  [ ] Somewhat related to to the robustness as listed above, I believe the settings/configuration could and should be separated better from the core program. I envision the core system restoring absent configuration, but it should otherwise be hidden away behind the /user directory.

LexiconCode:
Could you give me an example to copy and paste and explain a bit more what you mean by nonstandard key bindings?

> ...copies strings directly to the clipboard and I would like to be able to save them for later. I also occasionally use programs which use nonstandard key bindings for copy and paste and it would be equally nice to be able to set up a paste operation from a stored string....

Versatilus:
Caster's 'multi clipboard' is very convenient for moving around multiple
text fragments. The problem is also one of the things that makes them so
easy to use: they automatically use the Windows shortcut keys to move text
between the application and the clipboard.

There are many times when applications put information into the Windows
clipboard which I would like to save for later. One such example is the
file manager I use, FreeCommander. It allows you to copy the path of the
current file or folder to the clipboard using a menu or special shortcut. I
would like to be able to say `store stoosh five` to put the information in
slot 5 of the Caster clipboard.

Similarly, applications like `bash` don't use the standard Windows
shortcuts for copy and paste. Obviously, custom commands for these
applications are the 'correct' long-term solution, but the added
flexibility of moving text between Caster and the Windows clipboard without
executing the standard shortcuts would be nice.

[...]

PR #216, opened 2018-03-28
Add Import Tolerance

gerrish:
For the applications, the dynamic importation allows for error checking the application modules. While this is not a perfect solution, since runtime errors from invoking the commands will not stop the crashes, for any debugging purposes this makes testing changes much much easier.
For the ccr, I just added try except blocks because I couldn't figure out a similar solution due to the fact that class names are imported. I am not sure how I could import classes and their names without doing something like setting  the globals dictionary which is not a very good idea. I am mostly making this pull request so that I can get some feedback on how to make this solution better.


LexiconCode:
Relates to the 2nd goal of #214

LexiconCode:
@Versatilus would have insight on implementation.

gerrish:
For extra clarification, @Versatilus, would It be ok if I mess with the globals dictionary to emulate `from package import classname` statements?

LexiconCode:
Feel free to experiment as needed.

Versatilus:
This definitely looks like the right direction to me. Ideally, this process should be entirely dynamic, which might require refactoring some of the modules for consistent names. I'm still thinking about the whole process. In the interest of future portability, I think you might want to look into `importlib` instead of using `__import__` directly. `imp` might also be a possibility, but I expect it to be deprecated by the time we get the ability to run under Python 3.

caster/lib/ccr/core/nav.py
```diff
@@ -107,7 +107,7 @@ class Navigation(MergeRule):
                                                                lambda fnparams: UntilCancelled(Mimic(*filter(lambda s: s != "periodic", fnparams)), 1).execute(), \
                                                                use_spoken=True))]),
     # VoiceCoder-inspired -- these should be done at the IDE level
-    "fill <target>":                R(Key("escape, escape, end"), show=False) +
+    "fill <target>":                R(Key("escape, escape"), show=False) +
```
Versatilus: I'm curious what this change is about.

gerrish:
I'm having issues with getting importlib to find the modules. Using globals() with __import__ worked but no matter what package or module name I try with importlib, it won't find them.

caster/lib/ccr/core/nav.py
gerrish: An experiment which l forgot to remove. I have restored the original files in the next commits.

gerrish: An experiment which l forgot to remove. I have restored the original files in the next commits.

Versatilus:
I just played around with this a bit and I figured it out. The first argument to `import_module` is relative to `sys.path` so
```python
__temp__ = importlib.import_module(".lib.ccr." + module_name, package="caster")
module = getattr(__temp__, class_name)
```
is just as valid as
```python
__temp__ = importlib.import_module("caster.lib.ccr." + module_name)
module = getattr(__temp__, class_name)
```
I have some ideas I can elaborate on in a few hours.

gerrish:
So any ideas? The current implementation works pretty well and will be portable to python 3 with importlib's __import__ though.

caster/lib/ccr/__init__.py
```diff
@@ -1,16 +1,32 @@
...
+    "haxe.haxe": ("Haxe", ),
+    "html.html": ("HTML", ),
+    "java.java": ("Java", ),
+    "javascript.javascript": ("Javascript", ),
+    "python.python": ("Python", ),
+    "rust.rust": ("Rust", ),
+    "sql.sql": ("SQL", ),
+    "prolog.prolog": ("Prolog", ),
+    "vhdl.vhdl": ("VHDL", ),
+}
```
Versatilus: I would really like to make this list automatically. I just haven't settled on the implementation strategy.

caster/lib/ccr/__init__.py
```diff
@@ -1,16 +1,32 @@
...
+
+
+for module_name,class_name_tup in command_sets.iteritems():
+    for class_name in class_name_tup:
+        try:
+            module = __import__(module_name, globals(), locals(),[class_name])    #attempts to import the class
```
Versatilus: This works well, so I'll stop being so picky.

gerrish:
We could make the class names special or add some attribute to the classes so that we can automatically recognise if they should be imported. If this gets accepted, I will make another pull request with support for filters as well.

gerrish:
I added dynamic imports for filters and rules as well. The code is virtually the same as the applications.

Versatilus:
I'm still thinking about ways to make CCR imports fully dynamic.

gerrish:
To get the modules, we could just check for subdirectories of ccr and find all the Python modules inside. Since those modules will potentially contain classes which should not be imported and given global names, we could try one of the approaches below:
1.We could create a list in every ccr grammar which would contain the names of the classes which must be imported.
2.An attribute could be added to the `MergeRule` class which will declare it should be imported.
3.We could crosscheck the rules which have been added to the merger to know what should be imported.

LexiconCode:
I would open up a Issue since this pull request has been merged.

Issue #194, opened 2018-01-16
For people having issues with the grid system (Legion, Douglas, Rainbow)


LazoCoder:
I was having issues with the Legion, Rainbow and Douglas commands and installing Microsoft Visual C++ Redistributable Packages for Visual Studio 2013 (x86) did not resolve them. The issue was that even though the commands were registering, nothing was happening.

I have two Python versions installed on my computer (Windows 10), Python 2.7 (in C:\Python27) and Python 3.6 (in C:\Python36). The latest version of Python is used by default. I suspect that this is due to the way that my environment variables have been set up during the installation of the newer version. So if I go to command prompt and type ```python -V``` it outputs: ```Python 3.6.2```.

I wanted to see the error for the grid functionality so I navigated into ```C:\Natlink\Natlink\MacroSystem\caster\asynch\mouse``` and I ran the command ```python grids.py``` only to recieve the error ```No module named 'SimpleXMLRPCServer'```. Since this was using the default python instance, it was using Python 3.6. So I tried using Python 2.7 via the command ```C:\Python27\python grids.py -g d``` and it worked. I discovered that Caster was using the default Python (in this my case Python 3.6) instead of Python 2.7.

**TEMPORARY FIX**:
Go into ```C:\Natlink\Natlink\MacroSystem\caster\lib\navigation.py``` and on lines 48, 50 and 52 where it says ```Popen(["pythonw", ...``` and change ```"pythonw"``` to the location of your Python 2.7 instance. In my case: ```Popen(["c:/python27/pythonw", ...```

**PERMANENT FIX**:
Somebody should find the location of the Python interpreter that is running the script and use pythonw from that particular instance (assuming the script is run in Python 2.7). Or find a way to run Python 2.7 in some other way, but in either case, not to assume that a user's environment variables are using Python 2.7 by default.

The final thing that I would like to point out is that I wish the grids would come to the foreground by default because sometimes they don't and so I have to say "kick ... 7 kick" in order to first select it and bring it to the foreground (this is time-consuming) as opposed to just saying "7 kick". But then sometimes it does spawn in the foreground.

LazoCoder:
On another note I would like to thank @synkarius and all the other contributors for creating and maintaining this project. It's made a world of difference for me and has pretty much saved my programming career. God bless you all.

LexiconCode:
[...]

Issue #126, opened 2015-11-06
Legion breakdown


synkarius:
@Zone22 asked for a way to make Legion more precise. Changing the TiRG algorithm was shot down, but it _would_ be fairly easy to break the rectangles Legion returns into sub-rectangles every X% or N pixels. 

Credit alexander15w for the idea.

LazoCoder:
Solved in #220

PR #220, opened 2018-03-30
Legion Enhancement, Cheat sheet update, and more

LazoCoder:
- Fixed Issue #194 
  - Users can now have multiple versions of python installed and not encounter issues with grid systems & asynch tasks (specifically if their environment variables point to a different python version).
- Significantly enhanced legion's ability to find text with no noticeable difference in speed
  - I experienced issues with the prior implementation of legion where faded out font was ignored and also certain font and background color combinations would throw off the algorithm. This is no longer a problem.
legion
- Added "bench" and "squat" which perform "left mouse button down" and "left mouse button up" functions respectively.
  - This is for users who use mouse equipment such as the [SmartNav](https://www.naturalpoint.com/smartnav/) and desire the ability to drag.
  - I chose the words "bench" and "squat" because they are simple, crisp, single syllable words that are related to each other by definition. They are also related to "kick" since they are all athletic movements (easier to remember).
  - I've been using "bench" and "squat" now for a few weeks and haven't had any issues with dragon confusing it with any other words (I'm North American and don't speak with any discernible accent as far as I am aware).
- Updated the Cheat Sheet
  - (3) corrected “at sign” to “atty”
  - (4) wally
  - (7) bench, squat
  - (9) window snapping

Tested and working fine on my bootcamp installation of Windows 10 on my 15" Mac Book Pro (2015), using Dragon v14.


chilimangoes:
Nice work indeed! The one observation I would make is that, at least for me, "squat" would be more intuitive as the mouse down command since I think of squatting down. But it's also not a big deal if you prefer to keep it the way it is.

LazoCoder:
@chilimangoes Sure, that makes more sense. I'll change it.
@gerrish Can you be more specific? Is it [this](https://github.com/LazoCoder/caster/blob/b0356009fd701f49529f29b6ac7756701730d404/caster/asynch/mouse/legion.py#L133) line you want changed? What is the error that you get?

LazoCoder:
I've switched the functionality of bench and squat. Additionally, I've successfully broken down legions rectangles in *another branch* (as per #126). Here is what it looks like:
split_1
The question is what to do with legion's "light" command. How often do you guys use this? I can remove it, or leave it in, but keep in mind if I leave it in it will now highlight a smaller region. I haven't used it enough to gauge the usefulness of this so any pointers are appreciated.

Alternatively I can leave the original unbroken down rectangle but have it in a different color. This could be used for a greater highlighting area. See the following image to understand what I mean. I made this in Microsoft paint, the three blue rectangles can be used if the user wants to highlight more content than what the smaller rectangles would allow. The number would be on either the right or left side. So for example the user could say "111 light" and it would select 107, 108, 109 and 110 (the whole area).
split_2
But maybe this will look clustered, ugly or confusing for new users? If the highlight command is seldom used then it may not be worth it to add this complexity to both the code and the screen. I haven't implemented this part yet so let me know what you think before I do.

chilimangoes:
I've rarely the "light" command. Personally, I would vote for leaving it how it is, with the option of maybe modifying it later to do something like "light one hundred and forty through one hundred and forty four." That shouldn't be hard to implement and would give us the best of both worlds.

Versatilus:
@gerrish This is mostly my fault. I'm guessing that in this case the library loader requires a byte string for the path because of ctypes issues. This should be easily fixed by encoding the Unicode string to a byte string based on the file system locale.

LazoCoder:
Okay so I went ahead and I added in the rectangle division as demonstrated in my previous comment. I left the highlight function as it is.

Furthermore I had an issue where every time I said "again x times" a console would pop up and start printing things to the screen and preventing the function from occurring. This has been fixed.

I won't be adding anything more to this branch so you guys can test it. If there are any issues that you want me to fix let me know.

LazoCoder:
I ran a recursive command to convert all the files to yapf as per the .style.yapf file.
> C:/Python27/scripts/yapf -i -r --style .style.yapf -vv *

Versatilus:
I love this!

Issue #172, opened 2017-02-05
Legion grid is sometimes inaccurate


kendonB:
I think this is because I switch between two monitors. Is there a way to reset the parameters that go into legion?

![legion](https://cloud.githubusercontent.com/assets/5327505/22630734/e416ada2-ebb3-11e6-8393-5b027570be0c.png)

chilimangoes:
Man, this is weird. @imalloutofbubblegum [reported the same issue on Gitter](https://gitter.im/synkarius/caster?at=5850b462aeb490080487baf0), but for me it works fine on multiple monitors both at home and work. There used to be [an issue](https://github.com/synkarius/caster/issues/116) where it would always use a screenshot of the primary monitor for the Legion command, regardless of which monitor you displayed the grid on, but that should have been fixed by [this commit](https://github.com/synkarius/caster/commit/8eb74eb43fbd05d742cb19789f37c32ea4b7382b).

Some questions and things to look at that might help track down the issue:

1) What version of Windows are you running?
2) Is this sporadic or can you reproduce it reliably? Does it always happen on one monitor but not the other (e.g. only on your secondary monitor)?
3) When you look at the code you're running, do you see the changes from the fix noted above?
4) Are you comfortable enough with Python to try testing the `grab_screen` method with different `bbox` parameters to see if what is captured by the method actually corresponds to what is being requested of it? If not, I should be able to work up some code for you to run, but it'll probably take me a couple days.

kendonB:
1) I'm running Windows 10
2) It isn't sporadic exactly, but I can't reproduce it either. If it's happening, it seems to always happen until I restart the computer. Even then it can happen instantly as well.
3) I can see all the changes in that commit, yes (I downloaded caster in the last couple of weeks).
4) I worked it out. I tried these and they seemed to be correct:
```

>>> grab_screen([0, 0, 400, 400]).show()
>>>
>>> grab_screen([0, 0, 2000, 2000]).show()
```

chilimangoes:
Hmmm, curiouser and curiouser. It sounds like I might need to add some debugging code in there to output some additional info to the NatLink window to give us an idea of exactly where things are getting mixed up.

Regarding your answers to #2:

2a) Have you noticed it happening only on your secondary monitor, or does it happen on your primary as well? And if it happens to both, does it happen to both simultaneously, or always just one or the other.

2b) Does restarting Dragon work (maybe try restarting Dragon a couple times) or do you actually have to restart your computer?

[...]

PR #223, opened 2018-04-01
Legion DPI fix

gerrish:
Fix for issue #172,dpi settings properly accounted for.

### DIFF
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -205,6 +205,17 @@ def main(argv):
     monitor = 1
     dimensions = None
     auto_quit = False
+
+    errorCode = windll.shcore.SetProcessDpiAwareness(2)  #enable 1-1 pixel mapping
+    dpi_map = {96: 1.0, 120: 1.25, 144: 1.5, 192: 2.0}
+
+    LOGPIXELSX = 88
+    dc = windll.user32.GetDC(0)
+    horizontal_dpi = windll.gdi32.GetDeviceCaps(
+        dc, LOGPIXELSX)  # Horizontal DPI determines the DPI scaling value
+    windll.user32.ReleaseDC(0, dc)
+    dpi_scaling = dpi_map[horizontal_dpi]
+
     try:
         opts, args = getopt.getopt(argv, "ht:a:d:m:",
                                    ["tirg=", "dimensions=", "autoquit="])
@@ -228,7 +239,8 @@ def main(argv):
 
         if dimensions is None:
             r = monitors[int(monitor) - 1].rectangle
-            dimensions = Dimensions(int(r.dx), int(r.dy), int(r.x), int(r.y))
+            dimensions = Dimensions(
+                int(r.dx*dpi_scaling), int(r.dy*dpi_scaling), int(r.x), int(r.y))
         lg = LegionGrid(grid_size=dimensions, tirg=tirg, auto_quit=auto_quit)
     except Exception:
         utilities.simple_log(True)
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -227,7 +227,7 @@ def main(argv):
             elif opt == '-m':
                 monitor = arg
             elif opt in ("-d", "--dimensions"):
-                # wxh+x+ySave
+                # wxh+x+y
                 dimensions = Dimensions(*[int(n) for n in arg.split("_")])
             elif opt in ("-a", "--autoquit"):
                 auto_quit = arg in ("1", "t")
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -26,7 +26,6 @@ The width of each Legion box cannot exceed the width of the column.
 This way relative partitioning is achieved since larger resolutions 
 will be partitioned into columns of larger width.
 '''
-VERTICAL_PARTITIONS = 30
 
 
 class Rectangle:
@@ -41,7 +40,8 @@ class LegionGrid(TkTransparent):
         self.setup_xmlrpc_server()
         TkTransparent.__init__(self, settings.LEGION_TITLE, grid_size)
         self.attributes("-alpha", 0.7)
-        self.max_rectangle_width = int(grid_size.width/VERTICAL_PARTITIONS)
+        self.max_rectangle_width = int(
+            grid_size.width/settings.SETTINGS["miscellaneous"]["legion_vertical_columns"])
         self.tirg_positions = {}
         if tirg is not None:
             self.process_rectangles(tirg)
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -216,7 +216,7 @@ def main(argv):
     horizontal_dpi = windll.gdi32.GetDeviceCaps(
         dc, LOGPIXELSX)  # Horizontal DPI determines the DPI scaling value
     windll.user32.ReleaseDC(0, dc)
-    dpi_scaling = dpi_scaling/96.0
+    dpi_scaling = horizontal_dpi/96.0
 
     try:
         opts, args = getopt.getopt(argv, "ht:a:d:m:",
@@ -242,7 +242,8 @@ def main(argv):
         if dimensions is None:
             r = monitors[int(monitor) - 1].rectangle
             dimensions = Dimensions(
-                int(r.dx*dpi_scaling), int(r.dy*dpi_scaling), int(r.x), int(r.y))
+                int(r.dx*dpi_scaling), int(r.dy*dpi_scaling), int(r.x*dpi_scaling),
+                int(r.y*dpi_scaling))
         lg = LegionGrid(grid_size=dimensions, tirg=tirg, auto_quit=auto_quit)
     except Exception:
         utilities.simple_log(True)
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -206,7 +206,9 @@ def main(argv):
     dimensions = None
     auto_quit = False
 
-    errorCode = windll.shcore.SetProcessDpiAwareness(2)  #enable 1-1 pixel mapping
+    error_code = windll.shcore.SetProcessDpiAwareness(2)  #enable 1-1 pixel mapping
+    if error_code == -2147024891:
+        raise OSError("Failed to get system dpi settings")
     dpi_map = {96: 1.0, 120: 1.25, 144: 1.5, 192: 2.0}
 
     LOGPIXELSX = 88
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -4,8 +4,9 @@ import re
 import sys
 import threading
 from ctypes import *
+from win32api import EnumDisplayMonitors
 
-from dragonfly import monitors
+#from dragonfly import monitors
 
 try:  # Style C -- may be imported into Caster, or externally
     BASE_PATH = os.path.realpath(__file__).split("\\caster")[0].replace("\\", "/")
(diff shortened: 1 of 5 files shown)


Versatilus:
I don't have the hardware to properly test this (but it looks good on the surface), so I can't help with testing it.

gerrish:
I also do not have multiple monitors however this works for my single Monitor with 150% scaling. You can try adjusting the display scaling slider yourself to see if this works. The boxes should be properly placed.

caster/asynch/mouse/legion.py
```diff
@@ -205,6 +205,19 @@ def main(argv):
...
+    if error_code == -2147024891:
+        raise OSError("Failed to set app awareness")
+    dpi_map = {96: 1.0, 120: 1.25, 144: 1.5, 192: 2.0}
+
+    LOGPIXELSX = 88
+    dc = windll.user32.GetDC(0)
+    horizontal_dpi = windll.gdi32.GetDeviceCaps(
+        dc, LOGPIXELSX)  # Horizontal DPI determines the DPI scaling value
+    windll.user32.ReleaseDC(0, dc)
+    dpi_scaling = dpi_map[horizontal_dpi]
```
Versatilus: I was just thinking about ways I could test this and that made me wonder if there is any software that could fake it on my current hardware. In order to keep the simulated resolution high enough to be usable it wouldn't necessarily  scale at the same rates horizontally and vertically. It could also have weird DPI values.

What if a piece of hardware doesn't have square pixels? Also, wouldn't it be easier/faster to just divide the given DPI value by 96.0?

chilimangoes:
I have multiple monitors and will give this a try at work tomorrow. What I really wonder is whether this works on really high DPI screens, like they have on some laptops. I don't have anything like that to test with.

LexiconCode:
I have access to a 4K screen and will test tomorrow afternoon

LexiconCode:
I speak the `Legion` command and nothing happens. It looks like Legiongrid is not executed

`Traceback (most recent call last):
  File "c:\NatLink\NatLink\MacroSystem\caster\asynch\mouse\legion.py", line 252, in <module>
    main(sys.argv[1:])
  File "c:\NatLink\NatLink\MacroSystem\caster\asynch\mouse\legion.py", line 219, in main
    dpi_scaling = dpi_scaling/96.0
UnboundLocalError: local variable 'dpi_scaling' referenced before assignment
PS C:\NatLink\NatLink\MacroSystem>`

gerrish:
There's a typo from autocomplete in the last change. I won't be home for about an hour to change it. I have added a comment in the code if you want to test it before I can change it.

LexiconCode:
legion at 250% scale
3840 x 2160 screen resolution
Duplicated display

chilimangoes:
It looks like the problem is exacerbated on multiple monitors. At work I have a 1280x1024 and a pair of 1920x1080 monitors. If I set them all to 150%, and say "legion 3" (the rightmost monitor, labeled 2 in the screenshot), the legion grid is both too small and overlaps with monitor 1 in the middle. The good news is the legion rectangles are in the right places over the text. It looks like it's just a problem with the calculation of the screen capture rectangle and subsequent positioning of the Legion Grid window. Almost makes me wonder if we need to calculate anything or if just setting DPI awareness would be enough.

legion_dpi

gerrish:
@LexiconCode try signing in and out, windows fails to update the dpi value without a reboot or sign out. If it works I will try to find a more reliable way to get the settings.
@chilimangoes Just setting doing awareness results in proper positioning but a smaller rectangle for me.

gerrish:
I added the scaling to the x and y values as well. I think this Could fix @chilimangoes' problem of the screen overlapping. Do try to sign out and back in,after setting the scaling however  else windows reports the previous dpi settings rendering the solution useless.

LexiconCode:
screenshot_1

Strange screenshots I've tried signing out and rebooting. Happens with snip tool and other programs.

gerrish:
Is it just the screenshots or is it still completely whack?

LexiconCode:
Just the screenshots

chilimangoes:
That completely fixed the problem as long as all three monitors have the same DPI scaling. However, when I set monitor 1 to 100% scaling and the other two monitors to 125%, the "Legion" command works fine on monitor one, but looks like this on the other two monitors. My guess is the problem is that the DC you're grabbing (NULL) is for the entire desktop. In order for the scaling to work properly, I imagine you'll need to get the DC for each monitor and calculate them separately.

legion_dpi

gerrish:
@LexiconCode 
capture
My screenshot with 250% scaling. I'm not sure what is going on with that, not even sure if the issue is with Windows or the overlay itself.
@chilimangoes I suspected that this would happen. I have found an elegant function to deal with this. I will implement it tomorrow.

gerrish:
By using the Windows API there seems to be no need for calculating DPI settings. The updated code gets monitored rectangles with the proper dimensions. Something which I cannot test is whether the API returns monitors in their proper order. If it does, this should have no issues with multiple monitor setups as well.

chilimangoes:
I just tested this and it seems to work perfectly for two of my monitors, but then nothing displays at all for the monitor that is to the left of my primary monitor. This looks like a regression of a bug I remember from when I first introduced multi-monitor support. It had something to do with the dimensions of the monitor being reported as negative or something when the monitor is to the left of or above the primary monitor. I'll take a look back tonight and try to find what I did to fix that.

chilimangoes:
Okay, I found the problem. On the following lines, when getting the window dimensions to pass into the Legion grid:

```
monitor_choice = EnumDisplayMonitors(None, None)[int(monitor) - 1]
(x, y, dx, dy) = monitor_choice[2]
dimensions = Dimensions(int(dx), int(dy), int(x), int(y))
```

The Dimensions class is expecting the first two parameters to be the height and width of the window, but the `dx` and `dy` values returned by `EnumDisplayMonitors` are actually the X and Y coordinates of the lower right-hand corner of the selected monitor. For the primary monitor, these are equivalent to height and width. But on any other monitor, you need to calculate the height and width like so:

```
dimensions = Dimensions(int(dx - x), int(dy - y), int(x), int(y))
```

After making the above change, Legion works correctly for me with every monitor and DPI configuration that I tried. As you noted, it now also works without needing to log out and log back in.

Thanks for the work you put into this PR! I think we can close #172 unless anyone else has a test case that this fails for.

gerrish:
Thank you I will update it!

LexiconCode:
I'll see how it works on the 4k screen later today. I want to echo my thanks as well :)

LexiconCode:
An interesting error message the Legion overlay was blank when this occurred. Keep in mind this is a laptop connected to a 4K screen. My laptop has an integrated and dedicated GPU.  I'm inclined to believe that is causing the issue with an underlying cause with Windows.

```
ERROR:action:Exception from function send_input:
  File "C:\Python27\lib\site-packages\dragonfly-0.6.6b1-py2.7.egg\dragonfly\actions\action_function.py", line 109, in _execute
    self._function(**arguments)
  File "C:\NatLink\NatLink\MacroSystem\caster\apps\gridlegion.py", line 34, in send_input
    s.go(str(n))
  File "C:\Python27\Lib\xmlrpclib.py", line 1243, in __call__
    return self.__send(self.__name, args)
  File "C:\Python27\Lib\xmlrpclib.py", line 1602, in __request
    verbose=self.__verbose
  File "C:\Python27\Lib\xmlrpclib.py", line 1283, in request
    return self.single_request(host, handler, request_body, verbose)
  File "C:\Python27\Lib\xmlrpclib.py", line 1316, in single_request
    return self.parse_response(response)
  File "C:\Python27\Lib\xmlrpclib.py", line 1493, in parse_response
    return u.close()
  File "C:\Python27\Lib\xmlrpclib.py", line 800, in close
    raise Fault(**self._stack[0])
Fault: <Fault 1: "<type 'exceptions.KeyError'>:'1'">
ERROR:action.exec:Execution failed: Function(send_input): <Fault 1: "<type 'exceptions.KeyError'>:'1'">

```

LexiconCode:
screenshot_1

chilimangoes:
@LexiconCode I'm not sure that error has anything to do with your drivers or screen. The error occurred in the send_input processing code in gridlegion.py, which happens after you select an option, eg saying "15 kick" or similar. Maybe it had something to do with the fact that the Legion grid was blank, so there were no options to select from.

Regarding how long it takes to launch, that's because of how big your screen is. A 4K screen is 4X the number of pixels as a 1080p screen and will take about 4X as long to process.  :(

gerrish:
I think we need to lessen the squares. Maybe use a percentage of resolution approach because that's way too many squares on 4k.

gerrish:
I made a small change which should work fine for most resolutions except maybe 480p, which I hope no one is still using. Instead of having a constant maximum rectangle width, the grid is divided into a definite number of columns(30  by default). The maximum width is defined as the width of these columns which is relative to the resolution. Maybe the number of vertical columns could be put into the settings for tweaking.

LexiconCode:
screenshot_1

LexiconCode:
A point of clarification. `Legion` feature requires win32api and no longer relies on dragonfly? Therefore further implementation would be required for cross-platform to Linux, Mac OS?

gerrish:
We can always add an operating System check and keep the dragonfly implementation for the rest of the operating systems. Alternatively, I could contribute this to dragonfly since this seems to be a bug.

LexiconCode:
I like the alternative option. Caster will be moving to this repository for the default dragonfly install. [Danesprite/dragonfly](https://github.com/Danesprite/dragonfly)  Danesprite might be willing to help with the fix or at the very least actively integrate a pull request. If you could summarize or reference the issue there that would be great.

If that ends up not being a viable option we can continue on with the dragonfly implementation for the rest of the OS.

chilimangoes:
Note that fixing dragonfly would be more involved because it currently gets the monitor information on boot up and caches it in the exported `monitors` variable. But we want to retrieve fresh monitor configuration info each time so the user doesn't have to reboot dragon any time they change their DPI settings.

LexiconCode:
We might hold off posting the issue in Danesprite's repository until I've had a chance to discuss this with him.

gerrish:
We could always make the monitors variable a class with getters which would fetch the info on runtime. Or even add a new class.For Legion however to make the grid dpi aware an operating System check will be needed anyway. I don't think It's possible to make it completely cross-platform without dealing with each operating system's annoyances.

chilimangoes:
That's a good idea. My Python is a bit rusty and it didn't even occur to me to use a class with an indexer property, or whatever Python calls them.

Versatilus:
I'll admit that the source code for it is available and public domain, but since we're talking about using a Windows DLL that isn't currently implemented on other platforms, isn't the whole discussion kind of academic at this point?

Assuming this code is working for everyone and most people are satisfied with the box scaling code(though honestly I think it should be a setting in the configuration file), I think we should merge this as it is and separately work with Dane on fleshing out the cross-platform aspects of Dragonfly to better support this.

chilimangoes:
Absolutely. My thought was to just give it another couple days to let the dust settle on these latest commits to see if there are any other issues before pulling the merge trigger.

Versatilus:
Sounds good. I'm usually a little too trigger-happy.

gerrish:
Don't pull yet, I will put the option in the settings.

LexiconCode:
Excellent, I'm a little like that with trigger-happy as well.

LexiconCode:
@gerrish Danesprite is fine with issues and enhancements related to dragonfly be posted in his repository.

gerrish:
I have restored the dependency on dragonfly and tested it with danesprite's update. It works well and should make it easier to implement a multiplatform solution in the future.

LexiconCode:
A note for documentation.
The DPI fix implementation is not cross-platform in dragonfly. https://github.com/Danesprite/dragonfly/issues/8

gerrish:
What is required to finally merge this pull request? Something like changing the way caster is installed to include the modified dragonfly?

LexiconCode:
@gerrish we need to update the documentation to include Danesprite/dragonfly fork. 

[Readme](https://github.com/synkarius/caster/blob/master/README.md)
[readthedocs](http://caster.readthedocs.io/en/latest/README/)

gerrish:
Since the documentation currently uses videos to explain the installation process, how do you suggest we include the fork? Simply as a mention with a note to follow the instructions in the git hub page or a complete remake with full explanation?

LexiconCode:
It's up to you although I'm leaning towards keeping it simple. The install procedure my change if we are able to use https://github.com/synkarius/caster/issues/217.

gerrish:
It's done!

PR #229, opened 2018-04-11
Enclose text into quotes,parens,etc.

gerrish:
This is an experimental feature which is missing in castor but it needs testing because I have used the clipboard to make it. It works by copying the selected text into the clipboard (doesn't affect previous contents), appending the required openers and closers and pasting the new text back. 
I mostly want this to be tested by other system because it could be unstable if the text is not copied or pasted on time. Also, I would like some suggestions on the command's spec because the current form is kind of hard to utter.

To do:

1. Add the ability to enclose with multiple characters
2. Add stricter checks to avoid losing previous clipboard content.


LexiconCode:
It's a nice enhancement. I'll see if I can give it a test tomorrow!
I'll have to think on command's spec. Right now I can't get the word `squiggly` for `~` out of my head.

gerrish:
Looking at the code of dragonfly that's in a try except block. I have no Idea how that wasn't caught.

LexiconCode:
No it did not.

LexiconCode:
So I have a proposal. The functionality for formatting is awesome. Simplicity could possibly be improved.
The command `prekris` yields `()` the `trap prekris bow` yields `()` with an empty clipboard. 

Thoughts about about integrating the functionality of `trap prekris bow` into the `prekris` and subsequent commands? It's more of a question of what the default behavior is expected by the average end user and also how we set those expectations.

LexiconCode:
Looking at the source code I see we really need to eliminate the timer to make this reliable. That means not utilizing keyboard shortcuts.

Looking over dragonfly documentation.
[Clipboard toolkit](http://dragonfly.readthedocs.io/en/latest/clipboard.html?)
[action_paste](http://dragonfly.readthedocs.io/en/latest/_modules/dragonfly/actions/action_paste.html?)
It still uses by default, the *paste* action is the emulated  Key("c-v/20") keystroke with a pause.

Fundamentally this way we can paste without emulating keyboard shortcuts? 

 What are your thoughts on instead of pouring effort to solve this within Castor let's focus on Dragonfly implementation of the clipboard. [Cross-platform copy/pasting could be done with pyperclip](https://github.com/Danesprite/dragonfly/projects/2) That way Castor can benefit as well as Dragonfly.

LexiconCode:
I would love to have it simplified but I think for now the best solution like you said is an optional keyword to trigger formatting.

gerrish:
So I found something very annoying which sabotages simplification. In some editors like Visual Studio code, pressing control plus C actually copies the whole line if you haven't selected anything. So if you try this function, instead of adding empty enclosing punctuation, it pastes the whole line and closes it. This is not something which I think can be avoided and even if it can it will be pretty slow without editor integration. I have added a word to trigger this function as you said and in the future it can be revised.

PR #230, opened 2018-04-15
Add more text spacing formats e.g. hello.world, hello/world & hello\world

codingApprentice:
I decided to add more spacing formats that would be commonly used in coding or navigation of the directory. So the use of full stops for libraries, and slashes for directories.


codingApprentice:
No worries :-) @comodoro  , but are those not already included in the original Caster with the usage of the word 'spine'?
Or am I misunderstanding your request?

If you weren't aware of them already before now, text formatting commands are listed in  [section 5 of the Caster quick reference guide PDF](https://github.com/synkarius/caster/blob/master/CasterQuickReference0.5.3.pdf) , which I use so often now that I have it printed out in front of me. I really should improve my memory skills :-)

comodoro:
So should I, indeed I have missed it, thanks for pointing it out

LexiconCode:
Could you update the documentation to reflect your changes in the grammar? 
See https://github.com/synkarius/caster/pull/220 for reference

LexiconCode:
Excellence that just leaves updating CasterQuickReference.pdf
You can save to PFD from Microsoft Word as well.

PR #231, opened 2018-04-16
add navigation to explorer

kendonB:


LexiconCode:
That sounds like a plan let's get some feedback!

LexiconCode:
Let's go ahead and add this for now we can always make a revision later. Agreed?

kendonB:
Sounds good

PR #240, opened 2018-05-23
Fix Legion not appearing in the foreground

gerrish:
This is an experimental Fix for the aforementioned issue. Requires testing.

### DIFF
--- a/caster/asynch/mouse/legion.py
+++ b/caster/asynch/mouse/legion.py
@@ -5,6 +5,29 @@ import sys
 import threading
 from ctypes import *
 from dragonfly import monitors
+from os import name as osname
+
+if osname == "Windows":
+    import win32gui
+
+    def windowEnumerationHandler(hwnd, top_windows):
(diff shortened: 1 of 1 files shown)


LexiconCode:
I have not experienced this issue so I'll need a little guidance for testing.  Is there reliable method to reproduce the bug?

gerrish:
I haven't found any reliable way to reproduce it however it happens way too often to ignore. It seems to happen more often when I switch between apps.
Specifically the bug is that Legion opens up but in the background so the commands don't work.

LexiconCode:
Yeah that sounds fairly annoying. The patch seems to function as expected on my end but then again I really haven't experienced the issue. Unless you have some outstanding concerns I think we should just merge it. Unless someone else experiences this issue often enough to contribute to testing.

gerrish:
Ironically, while it seemed to work everywhere, one page in chromewith POP ups seems to have triggered the same behavior. I will just use the fix suggested in the issue instead.

PR #241, opened 2018-05-24
Include command module for editing  - multiedit

codingApprentice:
I previously mentioned on gitter, the inclusion of this particular command module for quicker editing.
It's from [t4ngo's repository](https://github.com/t4ngo/dragonfly-modules).

It's under a LGPL license, so it should be fine to include it here, since Caster is an open code source.

I know that there has been discussion about using commands which are distinguishable and audible vs commands which are easier to remember and have lower cognitive requirements to communicate between the brain and the mouth, so this module may give people the option of using easier commands to remember.

Let me know what you think. :-)


LexiconCode:
I think it's a question of more of how to include them. We need to keep consistency across the project. If These are indeed easier I'd rather have them integrated into castor rather than their own grammar. Otherwise it should be made into a filter rule which we could distribute as an alternative option. 

Does this sound reasonable? Review multi-edit commands that aren't duplicated functionality in Castor. For those that we don't have an equivalent functionality we will try to integrate. As for the others that have a different command names but identical functionality should be created as a filter rule to override the default command names without modifying source code.

LexiconCode:
By the way filter rules now work. 

Rule Filters Simplified

 Create the file `/caster/user/words.txt` and it will be read at boot time, and a rule filter created from it and added to Caster's list of rule filters. The following is an example `words.txt`:

```
<<<SPEC>>>
shock -> earthquake
<<<NOT_SPECS>>>
sauce -> up
dunce -> down
lease -> left
```

This `words.txt` will create a rule filter which goes through all Caster rules at boot time and replaces the word `shock` in any spec with the word  `earthquake `. It will also replace  `sauce `,  `dunce ` and  `lease ` with  `up `,  `down `, and  `left ` in extras and defaults in any rule. The `triple-angle-brackets` indicate mode changes. Valid modes are SPEC (for specs only), EXTRA, DEFAULT, NOT_SPECS (for extras and defaults, but not specs), and ANY (for specs, extras, and defaults). The default mode is ANY.

LexiconCode:
Closed due to lack of response

Issue #242, opened 2018-06-26
Enhance settings.py to account for alternate DNS install locations and DNS versions.


LexiconCode:
Currently `def _find_natspeak()` assumes the C drive is being utilized.  In addition we have to define each DNS version. Let's make it DNS version agnostic and search the DNS install on all drives.

>  def _find_natspeak():
>      '''Tries to find the natspeak engine.'''
>      possible_locations = [
>          "C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe",
>          "C:/Program Files (x86)/Nuance/NaturallySpeaking14/Program/natspeak.exe",
>          "C:/Program Files (x86)/Nuance/NaturallySpeaking13/Program/natspeak.exe",
>          "C:/Program Files (x86)/Nuance/NaturallySpeaking12/Program/natspeak.exe",
>  ]

[Code context](https://github.com/synkarius/caster/blob/a7b4c7a2472f0f2f3ae574287de9bd61d92be7ea/caster/lib/settings.py#L34)

LexiconCode:
We should probably also create a check that notifies the user that their version has has not been tested for example DNS 16.

LexiconCode:
Feature Complete. Waiting on review.

LexiconCode:
Implemented using [Windows Management Instrumentation ](https://pypi.org/project/WMI/) `wmi ` dependency.

`pip install wmi`

It takes approximately `15.909 seconds` to execute the `def _find_natspeak():`. I have yet to implement a mechanism to execute the function only will the absence of a "ENGINE_PATH" in `settings.Json`.

Past implementations including the original runs on startup.

If anyone has ideas for faster implementation let me know.

PR #243, opened 2018-06-30
 Detect alternate drive DNS install locations and versions #242

LexiconCode:
The premise is the search for DNS across all drives and to check DNS version.

### DIFF
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -5,7 +5,7 @@ import io
 import json
 import os
 import sys
-import win32api
+import wmi
 
 SETTINGS = {}
 _SETTINGS_PATH = os.path.realpath(__file__).split("lib")[0] + "bin\\data\\settings.json"
@@ -37,37 +37,23 @@ WSR = False
 
 
 def _find_natspeak():
-    '''Tries to find the natspeak engine on all drives.'''
-    AllDrives = win32api.GetLogicalDriveStrings()
-    DrivesList = AllDrives.split('\000')[:-1]
-    PossibleNuanceRoot = [
-        item.replace('\\', '/Program Files (x86)/Nuance/') for item in DrivesList
-    ]
-    for num, location in enumerate(PossibleNuanceRoot, start=1):
-        if os.path.isdir(location):
-            NuanceDir = os.listdir(location)
-
-            def DnsVersion(Dns):
-                UnsupportedDns = [
-                    'NaturallySpeaking9',
-                    'NaturallySpeaking10',
-                    'NaturallySpeaking11',
-                    'NaturallySpeaking12',
-                ]
-                Dns.startswith('NaturallySpeaking')
-                if Dns not in UnsupportedDns:
-                    return Dns
+    w = wmi.WMI()
+    for p in w.Win32_Product():
+        if p.Caption and 'Dragon' in p.Caption:
+            Name = '{}'.format(p.Name)
+            Version = '{}'.format(p.Version)
+            Vendor = '{}'.format(p.Vendor)
+            InstallLocation = '{}'.format(p.InstallLocation)
+            
+            if Vendor == 'Nuance Communications Inc.' and Name == 'Dragon':
+                DnsVersion = int(str(Version)[:2])
+                if DnsVersion >= 13:
+                    ExePath =  InstallLocation.replace('\\', '/') + 'Program/natspeak.exe'
+                    print ExePath
+                    if os.path.isfile(ExePath):
+                        return ExePath
                 else:
-                    print 'Dragon ' + Dns + ' is not supported by Castor. Only versions 13 and above are supported. Purchase Dragon Naturally Speaking 13 or above'
-
-            DnsVersion = filter(DnsVersion, NuanceDir)
-            NatSpeakPath = location + str(DnsVersion).strip(
-                "['']") + '/Program/natspeak.exe'
-            if os.path.isfile(NatSpeakPath):
-                return NatSpeakPath
-            else:
-                continue
-
+                    print ' Dragon Naturally Speaking' + str(Version) + ' is not supported by Castor. Only versions 13 and above are supported. Purchase Dragon Naturally Speaking 13 or above'
     print "Cannot find dragon engine path"
     return ""
 
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -54,7 +54,7 @@ def _validate_natspeak():  # Validates 'Engine Path' for DNS in settings.json
                         print "Setting DNS path to " + exe_path
                 except Exception as e:
                     print "Error saving json file " + str(e) + _SETTINGS_PATH
-                return ""
+                return exe_path
     else:
         return _find_natspeak()
 
@@ -82,7 +82,7 @@ def _find_natspeak():  # Finds DNS path and verifies supported DNS versions via 
                         version
                     ) + " is not supported by Caster. Only versions 13 and above are supported. Purchase Dragon Naturally Speaking 13 or above"
     print "Cannot find dragon engine path"
-    return ""
+    return exe_path
 
 
 # The defaults for every setting. Could be moved out into its own file.
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -37,21 +37,22 @@ WSR = False
 
 
 def _find_natspeak():
-    '''Tries to find the natspeak engine.'''
+    '''Tries to find the natspeak engine on all drives.'''
     AllDrives = win32api.GetLogicalDriveStrings()
     DrivesList = AllDrives.split('\000')[:-1]
-    PossibleDnsRoot = [item.replace('\\', '/Program Files (x86)/Nuance/')  for item in DrivesList] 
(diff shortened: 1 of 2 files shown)


caster/lib/settings.py
```diff
@@ -36,17 +37,24 @@
...
+            
+            if Vendor == 'Nuance Communications Inc.' and Name == 'Dragon':
+                DnsVersion = int(str(Version)[:2])
+                if DnsVersion >= 13:
+                    ExePath =  InstallLocation.replace('\\', '/') + 'Program/natspeak.exe'
+                    print ExePath
+                    if os.path.isfile(ExePath):
+                        return ExePath
+                else:
+                    print ' Dragon Naturally Speaking' + str(Version) + ' is not supported by Castor. Only versions 13 and above are supported. Purchase Dragon Naturally Speaking 13 or above'
```
masisley: Caster.

LexiconCode:
@masisley Nice catch! Because we use `wmi` and it takes time to search the registry of the `def _find_natspeak():`takes up to 15 seconds to complete.  I don't want to add 15 seconds to the start up time of Caster. 

I'll add a create something like  `def _verify_natspeak_path ():` which will check if the current `"ENGINE_PATH"` is valid if not `def _find_natspeak():` will be called to find the path.

LexiconCode:
This should be feature complete and robust. Your feedback would be appreciated!

There are at least two ways you can help test.

1. Change `"ENGINE_PATH": "C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe", ` in `settings.json` to an invalid path. Restart Dragon
2. Backup then delete `settings.json` regenerating default configuration. Restart Dragon

gerrish:
Will tested by tomorrow and report.

gerrish:
So:
1. This fails with an error:
```
Searching Windows Registry For DNS...
Search Complete.
Setting DNS path to C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe


ValueError('Extra data: line 85 column 2 - line 85 column 3 (char 3635 - 3636)',) while loading settings file: C:\NatLink\NatLink\MacroSystem\caster\bin\data\settings.json


(<type 'exceptions.ValueError'>, ValueError('Extra data: line 85 column 2 - line 85 column 3 (char 3635 - 3636)',), <traceback object at 0x0E9A3DC8>)

```
the engine path gets set to empty string. Restarting Dragon fixes this, the proper path is found and set,but the reboot command doesn't work until the path is set because it depends on it (extremely annoying). 
2. Works as expected.

caster/lib/settings.py
```diff
@@ -71,7 +106,7 @@ def _find_natspeak():
         # EXECUTABLES
         "DEFAULT_BROWSER_PATH": "C:/Program Files (x86)/Mozilla Firefox/firefox.exe",
         "DOUGLAS_PATH": BASE_PATH + "/asynch/mouse/grids.py",
-        "ENGINE_PATH": _find_natspeak(),
+        "ENGINE_PATH": _validate_natspeak(),
```
gerrish: Validate should return the exe path , instead it returns  empty string as noted in the above comment.

caster/lib/settings.py
```diff
@@ -35,18 +36,52 @@
...
+                data["paths"]["ENGINE_PATH"] = exe_path
+                try:
+                    formatted_data = unicode(
+                        json.dumps(data, sort_keys=True, indent=4, ensure_ascii=False))
+                    with io.open(_SETTINGS_PATH, "r+", encoding="utf-8") as json_file:
+                        json_file.write(formatted_data)
+                        print "Setting DNS path to " + exe_path
+                except Exception as e:
+                    print "Error saving json file " + str(e) + _SETTINGS_PATH
+                return ""
```
gerrish: This should return exe_path so that empty string is not set as the default engine path.

LexiconCode:
I appreciate the feedback! Showing my ignorance here. What were the steps to reproduce this? 

> the engine path gets set to empty string. Restarting Dragon fixes this, the proper path is found and set,but the reboot command doesn't work until the path is set because it depends on it (extremely annoying).

My intent was the only condition when an empty string is set when  `_find_natspeak()` cannot find DNS path/supported DNS or `_validate_natspeak()` has an error loading/saving `settings.json`. Earlier on I did encounter this error but it was due to `"ENGINE_PATH":` set to `Null` instead of a `string`. 

So an empty string shouldn't cause failure to `settings.json`. to load or break json formatting. I'll defer to your advice but I'm trying to understand my failure of logic. 

In addition, could you upload `settings.json` that contains the extra data?

LexiconCode:
Upon further thinking `_find_natspeak()`
if I were to return `exe_path` instead of `""` exe_path would be referenced before it was assigned as a local variable?

LexiconCode:
While I don't understand the steps to reproduce the described error I understand the logic of your recommendation. If loading/saving of `settings.json` throws an error at least the `ENGINE_PATH:` in memory contains the correct path instead of an empty string. This would allow the `reboot dragon` command to function even if `settings.Json` did not update as expected.

gerrish:
Destroy the path  and reboot:
```"ENGINE_PATH": "C:/Programhello Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe", ```
the new path is: 
`"ENGINE_PATH": "", `

The full error message because I missed an important line:
```Searching Windows Registry For DNS...
Search Complete.
Setting DNS path to C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe


ValueError('Extra data: line 85 column 3 - line 86 column 2 (char 3636 - 3639)',) while loading settings file: C:\NatLink\NatLink\MacroSystem\caster\bin\data\settings.json


(<type 'exceptions.ValueError'>, ValueError('Extra data: line 85 column 3 - line 86 column 2 (char 3636 - 3639)',), <traceback object at 0x0EB74DA0>)
Default settings values added: 7
 ```

The reboot Dragon command opens a file Explorer window to the correct path but doesn't actually run NatSpeak.exe. 
After restarting Dragon again:
```Searching Windows Registry For DNS...
Search Complete.
Setting DNS path to C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe
```
to see what happens, I put the _save call from the _init function in a comment. Your function somehow adds extra curly braces to the json file. This is why it fails to load and instead loads the default values which had the engine path set as empty string. I have no idea why this happens.

LexiconCode:
Thanks for explaining. 
So to reproduce
> > Destroy the path and reboot:
> "ENGINE_PATH": "C:/Programhello Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe",
> the new path is:
> "ENGINE_PATH": "",
I can't reproduce the error. 

Can you reproduce it by 
1. Backing up then deleting `settings.json`
2. Letting the default `settings.json` generate with the correct path
3. Edit the path and reboot:
> "ENGINE_PATH": "C:/Programhello Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe",
> the new path is:
> "ENGINE_PATH": "",
4. Is the error reproduced?

gerrish:
I first deleted settings.json, rebooted and then destroyed the path and rebooted again.
File generated from your code(commented out default settings override):https://ufile.io/gxd8h
original settings(after default settings override):https://ufile.io/7trlo
Exact same error , but the proper engine path is set after your latest change. Notice the extra braces in the last 2 lines of the first file.

gerrish:
This could be something wonky but what seems to fix the problem is to use a "w" flag instead of "r+" when you open settings.json inside validate_NatSpeak. Instead of truncating the file and then writing it ,you are  writing over existing character.Since your new configuration is actually a smaller file than the one I create with extra noise in the path ,some extra brackets remain in the end. After testing with longer noise text, this seems to be the case as the extra characters end up in other positions.

LexiconCode:
Thank you for taking the time to test.

Feels like I'm banging my head against a brick wall. Fortunately it makes sense why I can't reproduce it up till now. My bad assumption was I thought it was overwriting the file instead of truncating it. I really do dislike working with `json` in part because I'm new to utilizing it and because you can't update in place data.

I'm still flummoxed by the idea I can't reproduce it using.
>         Destroy the path and reboot:
>         "ENGINE_PATH": "C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe",
>         the new path is:
>         "ENGINE_PATH": "",
> 
> I can't reproduce the error.
> 
> Can you reproduce it by?
> 
>     Backing up then deleting settings.json
>     Letting the default settings.json generate with the correct path
>     Edit the path and reboot:
> 
>     "ENGINE_PATH": "C:/Program Files (x86)/Nuance/NaturallySpeaking15/Program/natspeak.exe",
>     the new path is:
>     "ENGINE_PATH": "",

Regardless I believe you hit the nail on the head though with the root of the problem. I'll include  `"w" flag instead of "r+"` in `def _validate_natspeak()`

LexiconCode:
Hopefully this is resolved and let me know. Thanks for your time to review and test!

gerrish:
Seems to work great now.

PR #245, opened 2018-07-02
Alias fix

gerrish:
1.Vanilla Alias class  has been added to the nexus.
2.Spec has been changed to `Alias <s>` from `vanilla Alias <s>`.


LexiconCode:
Tested and works as expected.

PR #246, opened 2018-07-16
Caster as a Python package

gerrish:
This is a preliminary pull request which will eventually include everything required to make caster a Python package. It is not ready for testing due to requiring changes to settings.py.

 The setup.py file will install the lib,asynch sub packages into the installed python packages. This eliminates the annoying import error messages when debugging caster modules and will eventually lead to the splitting of caster into 2 parts:

1. **The user space**, which will include user specific configuration files like settings.json and the user package which allows for the creation and application of new grammars and filters.
2. **The framework space**, which will be installed as a Python package and will contain the "backend", namely the default grammars, the base caster engine modules and caster utilities like the mouse grids.

Things required:

-Modify settings.py paths to point to the correct locations
-Store user data in an appropriate location
Feedback required on:
-The proposed new file structure for caster
-Other suggestions for the setup file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,21 @@
+from setuptools import setup,find_packages
+with open("README.md","r") as rm:
+    long_d=rm.read()
+with open("LICENSE.md","r") as lc:
+    _license=lc.read()
+setup(
+    name="caster",
+    version="0.5.10",
+    author="synkarius",
+    packages=find_packages(exclude=("user",)),
+    long_description=long_d,
+    license=_license,
+    url= "https://github.com/synkarius/caster",
+    install_requires=[
+        "pywin32;platform_system=='Windows'",
+        "git+ssh://git@github.com/Danesprite/dragonfly.git",
+
+        "Pillow",
+    ],
+
+)
--- a/setup.py
+++ b/setup.py
@@ -13,9 +13,12 @@ setup(
     url= "https://github.com/synkarius/caster",
     install_requires=[
         "pywin32;platform_system=='Windows'",
-        "git+ssh://git@github.com/Danesprite/dragonfly.git",
+        
 
         "Pillow",
     ],
+    dependency_links=[
+        "git+ssh://git@github.com/Danesprite/dragonfly.git#egg=dragonfly",
+    ]
 
 )


LexiconCode:
FYI Another interesting package manager [Poetry](https://poetry.eustace.io/)

In terms of managing packaging this looks really well done from start to finish. It publishes to PyPI so the end-user can utilize PIP. Anyway it's just another option among many that might be useful.

gerrish:
I would like to try the stock solution first, to avoid adding new dependencies and because there is a lot of documentation. When it's done, it shouldn't be too hard to upgrade it. Also, as far as publishing is concerned, shouldn't we consult with synkarius before doing so since we will need to provide authors names and emails etc. ?

Furthermore, some notes about the changes I plan to make for feedback:
1. There will be a user directory which will be set into {user directory}/caster/user_modules for user filters and grammars, /user_config for settings,aliases etc. (this should be user configurable but from the pyinstaller setup and not at this stage)
2. Asynch modules will be installed to Python site packages as well. They will be invoked as installed Python modules like `Python -m module_name`.
3. Settings.py will be updated to contain the new paths and to validate the existence of the user directory. Also, they will automatically generate the required files on first launch.
4. Due to the user modules being in a different directory than the _caster.py file, a new module will be created which will dynamically import the package using the directory path. (Maybe even the ccr grammars could be removed from /lib and act as plug-ins using the same method)
5. Ideally tests will need to be added to validate the new directory structure and other things(will probably be in another pull request as a complete feature)

gerrish:
I was thinking about using pyinstaller for them. This pull request will deal with the Core parts of caster.  after its done,  We will make a standard installer which will bundle Python, natlink, Windows32 and caster and install everything. Part of the installation process will be to use pip  to install caster as a Python package.

comodoro:
Hi @gerrish , I am interested particularly in the moving out of user data, please let me know if I can help. A small suggestion would be to start the user directory subdirectory with a dot (.caster), which is customary in  Unix and now Windows as well

LexiconCode:
@gerrish Danesprite/dragonfly Is now installable through pip as dragonfly2. Simply close Dragon NaturallySpeaking and run 'pip uninstall dragonfly' then `pip install dragonfly2` from command line to use the new distribution.

LexiconCode:
Fulfilled by https://github.com/dictation-toolbox/caster/commit/b244c1256c80ed479d7058cc60de62b070ae6422

PR #249, opened 2018-09-05
Update R functionality

mrob95:
Cleaned up some of the commands in shippy's R module, especially ones which were duplicating standard punctuation commands, updated the init file, and added some more common commands.


mrob95:
Are the conflicts a problem? Shall I reset my branch to the updated "develop" and then recommit the changes?

LexiconCode:
I resolved the conflicts through the web interface. Do a final check of the grammars and test their functionality as I don't have any programs that utilize R or matlab. Once you're done let me know and I'll merge.

Thanks for your work and I look forward to seeing what you might develop!

mrob95:
Just removed the long comment functionality from the R module – there is no such thing in R. I'm not familiar with Matlab but all of the commands now seem to be working as intended.  
All of the R commands are working except for one, mysteriously. Saying "as character" just returns "as character", which I assume is just from the standard Dragon Dictation output. I can't see any reason why this wouldn't work, given that "as.factor()" and "as.numeric()" work as intended. Can you replicate?

LexiconCode:
The `as character` command returns `as.character()` as expected without switching command mode in Dragon settings.

For testing purposes switch in Dragon NaturallySpeaking settings to "commands' mode instead of 'dictation and commands' That will force Dragon to recognize only commands and not recognize free dictation.

mrob95:
Ah that fixes it.

LexiconCode:
Great thanks again for your contribution!

PR #253, opened 2018-09-09
Make the SQL module output in caps.

mrob95:
Change the SQL module to output in capitals, as is standard. Added a couple more functions.

### DIFF
--- a/caster/lib/ccr/html/html.py
+++ b/caster/lib/ccr/html/html.py
@@ -65,14 +65,12 @@ class HTML(MergeRule):
             Text("<h5></h5>") + Key("left/10:5"),
         "H 6 | heading six":
             Text("<h6></h6>") + Key("left/10:5"),
-        "isolate ":
-            Text("<bdi></bdi>") + Key("left/10:6"),
         "H group | headings group":
             Text("<hgroup></hgroup>") + Key("left/10:9"),
         "navigation | navigate":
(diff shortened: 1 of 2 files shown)


mrob95:
Apologies for the spam. HTML functions are all working properly now though :-)

LexiconCode:
Not a problem! I appreciate the work 👍. There is a way to reduce the spam for the commit which would be nice. Something I'm looking to do for my own commits so their concise. [Squash All Commits Related to a Single Issue into a Single Commit](https://github.com/todotxt/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit), [stackoverflow]( https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git/5201642#5201642)

This also allows you to also change the commit message that is displayed on the repository before pushed. If I were to merge the pull request currently it would display 'another' in the repository versus a commit message that describes of the issue.

Something like 'Capitalized SQL syntax and add additional commands #253' 

Notice the pound symbol with the numbers. These are used for issues/pull requests and are unique. Including them in the commit message automatically references the issues making them easier to track. For instance you conclude the issue and the pull request if they are related. Eg. `#253 #342`

Often things are opened up as an issue then they turn into a pull request. So it's so it's helpful to reference both. In in your case it's just a pull request.

PR #256, opened 2018-09-11
Hearthstone WIP/proof of concept

mrob95:
A caster module which allows Hearthstone (the PC game) to be played entirely by voice. It has two elements:
* A screen overlay linking the positions of the cards in game to the commands which need to be spoken to click on them.
* A CCR module with the grammar and functions in (requires pyautogui).

An overlay PNG (from this project: https://github.com/wujerry573/TwitchPlaysHearthstone) is included, but only for 1080p resolution. I'm using this utility to lay it over my screen: http://customdesktoplogo.wikidot.com/download. I've included a command to save the current mouse position to a text file, which should make adding new positions fairly easy.

It's pretty hacky at the moment, but seems to be working in the few tests I've done. Long-term, ideally it would detect the size and position of the Hearthstone window, calculate (or recall) the positions, render and display the overlay, and update the commands all from within python.

hs

Ignore my rank! I only just started playing again :-). Yes I had lethal last turn, ran out of time because of background noise.


mrob95:
Good point, I'll update it. I'm not sure this is a good fit for adding to standard caster functionality since it's probably only of interest to a minority, and isn't coding related. Do you think it would be better to have it as a separate project, and link to it in the docs for those who are interested?

LexiconCode:
It could be a separate project for the reasons you stated. Include it in the documentation as a example.

LexiconCode:
I wonder if we could create a subfolder within apps folder for organizational purposes. That way we Include projects like this is a showcase within the caster project.

LexiconCode:
@mrob95 I assume this is feature complete?

mrob95:
It is as feature complete as I think I'm going to make it. I would be tempted to leave it out of the main caster repository, and perhaps just link to it for those interested.
https://github.com/mrob95/caster/tree/hearthstone/caster/lib/ccr/hearthstone

PR #259, opened 2018-09-13
Applications quick reference

mrob95:
Adds a quick reference guide for the application-specific commands.


mrob95:
Hmmm, I'm taking the keys directly from the AtomRule.mapping dictionary, so there's no obvious way to automatically include the comments. It would have been nice to preserve the original ordering too, but dictionaries aren't ordered so it's either alphabetic or random with this method.
https://github.com/mrob95/caster/blob/doc_update/generator_apps.py

LexiconCode:
Well that's a nifty script. All things considered what we have works.

PR #260, opened 2018-09-15
Fix find natspeak in DNS Home 13

comodoro:
Strangely the new method to find NatSpeak does not work for me, because the name in the registry Is not Dragon, but something like Dragon from Nuance Communications Inc. This fixes that


LexiconCode:
Thank you for your time and effort! I'll mark this as won't fix due to the mismatch in the CodeBase which is out of date with the upstream development branch.

PR #261, opened 2018-09-15
Stoosh spark update – needs testing

mrob95:
Previously, saying "duple" would copy the current line to the clipboard, make a new line and then paste. This works, but it replaces whatever is in the clipboard with the line you are duplicating, which can be annoying. This is fairly simply fixed by altering all commands which use the clipboard so that they save and restore the clipboard before and after operations.

Commands like "stoosh/spark five" and "duple" should now no longer replace the normal clipboard, provided what's in the clipboard is only a string. Now that this is done it should be fairly easy to implement the first part of #214. 

I've been playing around to get the timings right so that it is completely reliable, but it would be worthwhile someone else testing it to make sure it's working properly.


LexiconCode:
Everything works as expected. 

Format your code through yapf https://github.com/synkarius/caster/issues/197 and I'll get this merged.

PR #262, opened 2018-09-15
Add default format and secondary format

### LINKED ISSUE #255: Add secondary format and default format to language rules
When switching to a new language, one should not have to set format explicitly, but have format for the given language already set, like for Java Gerrish gum bow, for Python snake laws etc.

Second format convention exists in most languages, like ClassName in Java or Python, there should be a command for that

comodoro:
closes #255


LexiconCode:
For end-user configurable settings we should move out of the source code into settings.py 
```
    text_format = 3, 1 #gerrishGum
    secondary_text_format = 2, 1 #TieGum
```

comodoro:
I did have doubts if this should be explicitly configurable. A dictionary in default settings to be loaded from ccrmerger.merge?

LexiconCode:
> 
> 
> I did have doubts if this should be explicitly configurable. A dictionary in default settings to be loaded from ccrmerger.merge?

I agree ccrmerger.merge would be reasonable.

LexiconCode:
@comodoro The last thing to do would be to update the documentation to reflect your commands.

comodoro:
@LexiconCode I cannot find the appropriate place with related commands. I should probably create it in CCR.MD, under the Command sets paragraph, right?

LexiconCode:
> 
> 
> @LexiconCode I cannot find the appropriate place with related commands. I should probably create it in CCR.MD, under the Command sets paragraph, right?

I would include it in the CasterQuickReference under text formatting with a new subcategory.

comodoro:
OK, but it will not be immediately. The docx format is a bit unfortunate, I have tried Libre Office and Wordpad, but they both render it differently and both a bit wrong. I do not have MS Word at hand.

comodoro:
@LexiconCode Realizing that this does not reflect 1. restarting Caster, 2. disabling grammars. I will add an appropriate data structure to settings to track last enabled grammars?

comodoro:
Sorry I'm not clear. This does not change format from default if you restart Caster but enabled grammars are persisted. After disabling a grammar it does not revert to the last enabled grammar.

comodoro:
Pushed the docs, feel free to correct me.

LexiconCode:
@comodoro ready to merge?

comodoro:
@LexiconCode I should make it match enabled grammars. I am pleasantly surprised that the toml thing does not require me to do anything.

LexiconCode:
Thank you for this. We can re-factor at a later date as needed. Feel free to add in unit tests as that's important and something I know very little about.  Due to the size of the pull request I need at least one more person the test or review besides us before I am comfortable merging.

caster/lib/dfplus/merge/ccrmerger.py
```diff
@@ -299,7 +332,7 @@ def merge(self, time, name=None, enable=True, save=False):
             rule.set_context(context)
             active_apps.append(rule)
         '''negation context for appless version of base rule'''
-        contexts = [rule.get_context() for rule in self._app_rules.values() \
+        contexts = [app_rule.get_context() for app_rule in self._app_rules.values() \
                     if rule.get_context() is not None]# get all contexts
```
mrob95: Shouldn't this be?:
```suggestion
                    if app_rule.get_context() is not None]# get all contexts
```
or am I missing something?

comodoro:
Thanks @mrob95 ! This was an oversight and confirms the rule "If it ain't broken, don't fix it". The reason for this was Python 2's leaking listcomp variables, but unnecessary and introducing bug on my part

comodoro:
Added what I understand to be basic unit tests, E.g. without dependencies and very simple

mrob95:
Looks good, I'd be happy to merge. I don't mind updating the docs in one go, probably more efficient than making individual changes with every PR.

LexiconCode:
Alright so the plan is to merge and then add the docs in a second pull request?

Issue #247, opened 2018-07-19
Utilizing new format instead of json for Caster settings file.


LexiconCode:
Json is a great data interchange format but not really a format for end-user facing files like containing settings. While I realize Caster is made for developers but continuing to simplify towards the average end-user brings benefits to both populations. Put simply, simple is simple for everyone.

To summarize some of the limitations of json format.
1. It's spec is noisy requiring a lot of formatting. This impacts human errors with format and readability.
2. It does not support `comments` which are important for settings files.
3. Doesn't support multiline
and so on.

So thoughts about a new format? I'm leaning towards `toml` as it seems a simpler implementation then `yml`. 

Only `settings.json` would change but for now everything else that utilizes Json for data would be status quo.

A bit of research.
`pip install toml`
[Comparison to other json yaml](https://gohugohq.com/howto/toml-json-yaml-comparison/)
[Details on markup language](https://github.com/toml-lang/toml)

Caster settings file output comparison between `json`,`toml`, `yml` of settings.json.
[Settings.zip](https://github.com/synkarius/caster/files/2211288/Settings.zip)

gerrish:
I like the idea,toml seems like a much better fit for a settings file indeed. The current format is not easy on the eyes and seems more like an auto-generated file rather than something the user can tweak. I would prefer it if we could keep the current format until I can finish the Python installation pull request so as to not complicate that, unless the implementation stays The same.

LexiconCode:
I'll created new branch on my fork and experiment there. For the most part it would closely mirror the current implementation but there won't be a pull request until you finish your current project.

PR #269, opened 2018-10-08
Replace Json for Toml in Caster.

LexiconCode:
This was originally born out of utilizing new format instead of json for Caster settings file. #247  That issue states the reasons for the switch to a new format and gives examples of the conversion.  Toml is pretty much a drop in replacement for Json for our purposes. This pull request drops _all_ Json in favor of Toml in Caster.

To test install toml from command prompt/power shell.
`pip install toml`

The only thing I have not figured out the equivalency of jsons `sort_keys=True` for loading and saving toml files.


Versatilus:
I can definitely say I support this effort of yours and I'm in favor of generally moving forward with it, but I'm still playing catch-up and so I'm not yet ready to review the code itself. Don't let that slow you down, though. I'll catch up eventually.

LexiconCode:
Once this is merged default settings will be regenerated the toml format.

PR #277, opened 2018-10-11
Added a LaTeX ccr module

mrob95:
LaTeX is a markup language often used in academia for producing technical documents and scientific papers in plain text.

I have included all of the obvious commands which I could think of, but I will be using this quite a lot in the coming year so I may update it with anything I've forgotten.

https://www.latex-project.org/

formatted


LexiconCode:
That's great addition! Looks like the only thing left is to add the documentation.

PR #278, opened 2018-10-11
Add text size modifier to spark

### LINKED ISSUE #267: Apply Formatting Commands to text in system clipboard buffer.
@btlaxman in Gitter requested that formatting commands manipulate text in the clipboard.

The text would be formatted in Clipboard buffer then pasted. Example spec `yell spark` or `laws spark`.

comodoro:
closes #267

### DIFF
--- a/caster/lib/ccr/core/nav.py
+++ b/caster/lib/ccr/core/nav.py
@@ -195,7 +195,7 @@ class Navigation(MergeRule):
             R(Function(navigation.stoosh_keep_clipboard, nexus=_NEXUS), rspec="stoosh", rdescript="Copy"),
         "cut [<nnavi500>]":
             R(Function(navigation.cut_keep_clipboard, nexus=_NEXUS), rspec="cut", rdescript="Cut"),
-        "spark [<nnavi500>]":
+        "[<textsize>] spark [<nnavi500>]":
             R(Function(navigation.drop_keep_clipboard, nexus=_NEXUS), rspec="spark", rdescript="Paste"),
 
         "deli [<nnavi50>]":
@@ -288,6 +288,10 @@ class Navigation(MergeRule):
         Choice("extreme", {
             "Wally": "way",
         }),
+        Choice("textsize", {
+            "yell": True,
+            "laws": False,
+        })
     ]
 
     defaults = {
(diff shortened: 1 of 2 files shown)


LexiconCode:
An excellent feature. Thanks for implementing it.

Could we expand it to all the text formatting commands?
Then will also need to amend the documentation.

comodoro:
Hmm, maybe after #262. I know I have at least Word 2013 somewhere, that should suffice.

LexiconCode:
@comodoro I wonder we can ditch the Microsoft file altogether by re-creating it in markdown. I'm not sure if markdown format is flexible enough. If it does work I'm sure we could find a way to convert it to PDF from markdown. [Typora](https://typora.io/) makes writing markdown easy.

comodoro:
I have found Word 2013. It serves, but I wonder if it breaks something for you guys. I think changing the format is a very good idea. I would think [LaTeX](https://www.overleaf.com/) may be better, odt source is still not human readable. Plus merging conflicting changes is more or less impossible, like just now.

Markdown: Simple, but not sure it is capable of producing a sheet like that
LaTeX: May take some experimenting, but capable of producing almost any PDF

mrob95:
Here is my first stab at getting something like this working for all potential formats:

https://github.com/synkarius/caster/compare/develop...mrob95:spark_formatting

LexiconCode:
@mrob95 
Perhaps we can bypass the need to say the spec `bow` with  `Playback([(["bow"], 0.0)])`?

mrob95:
I believe you could do it just by removing `bow` from the spec, no? I think we should keep it consistent with `set format` though, either keeping it in for both or removing it from both.

comodoro:
@mrob95 @LexiconCode I have waited because of #262, because it is bound to cause conflicts, but this change does not look too big, so not too big of a problem. I have no idea if the `bow` suffix is beneficial or not, i think its purpose is acoustic differentiating, but `spark` will probably typically not be part of a long CCR sequence, so it may not be needed.

LexiconCode:
As soon as we can get one more person to test and look over #262 I'll merge.

mrob95:
Sorry I'm unable to test much at the moment, my working branch is quite a few commits behind. Going to do a big update once the Aenea support is merged.

LexiconCode:
Fulfilled by https://github.com/synkarius/caster/issues/326

PR #282, opened 2018-10-17
Detect browser dynamically

comodoro:
Browser path was hardcoded in settings. This attempts to get default browser from the registry an falls back to `os.startfile` (with the opening in background annoyance). The utility function can be used elsewhere


LexiconCode:
Fantastic I was hoping on implementing this myself in the future. 
Eventually I'd like to start thinking of Linux for such features as well.

LexiconCode:
Tests and works as expected.

PR #283, opened 2018-10-17
Cleaning up the R module

mrob95:
I've re-factored the R module. There were a large number of commands which all did basically the same thing – write a particular function and then move inside the brackets – which I've reduced to just a couple of commands, with options in the extras. There are a few benefits to this:
* Reduces code volume and repetition.
* Makes it easier and more intuitive to add new extras.
* Makes it easier to modify how the command works, e.g adding a prefix.

For consistency, I have made the prefix for all ggplot commands "graph". I have also added a prefix – "fun" – to the command which produces R functions. A small reduction in speed for a large gain in phonetic distinctness and a reduction in mistakes.

### DIFF
--- a/caster/doc/readthedocs/CCR languages quick reference.md
+++ b/caster/doc/readthedocs/CCR languages quick reference.md
@@ -315,7 +315,7 @@ lodge and|` and `|||``
 # R
 Command | Output | | Command | Output
 ---|---|---|---|---
-`fun <function>`|`<function>()`||lodge not|`!`
+`<function>`|`<function>()`||lodge not|`!`
 NA|`NA`||lodge or| &#124; &#124;
 add comment|`#`||print to console|`print()`
 assign|` <- `||return|`return()`
@@ -330,11 +330,12 @@ iffae|`if ()`||value not|`NULL`
 import|`library()`||value true|`TRUE`
 lodge and|` && `||while loop|`while ()`
 
-## Functions
-Should all be prefixed with `fun `  
+## Core/tidyverse functions
+If you are having trouble with over-recognition (R commands appearing where you don't want them) due to the large number of commands, consider adding a prefix to the `<function>` command - line 93 - in lib/ccr/r/r.py to increase phonetic distinctness.
 
 Command | Output | | Command | Output
 ---|---|---|---|---
+``|``||length|`length()`
 (LM / linear model)|`lm()`||library|`library()`
 arrange|`arrange()`||list|`list()`
 as character|`as.character()`||mean|`mean()`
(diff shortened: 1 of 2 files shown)


Digital-ecologist:
Hi. I just figured out how to submit a pull request with the commands I've added to r.py. Hopefully I did that correctly. I also did some more thinking, and realize all the "gee" commands should definitely be converted to "graph". As ggplot is the most common graphing package. If others like hexbin and or SPplot eventually get added in, they should be the ones with the less standard prefixing, as their less common. If adding the "fun" prefix solves a problem for you, go for it. I just started using this, so it's not like I have the old style burned into my brain. As I was editing the r.py  I realize we should include @shippy as he might be using this tool as well.

mrob95:
I think the best solution may be to change the command so that the original functionality is preserved, but put something in the docs like "if you are having trouble with over-recognition, consider adding a prefix to the command in r.py to increase phonetic distinctness". I'll test this now and make sure that it works as intended.

shippy:
I like this! My original R grammar didn't have much in the way of design; this is better.

"gee" vs. "graph" might be a good thing to extract into a customizable option - while "graph" is semantically expressive, "gee" rolls off the tongue easier.

Beyond the scope of this PR, the "what if we implement other plotting tools?" does bring up an interesting tension: I think many libraries really should be treated as language grammars in their own right; since their use in statistical languages is often a deeply personal choice, their inclusion or exclusion should be configured by the end user. (This means e.g. "enable tidyverse," treating it as a separate language, or some setting that determines which packages are co-activated with R? I'm not sure `caster` has architecture for this right now.)

Other than that, LGTM :)

mrob95:
@shippy I agree that it might be nice to split commands for specific libraries up in some way, to reduce the number of superfluous commands which are active at a given time. I'm not sure there is a good way to do this though – it would be a pain to have to say "enable R", "enable tidyverse"... every time you start.

My own (somewhat brute force) solution to this problem has just been to comment out anything which I'm unlikely to use. The change in this PR makes that easier – if you aren't going to use any of the ggplot commands then you can just comment out that one command to disable them.

LexiconCode:
All right when we consider this ready to merge let me know.

mrob95:
It's working well for me but it might be worth @Digital-ecologist giving it a go to make sure I haven't broken anything.

Digital-ecologist:
Hi, I was at work most of the day, so I didn't have time to test this until late. Right this minute, everything is working beautifully. But I had a half-hour in the beginning where nothing below line 93 was working. Looking at the structure, I assumed there was some "enable" commands that I had to speak. In order to activate that part of the grammar.
I tried several, and all of them produced text output, so I thought they didn't work, but then suddenly the whole grammar was working flawlessly, so I must have stumbled on the right enable commands. What exactly is the enable command for the extras/everything below 93?   Right now, everything works fantastic thank you!

LexiconCode:
With confirmation I'll merge.

PR #299, opened 2018-10-23
Migrate Quick Reference to LaTeX

mrob95:
Primary advantages are:

* Can be edited in any text editor, doesn't require MS word.
* Once the basic formatting has been set up, commands created etc, it is very simple to move things around and add new sections without breaking everything.
* Automatically generates footnotes and numbers section titles.

I hope that the syntax is clear, the main commands are \sectiontitle{} and \command{}{}.

I have moved some of the pieces around, and I think it makes more sense now. If there is any functionality which needs adding then let me know.


mrob95:
Need to add the "delete recorded macros" command to section 15. Will wait to see if anything else comes up before I do though.

LexiconCode:
As soon as is ready to merge the next step is to integrate automatically generating documentation. I have a functioning method of printing out all available specs directly from grammars loaded in the dragonfly engine.

mrob95:
Nice! Can it print the output as well? If not then we may need some creative use of the "Mimic" function.

LexiconCode:
It can! Currently it prints out to the natlink window. Once dragonfly is updated to 0.8.1 we can begin work.

LexiconCode:
You'll need to reopen under the development branch. I could've easily switched it if I would've caught it before merge. I apologize for that.

LexiconCode:
@mrob95 You'll need to reopen under the development branch. I could've easily switched it if I would've caught it before merge. I apologize for that.

Issue #302, opened 2018-10-25
Alternative approaches managing settings for 'alias'/'chain alias' and 'record from history' spec


LexiconCode:
**Issues**
1. Currently `ccr.toml` contains some duplicate settings found in `settings.toml`. In the case of `alias = true`  and `"chain alias" = true` they both need to be enabled `[feature_rules]` and `[selfmod]` for the spec to work as expected.  This is highly unexpected from the end-user perspective. 

2.  The `record from history` spec represented in settings as  `history = true` would fit under `[feature_rules]` instead of `[selfmod]`. I would expect to toggle these commands to the main settings file instead of ccr.

settings.toml
```
[feature_rules]
again = true
hmc = true
chainalias = true
alias = true
```
ccr.toml
```
[selfmod]
alias = true
"chain alias" = true
history = true
```
**Solutions**
So we can proceed with the following solutions.
1. Merge settings `[selfmod]` from `ccr.toml` into `[feature_rules]` in `settings.toml`
2. Eliminate `ccr.toml` and merge its settings into `settings.toml`
3. Edit the defaults for [selfmod] to `true` in  `ccr.toml`

**References**
The 'chain alias' command is nonfunctional #279
The 'record from history' command does not execute. #272

LexiconCode:
Alright well I'm leaning towards option one.

PR #306, opened 2018-10-26
Converts ccr.toml [selfmod] as true instead of false.

LexiconCode:
Option three defined in 'Alternative approaches managing settings for 'alias'/'chain alias' and 'record from history' spec #302'


LexiconCode:
Note you will need to delete ccr.toml or edit the values to true.

PR #310, opened 2018-10-28
Add "splat" and "shift click" commands

mrob95:
A couple of commands which I have added for text navigation, which I find very useful.

* Splat command adds functionality for deleting whole words, in both directions, using the c-del and c-bkspc keystrokes. eg `splat three`, `splat ross`.
* Shift click does what it says on the tin.

I may make one or two other changes this week so I will update the docs in one go, separately.


LexiconCode:
Fantastic! Are you tied to the spec names?

LexiconCode:
To reduce cognitive load it's smart like utilizing previous Caster specs as much as possible. So any feedback to that would be beneficial. Feel free to change it up though filter rules.

PR #319, opened 2018-11-11
Latex Quick reference

mrob95:
This should make it easier to edit and diff the quick reference document.


LexiconCode:
Excellent :)

PR #324, opened 2018-11-15
Implemented Caster dependency check and executing terminal/CMD commands

LexiconCode:
This pull request implements:
- Checking for dependencies (in this case dragonfly)
- Execution of terminal commands 
- A command to update dependencies
- A system to protect against potential intentional or accidental damage to a system by terminal/cmd commands. 

During the startup of Caster a function checks updates for the dragonfly package. The user is notified if the package about the package update status. in the event that the package is out of date you can say
 'update caster dependencies' to update and caster will automatically reboot.

In settings.py added "online_status" to denote Internet access as defined manually by the user. It's default is true but if set to false Caster will not search for updates during startup. This is useful if caster is an off-line environment for instance in a Aenea virtual environment. 

terminal.py executes the terminal/CMD commands without a terminal window pop-up in Windows.

**Making your own terminal voice commands.**
Spec can be used to call arbitrary CMD line or terminal commands. **Note**: the command string must be exactly as if you were to execute it in the terminal window.

`"spec": R(Function(terminal_command, command='pip install --upgrade dragonfly2'), rdescript="command definition"),`

Error Code Handling.
- If the valid command is executed but an error code is returned the user be notified to run the command manually in command prompt or terminal to debug.
- If the command is invalid then the user will be notified.
**Note**  An edge case that is not handled as if the 1st part of the command is valid but the 2nd is not. For example `ls gibberish`. `ls` lists directory structure and `gibberish` is invalid because it does not exist. the command will still try to execute.

**Trusted versus Suspect commands**
Terminal commands a powerful could do potential intentional or accidental damage to a system. Therefore the concept of trusted versus suspect commands.
- Trust commands don't notify as user defined in trusted_commands in terminal.py
- All other commands utilize ConfirmAction as a safeguard for execution.

Along side with the hardcoded trusted commands in the future I might include a user configurable settings file for defining your own trusted commands.

A big thanks to @Danesprite for helping implement threading and being a sounding board for my ideas.

### DIFF
--- a/caster/bin/inno/dependencies/Pip Install Dependencies.bat
+++ b/caster/bin/inno/dependencies/Pip Install Dependencies.bat
@@ -1,12 +1,14 @@
 @echo off
-echo Installing: dragonfly2, wxPython, pillow
+echo Installing: setuptools, future, dragonfly2, wxPython, pillow, toml
 
 cd c:\python27\scripts
-pip install setuptools
-pip install dragonfly2
+pip install -U pip
+pip install -U setuptools
+pip install -U future
+pip install -U dragonfly2
 pip install -U wxPython
-pip install pillow
-pip install toml
+pip install -U pillow
+pip install -U toml
 
 
 
--- a/caster/bin/inno/dependencies/Pip Uninstall Dependencies.bat
+++ b/caster/bin/inno/dependencies/Pip Uninstall Dependencies.bat
@@ -1,5 +1,5 @@
 @echo off
-echo Uninstall dragonfly2, wxPython, pillow, six, pyperclip, pillow, pywin32
+echo Uninstall dragonfly2, wxPython, future, pillow, six, pyperclip, pillow, pywin32
 echo ####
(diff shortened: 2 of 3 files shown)


LexiconCode:
I will need to add documentation.

LexiconCode:
Thank you for testing and even more so your feedback. As we talked about in gitter we will move `TerminalCommand`  into dragonfly. Then I will Implement customizations utilizing ConfirmAction, Trusted command system with in Caster.  

Out of curiosity, I'd be very interested to know with my original implementation if removing `t.join()` fixes blocking the main thread.

LexiconCode:
Thank you for testing and even more so your feedback. As we talked about in gitter we will move `TerminalCommand`  into dragonfly. Then I will Implement customizations utilizing ConfirmAction, Trusted command system with in Caster.  

Out of curiosity, I'd be very interested to know with my original implementation if removing `t.join()` fixes blocking the main thread.

PR #326, opened 2018-11-21
Add complete spark formatting functionality

mrob95:
Inspired by #278. A few points:

* If the user just says "spark", only ctrl-v is executed. This is important to make sure that the command still works as expected for images, files etc.
* Whether or not "bow" is needed as a closer is questionable, I've left it in for now.
* The original (ages ago) suggestion for this feature suggested that the command should actually edit the clipboard in situ. I personally think this is a better solution as it preserves the original data.

Sample command: "shackle cut spark yell bow". Capitalise the whole line.

EDIT: I tried changing this to use Text() instead of copying in and out of the clipboard. It's more elegant but unfortunately takes ages (potentially minutes) to execute long strings, so I think the clipboard is the better option.

### DIFF
--- a/caster/lib/ccr/core/nav.py
+++ b/caster/lib/ccr/core/nav.py
@@ -75,12 +75,15 @@ class NavigationNon(MappingRule):
               rdescript="Mouse: Ctrl + Left Click"),
         "garb [<nnavi500>]":
             R(Mouse("left") + Mouse("left") + Function(
-                navigation.stoosh_keep_clipboard, nexus=_NEXUS),
-                rdescript="Highlight @ Mouse + Copy"),
+                navigation.stoosh_keep_clipboard,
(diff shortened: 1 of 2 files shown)


LexiconCode:
I'll get back to you I think there's an alternative to `Text()` which is much faster.

LexiconCode:
The following method is faster. [ `Key`](https://github.com/dictation-toolbox/dragonfly-scripts/blob/master/_programming_help.py#L43) could be used for long strings. Given the context of its utilization here I'm not sure if it's much help.

mrob95:
This is fine really, it works quickly and reliably even with big strings (documents, for example). It just feels like a bit of a hack.

LexiconCode:
Would you consider this complete?

mrob95:
Yep, should be ready to merge

PR #328, opened 2018-11-23
Fixed bug introduced in #326

Versatilus:
When grammars are combined for CCR the `extras` get overwritten by the latest one as they are merged.  This behavior introduced a bug in the last merge which caused the spoken alphabet to wrap the letter in `SHIFT` keystrokes.
In this case, the `Choice` extra "big" existed several times for the same purpose but was implemented slightly differently. I made these uses consistent and simplified things.
I also refactored part of the code for the spoken alphabet to handle capital letters in a more platform neutral way.
I left `letters` in `alphanumeric.py` as it was because I couldn't find any use of it in the code base. I think the function might be an artifact from when @synkarius was originally developing Caster.

### DIFF
--- a/caster/lib/alphanumeric.py
+++ b/caster/lib/alphanumeric.py
@@ -79,10 +79,10 @@ def letters(big, dict1, dict2, letter):
     d1 = str(dict1)
     if d1 != "":
         Text(d1).execute()
-    if str(big) != "":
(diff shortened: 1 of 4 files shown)


LexiconCode:
If you like you can merge otherwise I'll do it when I get home.

Issue #327, opened 2018-11-21
Caster GUI settings window does not launch


LexiconCode:
Caster has a GUI for settings does not launch. To test use use use Spec "launch settings window"
```
ERROR:action.exec:Execution failed: Timeout while waiting for window context: WaitWindow(title='caster settings window v 0.5.11')
ERROR:action.exec:Execution failed: Failed to find window (title='Caster Settings Window v 0.5.11').
```

LexiconCode:
Works prior to  Add default format and secondary format (#262) `

mrob95:
I'm getting this on my normal branch, which is way behind #262. Increasing the time out didn't fix it.

LexiconCode:
Remember to delete the settings.toml

mrob95:
Still getting the same thing in both the `master` and `develop` branches

LexiconCode:
![capture](https://user-images.githubusercontent.com/24551569/48860656-201dec80-ed87-11e8-974c-9d1a974976eb.JPG)
1.   It works after I deleted Caster and re-cloned. 
       Reverted to 7ddf6409e30d76b5f5736493efbbdf95a582dbb1
       deleted .toml files in data directory
- UI Works as expected
2. Reset branch to commit #262 
- UI Fails with the error above.
3. Reverted to 7ddf6409e30d76b5f5736493efbbdf95a582dbb1 commit without deleting .toml from data
- UI Fails with the error above.
4. Still using 7ddf6409e30d76b5f5736493efbbdf95a582dbb1 but deleted all time all .toml files
- UI Works as expected

Therefore I'm left with the conclusion that UI cannot parse some of the new fields.

LexiconCode:
I cant imagine what would be causing it to be different on both of our machines.

mrob95:
This is what i'm seeing on the master branch. Manually modifying the settings files has fixed it on my main branch though... I am confused.

![settings error](https://user-images.githubusercontent.com/42875462/48862369-1794d880-edbe-11e8-96d2-a60a8ea01880.jpg)

LexiconCode:
Okay I'm not sure what to say but it works on the main branch for me as well. Hmmm it's going requires some thought. In addition a few other people to test would be helpful to see a trend.

comodoro:
[...]

PR #329, opened 2018-11-24
Settings window modified to distinguish between string lists and integer lists.

Versatilus:
This is a simple extension of the existing functionality in order to address #327. A proper solution would require rewriting much of the settings window code.


LexiconCode:
Should we wait to merge if there is a change in format to descriptive strings rather than plain integers?

Versatilus:
I think we might as well merge it in case we need the functionality for something else. I mostly do code changes as pull requests just to get feedback.

Issue #211, opened 2018-03-20
Consolidating command phrases from applications into core CCR and Application Core


LexiconCode:
I have been working on the API and reviewing many of the commands across different editors. There are identical commands across editers with varying degrees of implementation that are app specific. These are seen both at the castor core and supported applications. 

The idea is to have one command phrase per command set. Yet still preserve AppContext per application for each implementation not creating a global grammar.

This approach has some advantages.
- Enforces standard command phrases across all applications.
- Provides easy customization by end-user one command phrase per command set utilizing filter rules.
- Command phrases that are in for example `atom.py` are unique to that application.
- Eliminates duplicate command phrases conflicts.
- Simplifies documentation 

An example of duplicate commands that have the same functionality across different applications.
```
"next tab [<n>]":     R(Key("c-f6"), rdescript="Eclipse: Next Tab") * Repeat(extra="n"),
"next pane":          R(Key("cs-p") + Text("Window: Focus Next Pane") + Pause(atom_palette_wait) + Key("enter"), rdescript="Atom: Next Pane"),
"nexta [<n>]":        R(Key("c-pgdown"), rdescript="Visual Studio Code: Next Tab") * Repeat(extra="n"), 
"next tab [<n>]":     R(Key("c-tab"), rdescript="Foxit Reader: Next Tab") * Repeat(extra="n"),
"next tab [<n>]":     R(Key("c-tab"), rdescript="SSMS: Next Tab") * Repeat(extra="n"),
"next tab [<n>]":     R(Key("ca-pgdown"), rdescript="Visual Studio: Next Tab") * Repeat(extra="n"),
```

Let's move these out of their individual applications into ApplicationCore.py or CCR core. Application Core `.\caster\apps\ApplicationCore.py` could be a file created containing the commands that don't fit within CCR core. Commands that don't have duplicate functionality or that are unique a application still reside within their <ApplicationName>.py

Very pseudo-representation of  a `command set` in `ApplicationCore.py` 
```
"next tab [<n>]":  {
          R(Key("c-f6"), rdescript="Eclipse: Next Tab") * Repeat(extra="n"),
          R(Key("cs-p") + Text("Window: Focus Next Pane") + Pause(atom_palette_wait) + Key("enter"), rdescript="Atom: Next Pane"),
          R(Key("c-pgdown"), rdescript="Visual Studio Code: Next Tab") * Repeat(extra="n"), 
[...]

PR #333, opened 2018-12-01
Update Atom Command Rules

Versatilus:
After losing several hours to 'misfeatures' last night, I decided to spend some time refactoring this module. 
* Simplified and updated available commands.
  I added two helper functions to make adding various types of commands more streamlined. Some broken commands were fixed and some of the trigger phrases were rewritten to fix ambiguities (`"git (custom|run)"` in place of `"git custom|run"`, for example).
* Removed commands which duplicated Caster Core functionality.
  There were several commands which use nonstandard utterances to duplicate functionality provided in the core of Caster. Additionally, the `"paste"` command in particular would trap me for several minutes at a time due to the potential for very high repeat counts.
* Miscellaneous minor bug fixes.
  This includes fixing the `atom_palette_wait` option so that it refers to an integer.

### DIFF
--- a/caster/apps/atom.py
+++ b/caster/apps/atom.py
@@ -3,462 +3,408 @@ __author__ = 'LexiconCode'
 Command-module for Atom
 Official Site "https://atom.io/"
 """
-from dragonfly import (AppContext, Dictation, Grammar, IntegerRef, Key, MappingRule,
-                       Pause, Repeat, Text)
-from dragonfly.actions.action_mimic import Mimic
-
 from caster.lib import control, settings
 from caster.lib.dfplus.additions import IntegerRefST
 from caster.lib.dfplus.merge import gfilter
 from caster.lib.dfplus.merge.mergerule import MergeRule
 from caster.lib.dfplus.state.short import R
+from dragonfly import AppContext, Dictation, Grammar, Key, Pause, Repeat, Text
 
 # How long to wait for the Atom palette to load before hitting the enter key
-atom_palette_wait = "30"
+atom_palette_wait = 30
 if settings.SETTINGS["miscellaneous"]["atom_palette_wait"]:
-    atom_palette_wait = settings.SETTINGS["miscellaneous"]["atom_palette_wait"]
+    atom_palette_wait = int(settings.SETTINGS["miscellaneous"]["atom_palette_wait"])
 
 
-# Utilizes the Palette UI to leverage commands.
-def palettized(function_text):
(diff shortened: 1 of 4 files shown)


caster/apps/atom.py
```diff
@@ -3,462 +3,408 @@
...
-            R(palettized("Toggle Soft Wrap"), rdescript="Atom: Toggle Soft Wrap"),
-        ##"toggle command palette":                  R(Key(""), rdescript="Atom: Toggle Command Palette"),
+            ACP("Toggle Soft Wrap", "Atom: Toggle Soft Wrap"),
         "[toggle] treeview":
-            R(Key("c-backslash"), rdescript="Atom: Toggle Treeview"),
+            ACK("c-backslash", "Atom: Toggle Treeview"),
         #Panes Submenu
         "split above":
-            R(palettized("Pane: Split Up"), rdescript="Atom: Split Up"),
+            ACP("Pane: Split Up", "Atom: Split Up"),
```
mrob95: The "split" commands all have keyboard shortcuts - ctrl-k, up/down/left/right

caster/apps/atom.py
```diff
@@ -3,462 +3,408 @@
...
-            R(palettized("Find and Replace: Replace Next"),
-              rdescript="Atom: Replace Next"),
+            ACP("Find and Replace: Replace Next", "Atom: Replace Next"),
         "find replace all":
-            R(palettized("Find and Replace: Replace All"), rdescript="Atom: Replace All"),
+            ACP("Find and Replace: Replace All", "Atom: Replace All"),
         "find buffer":
-            R(Key("c-b"), rdescript="Atom: Find Buffer"),
+            ACK("c-b", "Atom: Find Buffer"),
         "find file":
```
mrob95: Perhaps "(find | go to) file" to be consistent with the other editors.

caster/apps/atom.py
Versatilus: They both don't quite do the same thing. Those keyboard shortcuts split the pane,  then duplicate the currently open tab. The existing voice commands create blank panes. I'll gladly change it, though, because I think that behavior is more intuitive.

mrob95: Ah okay, entirely up to you then :)

caster/apps/atom.py
Versatilus: I'll fix it, but it should eventually be handled by #211.

LexiconCode:
Looks ready to merge to me.

PR #316, opened 2018-11-03
Add partial Aenea support for Caster

Danesprite:
Re: #312.

This works using a new 'use_aenea' setting, set to false by default. If it is set to true, Aenea's `Key`, `Text` and `Mouse` actions will be used instead of dragonfly's. Aenea's `ProxyAppContext` class will also be imported as `AppContext`, as the two have compatible arguments.

To keep this clean, I've modified dragonfly imports for `Key`, `Text` and `Mouse` to import from `caster.lib.actions` and `dragonfly.AppContext` to be imported from `caster.lib.context`. I've not changed a few grammars like Dragon and Explorer which shouldn't use proxy actions/contexts.

If Aenea is unavailable, an error message will be printed and Caster will fallback on dragonfly's actions and context classes.

Let me know if there are any changes I should make. I don't think this should be merged just yet because I'd like to add what @tbenst suggested [here](https://github.com/synkarius/caster/issues/312#issuecomment-435262251).

PR #335, opened 2018-12-01
Add fman support

mrob95:
fman is a lightweight, two-pane file manager which can pretty much entirely be controlled using keyboard commands.

Note 1: Actions which use standard keypresses (return to open a file, tab to change pane, space to select files, etc) don't have their own commands.

Note 2: This imports Text and Key from .lib.actions in readiness for #316 .

### DIFF
--- a/caster/apps/fman.py
+++ b/caster/apps/fman.py
@@ -1,0 +1,91 @@
+from dragonfly import (Grammar, Pause, Choice)
+
+from caster.lib.actions import Key, Text
+from caster.lib.context import AppContext
+
+from caster.lib import control, settings
+from caster.lib.dfplus.additions import IntegerRefST
+from caster.lib.dfplus.merge import gfilter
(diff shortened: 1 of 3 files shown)


Versatilus:
Both this pull request and #334 introduced subtle errors and neither module could be imported. In the case of this patch, the problem was that the file implementing `caster.lib.actions` wasn't yet accessible because that patch hadn't been merged yet. In #334 it was due to a `__init__.py` file not existing.

I believe this happened because Python will load a `.pyc` cache file even if the original source file isn't there. The problem cropped up for me because I hadn't yet tried out the new Aenea feature branch.

I've been bitten by this before. It's just a reminder of the importance of deleting those files whenever we switch branches. I believe git allows this to be automated somehow, but I'm unfamiliar with the process.

LexiconCode:
Good catch. I'll make a simple script to clear out  `.pyc` for testing purposes.

Issue #330, opened 2018-11-25
Consider making the commands in gitbash.py safer?


kendonB:
Currently there are several commands in there which are annoying to reverse. I would remove *all* the enter commands from this script. Thoughts?

kendonB:
I think I would also add "git|get" as a prefix to all of the git voice commands to ensure that the user is making those commands intentionally.

Versatilus:
This has been a pet peeve of mine for a very long time, but I haven't taken the time to fix it. Go for it!

LexiconCode:
Another aspect that might be helpful is using ConfirmAction which uses a GUI dialogue to confirm or disconfirm a command before it executed.

LexiconCode:
[NodeRule](https://caster.readthedocs.io/en/latest/caster/doc/readthedocs/NodeRule/) can be used to make commands safer and less error-prone to recognition as well.

LexiconCode:
Is this something you plan on implementing @kendonB?

kendonB:
I can do this in a couple of days if no one else gets to it. Cheers

kendonB:
I have made my initial suggested changes as well as switching to quotes rather than apostrophes in the commit command. I also removed the -a command (which automatically adds modified and deleted files), which I think is a risky option by default.

PR #340, opened 2018-12-13
making gitbash.py safer

kendonB:
Begins to address #330

### DIFF
--- a/caster/apps/gitbash.py
+++ b/caster/apps/gitbash.py
@@ -9,12 +9,10 @@ Command-module for git
 """
 #---------------------------------------------------------------------------
 
-from dragonfly import (Grammar, Mimic, Function)
+from dragonfly import (Grammar, AppContext, Mimic, Key, Text, Function)
 
 from caster.lib import control
 from caster.lib import settings
-from caster.lib.actions import Key, Text
-from caster.lib.context import AppContext
 from caster.lib.dfplus.additions import IntegerRefST
 from caster.lib.dfplus.merge import gfilter
 from caster.lib.dfplus.merge.mergerule import MergeRule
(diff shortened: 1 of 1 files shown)


caster/apps/gitbash.py
```diff
@@ -9,12 +9,10 @@
 """
 #---------------------------------------------------------------------------
 
-from dragonfly import (Grammar, Mimic, Function)
+from dragonfly import (Grammar, AppContext, Mimic, Key, Text, Function)
```
Versatilus: `AppContext`, `Key`, and `Text` should now be imported from `caster.lib.context` and `caster.lib.actions` to support optional use with `Aenea`.

caster/apps/gitbash.py
```diff
@@ -30,86 +28,86 @@ class GitBashRule(MergeRule):
     pronunciation = "git bash"
 
     mapping = {
-        "initialize repository":
-            Text("git init") + Key("enter"),
-        "add":
-            R(Key("g, i, t, space, a, d, d, space, dot, enter"),
+        "git|get base":
```
Versatilus: I believe that the grammar parser turns this into `git` and `get base` instead of `git base` and `get base` as intended. I recommend using parentheses for explicit grouping, such as `(git | get) base`.

Other than that, I love all of these changes.

caster/apps/gitbash.py
kendonB: Can we put these into another issue? This problem appears to be in many scripts and it'll be tidier to do them all at once, along with any more that need to be switched.

caster/apps/gitbash.py
kendonB: I think I fixed these all up now

caster/apps/gitbash.py
Versatilus: You make a fair point. We'll fix it later.

Issue #153, opened 2016-06-17
The termination word "cancel" is already taken by Dragon


synkarius:
I used to be able to use `cancel` as a termination word for the `ross fast`/etc. AsynchronousAction and other AsynchronousActions. Now I realize that I was only able to use it because I'd turned that command off in the Dragon settings. (Saying `cancel` causes Dragon to press the escape key.)  Fortunately, (thanks @Zone22 ) there is an alternative termination word that I put in a while back for use with WSR which also works for Dragon's default settings: `escape`. I'm leaving `escape` and replacing `cancel` with `terminate`.

PR #342, opened 2018-12-14
change comment from cancel to terminate

kendonB:
I'm not sure if the "cancel" in quotes should also be changed/removed

### DIFF
--- a/caster/lib/ccr/core/nav.py
+++ b/caster/lib/ccr/core/nav.py
@@ -164,7 +164,7 @@ class Navigation(MergeRule):
     pronunciation = CCRMerger.CORE[1]
 
     mapping = {
-    # "periodic" repeats whatever comes next at 1-second intervals until "cancel" is spoken or 100 tries occur
+    # "periodic" repeats whatever comes next at 1-second intervals until "terminate" is spoken or 100 tries occur
         "periodic":
             ContextSeeker(forward=[L(S(["cancel"], lambda: None),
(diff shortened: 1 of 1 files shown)


Versatilus:
I'll check that out after dinner.

Versatilus:
Technically, it runs until `SymbolSpecs.CANCEL` is said.  I'm still not sure if "cancel" should be removed. There's a lot of code here to unwind. I'll keep on staring at it tonight.

LexiconCode:
I'm not exactly sure why we would change it from "cancel" to "terminate". I would think the average person would think of "cancel" before "terminate" to stop an action.

Versatilus:
> I'm not exactly sure why we would change it from "cancel" to "terminate". I would think the average person would think of "cancel" before "terminate" to stop an action.

Unintuitively, "cancel" doesn't cause the action because the word isn't part of the spec defined by `SymbolSpecs.CANCEL`. Now that you have pointed this out it makes me think the right change is to change the spoken words  to include "cancel".

kendonB:
Ref #153

LexiconCode:
> 
> 
> Ref #153
I'll do some testing but I think dragonfly or natlink now overrides built-in Dragon commands.

Versatilus:
> I'll do some testing but I think dragonfly or natlink now overrides built-in Dragon commands.

That's been my experience for a while now. If others can confirm this experience , I think we should simply revert the utterance.

kendonB:
I must have inadvertently deleted the branch - closing here

Issue #354, opened 2018-12-16
Weird bug when I try to replace "fly" 


kendonB:
When trying to replace the word "fly" in `words.txt`, I get an odd error. My `words.txt`:

```
<<<NOT_SPECS>>>
fly -> new
```

```
UserDirectory: C:\Users\kendon\Documents\caster
UnimacroDirectory: C:\NatLink\Unimacro
start of natlinkstartup
Vocola not active
Global rule filter from file 'words.txt' activated ...
Error loading _caster from C:\Users\kendon\Documents\caster\_caster.py
Traceback (most recent call last):
  File "C:\NatLink\NatLink\MacroSystem\core\natlinkmain.py", line 317, in loadFile
    imp.load_module(modName,fndFile,fndName,fndDesc)
  File "C:\Users\kendon\Documents\caster\_caster.py", line 152, in <module>
    gfilter.run_on(bring_rule)
  File "C:\Users\kendon\Documents\caster\caster\lib\dfplus\merge\gfilter.py", line 231, in run_on
    spec_override_from_config(mp)
  File "C:\Users\kendon\Documents\caster\caster\lib\dfplus\merge\gfilter.py", line 221, in spec_override_from_config
    rule._mcontext, rule._mwith)
TypeError: __init__() takes at most 7 arguments (11 given)
-- skip unchanged wrong grammar file: C:\Users\kendon\Documents\caster\_caster.py
-- skip unchanged wrong grammar file: C:\Users\kendon\Documents\caster\_caster.py
natlinkmain started from C:\NatLink\NatLink\MacroSystem\core:
  NatLink version: 4.1uniform
  DNS version: 15
  Python version: 27
  Windows Version: 7

User defined macro's (UserDirectory) enabled: C:\Users\kendon\Documents\caster
----------------------------------------

```

I do not get the same error when trying to replace `shin` or `queue`.

comodoro:
I can reproduce. BringMeRule is SelfModifyingRule and has fewer constructor arguments; gfilter assumes MergeRule. Note that this is not specific to bring me. I can reproduce with "alias" added to word.txt <<<SPEC>>> for example.

Quick fix would probably be adding instance check to gfilter. I am not sure what to do with it long term as I mostly do not use filter rules.

kendonB:
Both should be replacing words only IMO.

synkarius:
[...]

PR #356, opened 2018-12-18
[WIP] Fix for words.txt and filter rules

comodoro:
For #354, starting with the quick fix

### DIFF
--- a/caster/lib/dfplus/merge/gfilter.py
+++ b/caster/lib/dfplus/merge/gfilter.py
@@ -3,9 +3,10 @@
 
 import io
 import os
-
+import traceback
 from caster.lib import settings
 from caster.lib.dfplus.merge.mergepair import MergeInf, MergePair
+from caster.lib.dfplus.merge.selfmodrule import SelfModifyingRule
 from dragonfly.grammar.elements import Choice
 
 
@@ -131,7 +132,8 @@ if os.path.isfile(settings.SETTINGS["paths"]["FILTER_DEFS_PATH"]):
             DEFS = GlobalFilterDefs(lines)
         except Exception:
             print("Unable to parse words.txt")
-
+            traceback.print_exc()
+            
 
 def spec_override_from_config(mp):
     '''run at boot time only: changes are permanent'''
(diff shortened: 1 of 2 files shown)


LexiconCode:
Do you consider this complete?

comodoro:
I am not sure how is the replacement supposed to work, I commented on the underlying issue (#354), maybe I should comment here as well

LexiconCode:
Experimenting with the word.txt global simplified filter rules with your pull request.

Testing **selfmodifying** rules
These have two seem to have no effect
```
<<<SPEC>>>
alias -> gumption
```
```
<<<NOT_SPECS>>>
alias -> gumption
```

Testing **MergeRule**
Does not work but it does not work in the development branch either
```
<<<NOT_SPECS>>>
nexta -> cat
```

Works as expected
```
<<<SPEC>>>
nexta -> cat
```
I'll have to look into it further but let me know what you think.

comodoro:
@LexiconCode Thanks for pinging me. The pull request does not contain the modifications for the replacement logic yet, only fix for the exception mentioned in the issue. The alias command (and chain alias) are a bit special in that the specs are hardcoded and reset with init and each change. Besides there is that it is being loaded twice (_caster.py and alias.py), I am not sure where we landed on that.

So I think we are agreed on replacement of words(and phrases), but it might be worth considering also supporting whole specs (commands) besides that, like it is now.

LexiconCode:
I apologize, thanks for clarifying. It might be worth considering utilizing toml instead of txt. In the future pull request we could add modifications by voice.

At this for replacing specs the logic is in place already in [scen4.py](https://github.com/synkarius/caster/blob/bc5b2b8bd4a2495c910a4beb4260d50847f922a9/caster/user/filters/examples/scen4.py#L34) I thought about making its own toml file for complete spec replacement. That was my thought but it's worth discussing which is a better approach.

comodoro:
That is a good point, toml is readable, and makes things easier. A might be perceived disadvantage might be having to quote spaces and Unicode. Voice modifications seem possible.

Complete spec replacement is currently there
https://github.com/synkarius/caster/blob/bc5b2b8bd4a2495c910a4beb4260d50847f922a9/caster/lib/dfplus/merge/gfilter.py#L153-L156
I think easiest currently would be adding a new header for words and phrases and altering current extras replacement from substrings to words and phrases. The other things can be addressed in future.

LexiconCode:
@comodoro are you okay with me closing this issue or should we keep it open?

comodoro:
OK, This is too old, I will review the issue and possibly create a new one.

PR #359, opened 2018-12-19
Update installation instructions

mrob95:
A couple of modifications to the installation process/instructions:

* Changed the order so that all dependencies are installed at the same time using the .bat file.
* Added instructions for modifying natlink settings using the GUI.
* Added "future" to the dependency install, this seems to be part of NatLink's move to python 3.

I'm open to suggestions for anything else which isn't clear or any more potential issues.


Versatilus:
Personally, I would throw a `-U` flag on all of them and install/upgrade `pip` first.

I'm happy to see `future` installed. I have a significant block of code in the works that will be made a lot cleaner with that module.

PR #360, opened 2018-12-24
 A dynamically reloaded grammar file in user space.

LexiconCode:
I get the impression that many people experiment within the source code of Caster. This is going to be an issue when Castor becomes a package or currently during updates. 

Primary goals
- Keep it simple - We don't want to overwhelm newcomers.
- Provide a safe area for users to experiment with commands and functions without having to reboot Dragon. 
- Include functional examples demonstrating a variety of implementation.

Currently it works as advertised but is still a work in progress. I would like feedback and community in regard to what examples to include. This is be active by default but in the future it will include a welcome message guiding the user to utilizing the file.

Current thoughts for a name is playground but that's up for debate. This is based off the trim down version of Dev and Devgen. 

The `configdebug.txt` which is utilized by Dev and Devgen contain syntax errors.
I've effectively combined and simplified `Dev `and `Devgen ` in playgroundhelper.py and put a new file called `playground.txt` located in the user directory `caster\user\playground.txt` playground.txt generates a ccr global a grammar.

**Getting Started**
Make your edits to `playground.txt` then say the `rebuild playground` command. If there is an error it will print out in the natlink window. Simply rerun the command as needed.
I've not had a chance to test  `run remote debugger` command.

### DIFF
--- a/caster/lib/settings.py
+++ b/caster/lib/settings.py
@@ -152,6 +152,7 @@ _DEFAULT_SETTINGS = {
 
         # CCR
         "CONFIGDEBUGTXT_PATH": BASE_PATH + "/bin/data/configdebug.txt",
+        "PLAYGROUNDTXT_PATH": BASE_PATH + "/user/playground.txt",
 
         # PYTHON
         "PYTHONW": "C:/Python27/pythonw",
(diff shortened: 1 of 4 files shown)


LexiconCode:
This issue will be closed with to move configdebug.txt  Into the user's place in the `restructure` branch

PR #362, opened 2018-12-28
Minor core modifications

mrob95:
* Simplify a few commands by removing unnecessary calls to functions in other files (for example, the number command).
* Moved the window management commands which were in _caster.py into nav.py, so they are all in the same place.
* The window snapping command now allows multiple direction commands (eg window lease sauce to put a window in the top left corner). 
* Added commands for working with Windows workspaces, it's frustrating that there isn't a key binding for moving a program from one workspace to another but I still find this feature fairly useful if I'm working on multiple things and need to prevent distractions.
* Added "command mode" command to dragon.py for completeness.
* Added the above and the new "bring me" functionality to the quick reference. 

The nav.py file is getting rather large and seems to be a bit of a dumping ground for a lot of different things. Should it be split up into a few smaller modules? Mouse, clipboard, window management and text manipulation are breakpoints which spring to mind.


mrob95:
I'll start planning it out, but probably won't make a move until master has been updated

PR #365, opened 2019-01-03
Choice from toml files

mrob95:
Utility function to allow the import of Choice elements from one or more toml files. e.g.
```
extras = [
    Choice_from_file("symbol", "caster/ccr/latex/latex.toml"),
]
```
where the toml file looks like:
```
[symbol]
"symbol one" = "symbol one"
...
```

At the moment I'm just thinking about using this to de-clutter a couple of the language modules, but potentially it could also be used for importing from user space, simply by passing a list of paths rather than just one.

### DIFF
--- a/caster/lib/utilities.py
+++ b/caster/lib/utilities.py
@@ -18,6 +18,7 @@ from _winreg import (CloseKey, ConnectRegistry, HKEY_CLASSES_ROOT,
     HKEY_CURRENT_USER, OpenKey, QueryValueEx)
 
 from dragonfly.windows.window import Window
+from dragonfly import Choice
 
 try:  # Style C -- may be imported into Caster, or externally
     BASE_PATH = os.path.realpath(__file__).rsplit(os.path.sep + "caster", 1)[0]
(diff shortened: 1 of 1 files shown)


LexiconCode:
After giving this some thought. I have a few reservations about moving core code out of .py file format. Mainly because we would lose a lot of the power of an IDE and linting that applies to the Python language but not toml. I know that `choices `is as simple as it gets but I'm thinking for the future as we might add other elements eg MergeRules. Perhaps split it up into choices.py?

What do you think of this approach. Alternative be to read Choice from .py files and write generate to Choice toml (latex.toml) in the user directory. Changes in latex.toml overrides the settings choices.py.

First run or run on update latex.toml would be modified. Much of the groundwork for this is actually already in setting.py.

mrob95:
Sorry, have been busy the last couple of days. I'm happy to structure it as you'd like. The more I use my setup, which is almost entirely based on toml files now, the more I like it for providing an easy and robust interface for adding new commands. I guess it doesn't really matter whether those files come bundled with caster or whether they are generated from source on the first start-up.

LexiconCode:
A question I have I'm going down this path. Say a user wants to create their own private functions in Python file in the user space. Then include wants them in (latex.toml)? So comes down to a question of imports.

mrob95:
Hmmm, It could be done but might be tricky. I think you would have to have a .py file in user space containing the functions, the mappings, extras and defaults. Then the main file would update itself to include those on start-up.

You can get a surprising amount done just by using Choice elements though.

PR #366, opened 2019-01-09
Add basic LyX functionality

mrob95:
App rule for LyX. This may be added to over time but it's a start. I'm planning to document all of the mathematics functionality together separately.


mrob95:
Looks good, I completely forgot to add functionality for just normal document production (author, sections etc). Will add this to my to-do list.

alexboche:
(probably you know this and that's why you wrote what you wrote but I am finding your post ambiguous so just to make sure:) that file has commands for that e.g. title author paragraph and so on

PR #367, opened 2019-01-09
Basic commands for enabling and disabling CCR

mrob95:
With my limited knowledge of ccrmerger.py this is the best way of implementing this I could think of. Someone else may know a more elegant way. The basic behaviour is this:
* Saying "disable caster" disables all CCR rules. It doesn't disable app rules but I don't think this is too important.
* Saying either "enable caster" or enabling any CCR rule re-enables CCR.
* CCR remains disabled after reboot.

Main use cases are:
* Using caster for one specific thing (coding, maths etc) but not for everything.
* Disabling caster while you're doing a specific task (eg dictating text).


LexiconCode:
Added it the docs and I'll get it merged.

PR #370, opened 2019-01-13
Text normalization bug fix

Versatilus:
This was my solution to the issue addressed by @mrob95 in #368. I've been sitting on this for a while, as can be seen from the date in the branch name.

The tactic I took solving this was to simply apply the existing normalization to dictated text. There are several more extensive changes I would like to make in this area, but I thought this was an acceptable minimum for fixing the bug. I explicitly wanted to avoid messing with paste functionality due to that being part of the larger scope of changes I would like to make in the future.

The meaningful changes are:
 * Line 1 makes `str` Unicode. This is important to be able to handle the (currently optional) changes to the Windows implementation of `Text` in Dragonfly. This section also has some basic cleanup of the imports.
 * Line 34 is a simplification of `formatted_text` to remove an unnecessary length test.
 * Lines 72-79 turns `_normalize_text_format` into class method `normalize_text_format`.
 * Lines 87 and 88 convert `set_text_format` to use the new `normalize_text_format` class method.
 * Line 134 does the same for `master_format_text`.

Everything else is noise introduced from `yapf` processing.

I'm open to merging whichever pull request others are most satisfied with. #368 is a more obvious and straightforward solution. My code allows `format_text` to be called without forcing normalization.

### DIFF
--- a/caster/lib/textformat.py
+++ b/caster/lib/textformat.py
@@ -1,10 +1,7 @@
-import time
-import sys
+from builtins import str
 
-from caster.lib.actions import Key, Text
-from caster.lib.clipboard import Clipboard
-from caster.lib import settings, context
 from caster.lib import settings
+from caster.lib.actions import Text
 
 
 class TextFormat():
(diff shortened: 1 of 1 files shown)


mrob95:
Looks good, I will probably revert the first commit in #368 and leave the second, which deals with the effect of this fix on the spark command.

LexiconCode:
As for which implementation. I would be satisfied with whatever you guys decide.

Versatilus:
That sounds like a good compromise. The important thing to me is that the bug is fixed so that we can update `master`.

Issue #358, opened 2018-12-19
Enhancements to chain alias and alias command


mrob95:
As far as I can tell, since #245 the alias and chain alias commands both do basically the same thing - create a CCR alias for the highlighted text. Is there any reason not to remove one of them?

LexiconCode:
There are two differences with the chain version of the alias command. The first is that you don't have to speak the spec. A box will pop up and ask you for it. The second is that commands created with the chain version will be incorporated into the active CCR set, meaning, they'll be chainable with the other active CCR commands.

mrob95:
I've only tested it briefly but they both create chainable commands for me, so the only difference is the input method.

LexiconCode:
I can confirm that alias is chainable when it should not. It might be how is added to the nexus https://github.com/synkarius/caster/blob/bc5b2b8bd4a2495c910a4beb4260d50847f922a9/caster/lib/ccr/recording/alias.py#L102

@mrob95 
If the alias commands function as written in the documentation they would have enough utility as separate commands.

Otherwise we could create a single command with a conditional for manual input that uses chain alias GUI. Thoughts?

mrob95:
Personally I like the way it works at the moment, I use it for tricky variable/function names which would be a pain to dictate out each time, so I usually want the result to be chainable. I also don't really use the GUI - if you can't create it using a voice command then it's probably not a good alias in the first place. I may be missing some potential use cases though.

LexiconCode:
So let me break it down a little bit. 
1. Does the alias commands need to have a chainable and non-chainable version?
2. Does the GUI enhance alias commands functionality?

A few of my use cases were related to commands that aren't suitable for CCR and utilizing the GUI on top of other windows when inputting data. Great for when you're referencing multiple pages of documentation to create a single command/macro/function. 

My thought is to preserve the functionality as it was intended with a few enhancements.

The _alias_ would not be chainable. 
The _chain alias_ would be chainable
Both commands would optional spec parameter to trigger GUI.

You could switch the specs using filter rules. Once https://github.com/synkarius/caster/pull/356 is merge

mrob95:
That sounds good, I guess the GUI could be triggered if there is no dictation registered after the command. I'll have a look once the xmas chaos is over :-)

LexiconCode:
Great! Perhaps we should update the title to reflect the issue.

PR #373, opened 2019-01-18
Alias updates

mrob95:
Re #358 

* Removed CCR from regular alias
* Both regular aliases and chain aliases can now be added both with a voice command and through the GUI. "alias example command" will set the spec as "example command". "alias" will bring up the GUI.

Seems like an intuitive way of doing things but I'm happy to take suggestions.


LexiconCode:
Functions as expected :)

PR #393, opened 2019-02-24
Workspace management commands

mrob95:
Workspaces in Windows provide the ability to have separate desktops active at the same time, each with their own set of programs and windows open. Switching between workspaces is quite a lot easier than having every window fighting for space. I tend to have one workspace for my browsers, one for my reading, one for code, one for documents I'm preparing, etc.

Basic functionality (creating, closing, and tabbing between workspaces) is provided through standard Windows keystrokes, but unfortunately some fairly important functionality is missing. By default, in order to move a window to another workspace you have to drag and drop it in the workspace menu - impractical for those using dictation only.

Thankfully there is a workaround [here](https://github.com/Ciantic/VirtualDesktopAccessor) which allows for windows to be moved around with a dll call.

Everything is working as far as I can tell, but this probably could do with some testing.

PS. I have also added all of this, and the "bring me" commands to the quick reference.
PPS. I have also added a "command mode" command to dragon.py, for completeness.

If there are problems with this, then the most likely culprit is the lack of [VS 2017 runtimes](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads)


LexiconCode:
Perhaps we should remove the following and use your implementation "command mode" to toggle?
```Python
        'com on':
            R(Playback([(["command", "mode", "on"], 0.0)]),
              rdescript="Dragon: Command Mode (On)"),
        'com off':
            R(Playback([(["command", "mode", "off"], 0.0)]),
              rdescript="Dragon: Command Mode (Off)"),
```

LexiconCode:
Would it be possible to include home or default command to return to desktop?

mrob95:
This is ready to merge as far as I'm concerned, as long as you're happy that it's working and there's nothing else obvious which should be added. There is obviously more stuff which could be done with this functionality but I think it will probably depend on how individual users want to use it, so I'm happy to leave it up to them.

LexiconCode:
Once the merge conflicts have been resolved I can merge. Everything's great out until then on my end.

mrob95:
Okay I have added all of the changes which were made to the quick reference in #369 so it should be fine to ignore the conflicts. I had to start a new page on the quick reference but this was inevitably going to be necessary at some point, and gives us a lot more space.

LexiconCode:
I appreciate that!

PR #397, opened 2019-03-06
Create dragonfly_dev_commands.py

alexboche:
These are commands for creating dragonfly commands faster. Most of them should be CCR. I don't know where you want to put them. The file will require some tweaking since it was written purely for dragonfly not caster. Sorry I don't know exactly how to tweak it.
A couple of the more sophisticated commands are for making mimic and playback commands with dictation. Thanks to DaneSprite for helping a lot with this.

PR #398, opened 2019-03-07
implementing changes requested for dragonfly_dev_commands

alexboche:
Sorry I don't know how to make this part of the previous pull request so I think I'm making a new one here. 

In the future, it would be nice if we could extend the commands for creating mouse commands as follows to supplement the Sikuli commands. Here's how I think would work. We would have one big toml file for storing mouse positions in different applications. Then there would be a command that automatically records 1) the mouse position (absolute or relative), 2) the active application, 3) and a keyword for calling it. 
The command would automatically take all that information and place it inside the toml file so that so could then be called by the keyword (and "left" or "right") while in that application active inputs it into the toml file
thanks to lunis for this idea.


LexiconCode:
Let's try to include these files in your previous pull request.

Your previous pull request pas created on your `patch-2` Anytime you push new changes to that branch your pull request will be updated automatically!

open up `git bash`
1. `git checkout patch-2` <- loads all the files related to that branch `patch-2`
2. Then copy over the changes that you made on the files here.
3. `git push origin patch-2` <- push the changes which will update your branch and the pull request.

Open up https://github.com/dictation-toolbox/caster/pull/397 and see if the changes have been made.

I'll see if I can find a keyboard Git client that easy to use and create a grammar for it to make this easier for you.

PR #401, opened 2019-03-09
Updated VSCode commands

alexboche:
I'm adding a bunch of new commands from https://code.visualstudio.com/shortcuts/keyboard-shortcuts-windows.pdf and also a few from someone named Casper. In practice many of these will be disabled. Right now I have named them long names so that it is easy to see what they do – users will want to shorten them probably.


LexiconCode:
1. A bit of feedback when updating grammars. Make sure to preserve command names. In other words don't change command names unless you absolutely have to. This is because other people uses grammar already and are used to command set.  That minimizes the process for updating grammars.


2. Include in all of your commands `rdescript` this notifies the user that a command is been spoken to in in that link window. In addition it allows you to shorten the command significantly because description of what it does is in rdescript.

mrob95:
To be clear, are these commands for the Visual Studio IDE or the VSCode text editor?

alexboche:
Oh yeah this is VSCode. I didn't realize there was a difference between VSCode and Visual Studio IDE – I've never heard of the latter since I am very new to all this. So looks like I put these commands in the wrong file sorry. Will try to fix this soon

PR #405, opened 2019-03-12
Fixed "[dev] choice" command  in dragonfly_dev_commands.py 

alexboche:
I fixed the "[dev] choice" command which was not working properly at least not in VS code even though it was working before. May depend on the editor because I think the issue has to do with the automatic insertion of parable characters once you insert one side which may depend on the editor. Adding in the pause also seems to be important.


LexiconCode:
It looks like a lot of other things get mixed into this commit. Try rebasing on top of the latest development branch.

PR #409, opened 2019-03-14
add Microsoft Excel module

alexboche:
Apparently I accidentally closed my initial pull request to add the excel module. So I am making another one since I don't know how to reopen the initial 1 – this 1 is exactly the same as the other one. Apologies for my general incompetence using git.

### DIFF
--- a/castervoice/apps/excel.py
+++ b/castervoice/apps/excel.py
@@ -1,0 +1,164 @@
+#
+# This file is a command-module for Dragonfly.
+# (c) Copyright 2008 by Christo Butcher
+# Licensed under the LGPL, see <http://www.gnu.org/licenses/>
+#
+"""
+Command-module for Microsoft Excel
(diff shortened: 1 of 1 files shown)


LexiconCode:
Thank you!

Digital-ecologist:
Hi. I just upgraded to the latest version of developed today, and started experimenting with these commands. Thanks for implementing these! They're very nice to have. I did notice one unusual detail, at first, I couldn't get the range selection commands to work, then, after taking a look at the py file, I realized that command set was based on the NATO-phonetic alphabet, not the caster alphabet. Is there a strategic reason for this? It seems more intuitive to have all the caster modules use the caster alphabet.

LexiconCode:
Good to see you Digital-ecologist!

Yep something that didn't quite get finished in time. Eventually a better workaround for integrating the alphabet will be added later. So it's been converted over eb2b8cd14d2ce22324a5d0eac2669d18cb892dcd

PR #411, opened 2019-03-14
fix dev choice command and made 'dev' required as a prefix to avoid m…

alexboche:
…isfires (for real this time)


alexboche:
I messed up the previous pull requests for this. I think this 1 should be the final word for now.

LexiconCode:
I really appreciate your work this will be great and has so much potential to grow even more! Now that it includes caster and dragonfly commands would you mind if we change the name from `Dragonfly Dev Commands` to `Command Dev`?

Once this is merge I have some refinements with formatting and a few other small things as well.

PR #413, opened 2019-03-16
Development command modification + documentation

mrob95:
I've added documentation for the new Outlook, Excel and development grammars.

I've also made some modifications to the development grammar to make it a little more intuitive, and also removed the mouse movement command, which is not really related to development and is almost identical to the "curse" command in core navigation.


LexiconCode:
Thanks for taking the documentation up to the next level!

Issue #331, opened 2018-11-26
Update rdescript in CCR Caster commands with an identifier


LexiconCode:
Many of the applications include the application name as part of rdescript as a means of identification. Hopefully in the future we can include a category but for now to increase consistency across the project CCR should include the word unique identifier in the beginning of rdescript for core Caster spec.

For instance with apps
```Python
        "comment line":
            R(Key("c-slash"), rdescript="Sublime: Comment Line"),
```
Core CCR commands would look like...
```Python
        "hash tag":
            R(Text("#"), rdescript="Core: Hash Tag"),
```
```Python
        "move window":
            R(Key("a-space, r, a-space, m"), rdescript="Core: Move Window"),
```

PR #415, opened 2019-03-16
Added to "Core:" rdescript for the core Caster ccr files. 

LexiconCode:
Differentiates Caster commands from other commands sources such as Dragon. Thus brings caster core commands in line with the rest of the CCR standards regarding rdescript grammar identification.


LexiconCode:
Fulfills https://github.com/dictation-toolbox/caster/issues/331

PR #417, opened 2019-03-16
Updated vscode module

alexboche:
… relevant line to ccr/__init__.py, create vscode/__init__.py, and remove apps/vscode.py


mrob95:
Looks good! CCR is probably something which should be added for all editors, at least for commands which are regularly used.

I definitely think this should be in apps/vscode.py though rather than CCR.

LexiconCode:
This will need to be converted back to an app grammar. 
The following commands also need to be reevaluated. In part because the hard to remember because are not plain English and because there are hard or impossible to pronounce for some users.
        "surfy"
        "nerdy"
        "sperdy"
        "lerdy"

PR #423, opened 2019-03-20
Markdown CCR module

mrob95:
I've started trying to use markdown instead of latex for as much as possible (documents, slides, websites etc). It's much nicer.

### DIFF
--- a/castervoice/lib/ccr/markdown/markdown.py
+++ b/castervoice/lib/ccr/markdown/markdown.py
@@ -14,18 +14,23 @@ class Markdown(MergeRule):
     mapping = {
         "heading [<num>] [<dict>]":
             R(Function(lambda num, dict:
-                Text(("#" * num) + " " + str(dict).capitalize() + "\n").execute())),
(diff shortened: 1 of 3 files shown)


castervoice/lib/ccr/markdown/markdown.py
```diff
@@ -14,44 +14,49 @@ class Markdown(MergeRule):
...
+                rdescript="Markdown: table row <n>"),
         "table (break | split) <n>":
-            R(Function(lambda n: Text("---|"*(n-1) + "---").execute()) + Key("enter")),
+            R(Function(lambda n: Text("---|"*(n-1) + "---").execute()) + Key("enter"),
+                rdescript="Markdown: table (break | split) <n>"),
 
         "insert <element>":
-            R(Key("%(element)s")),
+            R(Key("%(element)s"),
+                rdescript="Markdown: insert <element>"),
```
LexiconCode: It never occurred to me to be able to include <element> in a rdescript

Issue #414, opened 2019-03-16
Post install script handle _caster incorrectly in some cases.


LexiconCode:
The post install script must be failing silently and does not place _caster on the desktop or natlink user directory as expected. Does not appear to be pip version dependent.

LexiconCode:
Other Known Issues
- The script is being triggered before resolving dependencies through PiP.
- Items correctly print counsel installed by the following `python setup.py install` but not through `pip install castervoice`

PR #425, opened 2019-03-22
Post install Fix

LexiconCode:
This pull request is to resolve https://github.com/dictation-toolbox/Caster/issues/414.

This pull request is open mainly for people to provide feedback or suggestions on implementation.


LexiconCode:
The Post install script is now being triggered _after_ resolving dependencies through PiP.
ToDo: Printing out messages during PIP install from post post_setup.py.

LexiconCode:
--verbose argument can be used with PIP to show the messages and I've updated the documentation to reflect that until the more advanced solution is found.

PR #427, opened 2019-03-24
Create an action wrapper for read_selected_without_altering_clipboard

mrob95:
Stores the currently highlighted text in a temporary variable, to be Retrieved after some other action. If no text was highlighted, an empty string will be stored.
Sample usage:
```
"find that": Store() + Key("c-f") + Retrieve() + Key("enter")
```

In order to enable use with web URLs, Store() takes a string, space, which will replace all space characters, and a bool, remove_cr, which if true will remove any newlines in the selection, to avoid them triggering the request early.
Sample usage:
```
"wikipedia that":
    Store(space="+", remove_cr=True) + Key("c-t") +
    Text("https://en.wikipedia.org/w/index.php?search=") +
    Retrieve() + Key("enter")
```

There are cases where you may want the same function to do different things depending on whether or not text was highlighted. The action_if_no_text and action_if_text arguments to Retrieve() are calls to Key() and allow this. For example, you may want to finish inside a set of brackets if no text was highlighted, but outside if there was text.
Sample usage:
```
"insert bold text":
    Store() + Text("\\textbf{}") + Key("left") + Retrieve(action_if_text="right")
```

So far I have only implemented this in a couple of chrome and sublime commands, but there are probably quite a few situations where it would be useful to do something with highlighted text (generally searching for it somewhere or wrapping it in other text).

### DIFF
--- a/castervoice/apps/chrome.py
+++ b/castervoice/apps/chrome.py
@@ -121,7 +121,13 @@ class ChromeRule(MergeRule):
             R(Text("/msg NickServ identify PASSWORD"),
               rdescript="IRC Chat Channel Identify"),
 
-        "google that": Store() + Key("c-t") + Retrieve() + Key("enter"),
+        "google that":
+            R(Store(remove_cr=True) + Key("c-t") + Retrieve() + Key("enter"),
+                rdescript="Chrome: google that"),
+
+        "wikipedia that":
+            R(Store(space="+", remove_cr=True) + Key("c-t") + Text("https://en.wikipedia.org/w/index.php?search=") + Retrieve() + Key("enter"),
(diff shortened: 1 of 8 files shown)


LexiconCode:
Very nice. This would be text-only therefore a few changes to the variables would make things a bit clear.  Although it's really up to debate though I tend to error more side of the explanatory.

Looking at a grammar I wouldn't know what I'm storing or retrieving at a glance.

I'll see if I can get this tested in the next few days.

mrob95:
Fair point, I wanted to keep the names short but you're right that it's not immediately obvious what they do just from reading the grammar. Up to you if you'd like to change them.

PR #428, opened 2019-03-25
Fix non-CCR for app rules

mrob95:
This has been bugging me for a while. 

Previously, declaring a non-CCR rule to be loaded at the same time as a CCR app rule added with `control.nexus().merger.add_app_rule` didn't work. As far as I can tell this is because the `non` property doesn't survive the `MergePair` process. Ideally it would be fixed there but I have never worked with that code so thought it better to leave it alone. This just stores the `non` before merging and restores it afterwards.


LexiconCode:
All right we might figure out a place to document that this is now possible and I'll merge.

PR #429, opened 2019-03-26
Tidy up phonetic alphabets

mrob95:
re #409 . I don't have Excel installed so if somebody could test then that would be great, it's a fairly obvious fix though so shouldn't have made any difference.


LexiconCode:
Yep everything works as expected. Tested on Excel application and casters basic alphabet.

PR #430, opened 2019-03-29
Post install enhancements with logging in new window during pip install

LexiconCode:
This opens up a new command prompt window during install providing instructions or showing any errors in the post install script.

An effective workaround for requiring the verbose argument during PIP install.

### DIFF
--- a/post_setup.py
+++ b/post_setup.py
@@ -1,5 +1,11 @@
 import os
 import urllib2
+import logging
+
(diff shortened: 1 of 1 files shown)


LexiconCode:
Accidentally merged automatically during when pushing a modification to this pull request to upstream instead of origin. However a review would still be prudent.

PR #434, opened 2019-03-31
Add commands and dictionaries for Python built-ins, keywords, and libraries

alexboche:
I have commented out large parts of the dictionaries to avoid the Brad grammar error for now. Hopefully a more sophisticated way of dealing with that error is on the way soon. Some of the dictionaries are larger than anyone would need so it shouldn't be too much of an issue.


LexiconCode:
Due to grammar complexity you could include those that are commented out as individual commands.

LexiconCode:
@mrob95 how should Choice elements be handled in the documentation? Especially with the large number elements a format doesn't come easily to mind.

alexboche:
No rush, but just a reminder about this pull request. Grammar complexity should not be an issue anymore since the latest updates of Dragonfly.

LexiconCode:
Sure could you add  "Python:" to "rdescripts"?

alexboche:
Okay. Can I also get rid of a couple commands like "yield" and "pie" that are one syllable commands and not very commonly used? These cause misfires in my experience. These would be replaced by the keyword commands that are prefixed by the word "word" or "keyword"e.g. "word yield".
It looks like they added on some extra commits by accident, so I'll probably just do a new pull request.

LexiconCode:
I would include them in a second pull request.

PR #436, opened 2019-04-02
VScode

alexboche:
Sorry, for some reason it isn't letting me push this onto the previous pull request. So I've had to make a new one. I recommend closing the other one #417. I renamed some of the commands to make them easier to remember/pronounce and I added commands for selecting in between the different types of parable characters inclusive and exclusive.


mrob95:
@LexiconCode

Couple of thoughts on App CCR rules.

1. What is RDP mode? What is it supposed to achieve? I've only tried to use it a couple of times and couldn't make head or tail of it. From what I can see it initialises all app rules as normal CCR rules to be enabled and disabled the usual way. Why would you want this? Seems like a major faff to be saying "enable visual studio code" every time you want to use your editor, what's wrong with app contexts?

2. I've been testing CCR app rules a lot over the past month or so and have transitioned pretty much all my language rules and a few of my app rules over to using them. Having for example the Python rule automatically be active when it detects ".py" context is much easier than manually enabling and disabling it. If an app rule would be improved by CCR, then `merger.add_app_rule` works well.

3. Example: all my git bash commands are CCR, so I don't have to pause between commands and dictation in for example "git new branch format testing" or "git push shock".

3. I don't think most app rules need to be CCR, and in some cases it can even get in the way.

4. I think it would be good to have one or two examples of this in Caster so that users can see that it works and how to set it up.

5. Fixing the lack of filters is trivial now that non's work with app rules - filters are automatically run on everything in the merger.

Overall then I'd say that I'm positive on CCR in app grammars, but it's a case-by-case thing and most CCR rules will probably need to have non-CCR elements to avoid misrecognition of commands which are rarely required.

LexiconCode:
>     1. What is RDP mode? What is it supposed to achieve? I've only tried to use it a couple of times and couldn't make head or tail of it. From what I can see it initialises all app rules as normal CCR rules to be enabled and disabled the usual way. Why would you want this? Seems like a major faff to be saying "enable visual studio code" every time you want to use your editor, what's wrong with app contexts?

I can see how it might be confusing but it's meant for Remote Desktop Protocol (RDP) applications.  Contexts aren't seen over remote desktops. Therefore this is a workaround to allow people to control the context manually.

LexiconCode:
>     2\. I've been testing CCR app rules a lot over the past month or so and have transitioned pretty much all my language rules and a few of my app rules over to using them.

Why would you need to transition language rules since there are already CCR? What difference does it make for those with your implementation?

LexiconCode:
>     4\. I don't think most app rules need to be CCR, and in some cases it can even get in the way.

I agree especially when you consider the more larger CCR increase for grammar complexity issues.

LexiconCode:
>     5\. I think it would be good to have one or two examples of this in Caster so that users can see that it works and how to set it up.

I agree

>     6\. Fixing the lack of filters is trivial now that non's work with app rules - filters are automatically run on everything in the merger.

so that eliminates gfilter.run_on the need for at the end of an app grammar?

> it is in every app rule, should probably be made into one function, possibly in `ccrmerger.py`?

By definition app grammar isn't CCR. Think how that changes the documentation. I'm not saying necessarily no but it is something to think about.

LexiconCode:
>     2\. Having for example the Python rule automatically be active when it detects ".py" context is much easier than manually enabling and disabling it.

Automatic CCR module and Command Mode switching
This actually existed as a feature previously. albeit implemented it bit differently.
https://github.com/dictation-toolbox/Caster/blob/c5074e2a8b7de348cbfa80a4686e7e2d5db2ed9a/caster/doc/readthedocs/Auto%20CCR%20and%20Cmd-Mode.md

We would have to go back and re-implemented through the commit history.

mrob95:
> I can see how it might be confusing but it's meant for Remote Desktop Protocol (RDP) applications. Contexts aren't seen over remote desktops. Therefore this is a workaround to allow people to control the context manually.

Gotcha. When I get time I'll format this PR correctly so that CCR works for the bits where it's worth having and RDP mode is still an option

LexiconCode:
@alexboche I think it's ready to merge. Any thoughts or suggestions with some of the changes we've made?

alexboche:
I want to make the following changes: it should press enter at the end of the settings and snippets commands, and I think the indent/outdent commands should begin with pressing home. I tried to push a new commit on the same branch that I originally made this pull request but it didn't seem to work, probably because I deleted the parent branch? Anyways, I don't know how to push more commits to this pull request especially while keeping current with the latest changes that you have made to the pull request. If you tell me how to do this, I can do it or if you want to just make the changes I said above yourself, that works too. Eventually I'd like to add some of the line manipulation commands that mRob added to sublime but I wouldn't wait for that.

mrob95:
Okay am going to merge this since it is most of the way there and I think the branches are starting to get slightly messy. It doesn't sound like the changes which need making a particularly urgent so I'd be happy to make another pull request sometime this weekend to get them done and also add the line manipulation functionality.

PR #437, opened 2019-04-02
Couple of new sublime rules

mrob95:
Mostly minor updates and I still need to document these.

Couple of commands of interest which could easily be replicated in other editors:

* `go to <dict> [<filetype>]` - allows for example "go to sublime pie" which will take you straight to sublime.py. Doesn't work for tricky filenames but makes navigating a fair bit simpler.
* action lines - for example "copy line thirty three" or "select line five to fifteen".

### DIFF
--- a/castervoice/apps/sublime.py
+++ b/castervoice/apps/sublime.py
@@ -1,4 +1,4 @@
-from dragonfly import (Choice, Dictation, Grammar, Repeat)
+from dragonfly import (Choice, Dictation, Grammar, Repeat, Function)
 
 from castervoice.lib import control
 from castervoice.lib import settings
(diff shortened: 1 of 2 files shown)


castervoice/apps/sublime.py
```diff
@@ -159,16 +194,24 @@ class SublimeRule(MergeRule):
...
+        "nn": None,
         "n2": 1,
         "n3": 1,
+        "filetype": "",
     }
 
 
 #---------------------------------------------------------------------------
 
-context = AppContext(executable="sublime_text", title="Sublime Text")
```
LexiconCode: why choose the window title over the executable?

castervoice/apps/sublime.py
```diff
@@ -10,6 +10,15 @@
 from castervoice.lib.dfplus.merge.mergerule import MergeRule
 from castervoice.lib.dfplus.state.short import R
 
+def action_lines(action, n, nn):
```
LexiconCode: Very clever. When Action Context becomes available. We can move this kind of function into AppStandard

castervoice/apps/sublime.py
mrob95: Well spotted, too much copy pasting!

PR #444, opened 2019-04-07
1) add text manipulation commands to nav.py with supporting functions in naviga…

alexboche:
…tion.py, 2) add miscellaneous keystroke navigation commands, 3) move some existing commands into the CCR rule from non-CCR rule of nav

### DIFF
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -19,7 +19,6 @@ from castervoice.lib.ccr.standard import SymbolSpecs
 
 _NEXUS = control.nexus()
 
-
 class NavigationNon(MappingRule):
     mapping = {
         "<direction> <time_in_seconds>":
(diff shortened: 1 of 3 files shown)


alexboche:
I'm going to add "change phrase to phrase" and also rewrite the supporting functions so that they don't alter the clipboard.

alexboche:
I just pushed another commit which adds a few new text manipulation commands like "change lease <dictation> to <dictation2>" and "remove lease <dictation>" and changed the text manipulation functions to use caster's clipboard functions (namely, context.paste_string_without_altering_clipboard and context.read_selected_without_altering_clipboard) instead of pyperclip. Unlike pyperclip, caster's clipboard functions don't alter the clipboard. Unfortunately, when I use caster's clipboard functions, the commands don't seem to work well in certain applications especially those that are Dragon-full-text-control. I guess that's okay because you can just use built-in Dragon commands for those. But if you use paperclip it does seem to work better in those applications e.g. Microsoft Word and SP dictation box. I'm guessing the issue might have something to do with the pause time in caster's clipboard functions. I don't think the issue is specific to the new commands I'm creating because 
previously existing commands like the hug commands also don't work in Microsoft word, but again I think this is mostly okay because Dragon already has nice commands like "quote that".


Unrelatedly, I've noticed that the tell dock/sink are too slow. Accordingly, as part of this latest commit, I also reduced the wait time in the function that powers those commands which is called next_line.

alexboche:
I find that the splat commands don't work in certain applications that don't allow you to press c-backspace and c-del such as SP dictation box in Notepad. I proposed a change which instead uses (cs-left, del). I also found occasionally that Dragon interprets splat lease/ross (which I believe is supposed to be a single command) as splat followed by lease/ross. I proposed making it splat (for left) and sprat (for right) unless I'm the only one experiencing this problem.

LexiconCode:
It will be a number of weeks before I can get this with finals coming up with academics.

alexboche:
I am closing this pull request because it contains too many disparate changes so I want to spread them out into separate pull requests.

PR #447, opened 2019-04-10
Change jump specs and change behavior of "jump in lease" to be the inverse of "jump out ross"

kendonB:
I'm not sure if this is a bug or not. I was trying these commands out and found that "jump back in" had odd behavior that would go to the start of the parentheses, when I expected it to jump in once it found a closed parentheses.

### DIFF
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -209,7 +209,7 @@ class Navigation(MergeRule):
             AsynchronousAction([L(S(["cancel"], context.nav, ["left", "(~[~{~<"]))],
             time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back"),
         "jump back in":
-            AsynchronousAction([L(S(["cancel"], context.nav, ["left", "(~[~{~<"]))],
+            AsynchronousAction([L(S(["cancel"], context.nav, ["left", ")~]~}~>"]))],
             finisher=Key("right"), time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back In" ),
 
     # keyboard shortcuts
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -210,7 +210,7 @@ class Navigation(MergeRule):
             time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back"),
         "jump back in":
             AsynchronousAction([L(S(["cancel"], context.nav, ["left", ")~]~}~>"]))],
-            finisher=Key("right"), time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back In" ),
+            time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back In" ),
 
     # keyboard shortcuts
         'save':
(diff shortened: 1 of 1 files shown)


LexiconCode:
Okay to clarify a few things so "jump back in" is exhibiting the same behavior as "jump back" which moves to start of the parentheses. Which editor are you experiencing this issue?
```
        "jump back in":
            AsynchronousAction([L(S(["cancel"], context.nav, ["left", ")~]~}~>"]))],
finisher=Key("right"), time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back In" ),
```

```
        "jump back":
            AsynchronousAction([L(S(["cancel"], context.nav, ["left", "(~[~{~<"]))],
time_in_seconds=0.1, repetitions=50, rdescript="Core: Jump: Back"),
```

kendonB:
"jump back" jumps out of the parentheses etc immediately to its left and is the inverse of "jump in"

"jump out" jumps out of the parentheses etc immediately to its right and should be the inverse of "jump back in". It currently doesn't work that way and goes to first open parenthesis etc. My version also still contains an error - it shouldn't include the finisher which I'll fix now.

We probably also want to consider allowing for "jump back out" as the spec . Then "back" always indicates going left. 

If I was doing this from scratch, I might also make it a bit more verbose and use ross/lease.

So the four commands would be "jump in ross", "jump out ross", "jump in lease", and "jump out lease". Should we consider making this change?

kendonB:
It would also be great if these combined with "shin" to help with selection.

LexiconCode:
Thanks for clarifying yes that does make sense. Feel free to move ahead with the spec changes.

kendonB:
> It would also be great if these combined with "shin" to help with selection.

I haven't been able to figure out how to achieve this. Any thoughts?

LexiconCode:
It might be easier to do a separate function for selecting texts within quotes, parentheses and so on rather than relying on vanilla "shin".

kendonB:
What about also having "jump start [lease]" and "jump end [Ross]" which would remain inside the parentheses?

What about also adding Repeats?

LexiconCode:
> What about also having "jump start [lease]" and "jump end [Ross]" which would remain inside the parentheses?
> 
> What about also adding Repeats?

That's a good idea to include it is optional for those that remain inside the parentheses. Repeat would be useful although with a small range, may be a max of 10?

Maybe the word "expand selection" could expand to fill "" then the command could be repeated to expand to parentheses or brackets? it could be useful in other contexts besides navigation.

kendonB:
Trying to think of a better spec than "jump" for when the user wants to stay within the current delimiters. For one, "jump end" is too similar to "jump in", both of which are allowed in the current proposal. Second, "jump" means "jump over the delimiter". What about "but" as in "but up against the delimiter"?

kendonB:
I've also been unable to figure out how to get Repeats to work with these types of functions.

LexiconCode:
> Trying to think of a better spec than "jump" for when the user wants to stay within the current delimiters. For one, "jump end" is too similar to "jump in", both of which are allowed in the current proposal. Second, "jump" means "jump over the delimiter". What about "but" as in "but up against the delimiter"?

For the moment will experiment with that. I can't help but have a nagging suspicion there might be a clever way of conditionally based on previous commands making this possible.

> I've also been unable to figure out how to get Repeats to work with these types of functions.
I haven't tried this before with this type of function either so I'll have to look into it.

kendonB:
Quotes don't have lefts and rights so the implementation would have to be different.

Perhaps we could replace "in" with "quotes" and it would just treat all quotes as equivalent.

As in to move the cursor from the "|" to in front of the "f":
```
st|ringfun("dictation", "is", "fan")
```
you would say "jump quotes five". Likewise, to get behind the "n" in "fan", you would say "butt quotes six".

What do you think?

kendonB:
Need to make sure these work when immediately outside the bracket. On mine they currently don't.

kendonB:
this one really has been superseded by the text manipulation functions. I'm going to close this

Issue #235, opened 2018-04-19
Redesign of Registered Actions with additional properties


LexiconCode:
These are some thoughts I've had on adding more metadata to [Registered Actions](http://caster.readthedocs.io/en/latest/caster/doc/readthedocs/examples/rules/Caster%20Rules/#the-context-stack). [Current Code implementation](https://github.com/synkarius/caster/blob/42a7c9cc64201ea71fc1a66d3e29790106e76701/caster/lib/dfplus/state/actions.py)

Currently Registered Actions provide the following but further detail in docs. 

- Registration with various printing services (such as the Python console, the status window, or loggers) allowing to see which actions executed, or that they have executed.
- Registration can be used mark a command as a part of another command. This is one significant difference between Caster and Dragonfly. Dragonfly commands are once-and-done. Caster commands can interact with other commands which have been spoken in the past or will be spoken in the future. In order to use this functionality, we set the rspec property of the R action.

The design should include [automatic document generation](https://github.com/synkarius/caster/issues/234), API, GUI interface, and [Consolidating command phrases](https://github.com/synkarius/caster/issues/211). In addition this should also simplify writing caster rules.

Overarching design thoughts for additional properties

- Properties may be toggled to not show up within logging/natlink window
- Properties may be optional but have default values and formatting.

Truncated code from Atom.py formatted for clarity

```python
# Utilizes the Palette UI to leverage commands.
def palettized(function_text):
    return Key("cs-p") + Text(function_text) + Pause(atom_palette_wait) + Key("enter")

class AtomRule(MergeRule):
    pronunciation = "atom"
```

```python
    # ----------Spoken Command/Action-----------------> Shortcut keys----------> #Displayed Text
#Basic Cursor Navigation
    "up [<n>]":
        R(Key("up"), rdescript="Atom: Move Cursor Up #")*Repeat(extra="n"),
#Basic White Text Manipulation
    "tab|indent [<n>]":
        R(Key("tab"), rdescript="Atom: Press Tab Key #")*Repeat(extra="n"),
    "space [<n>]":
        R(Key("space"), rdescript="Atom: Press Tab Key #")*Repeat(extra="n"),
# Menu UI-----------------------------------------------------------------------------------
    #File Menu
    "[open] new window":
        R(Key("cs-n"), rdescript="Atom: New Window"),
#Script
    "script run":
        R(palettized("Script: Run"), rdescript="Atom: Script: run"),
#Git Plus
    "git checkout [branch|tag]":
[...]

Issue #234, opened 2018-04-17
Automatic Grammar documentation


gerrish:
The Idea is to make a tool which will automatically generate documentation from Grammars. The way this will be done is by using the descriptions (`rdescript` and maybe an extra attribute which will be an extended description) in the mapping to create read me in markdown.

To make this work, We will need to standardize the way descriptions are written. The description Will need to contain the category In which it belongs as well as a short and concise description of the voice command's function.
1 example:
`"stoosh": R(...,rdescript="text editing: copy selected text to caster clipboard")`
This will produce a text editing section which will contain the command or add it to an existing section with this name. Every Grammar will also be different section with Subsections to avoid name collisions.
Alternatively every Grammar will produce

LexiconCode:
I think this would coincide very well with implementing https://github.com/synkarius/caster/issues/211 as it simplifies the documentation and requires re-factoring.

codingApprentice:
Would it be viable to produce something similar to Dragon's Learning Centre? That is to say, have a window which lists the commands depending on which application is in focus? (Include global commands as well of course). I assume there is some way of running this process in the background without removing the focus from the desired application you want to work on? And have a global command similar to Dragon's "what can I say?"

LexiconCode:
@codingApprentice that would be an awesome feature. We used to have the status window which was very basic and was cut feature. I would like to see something implemented in [kivy](https://kivy.org/#home). Why don't we start off by opening a ticket and as it's a distinctly different feature than automatically generating documentation. How we generate documentation should be considered when implementing if we want implemented GUI.

gerrish:
I like the Idea. I Could export a json with the data tο be used in such a program

LexiconCode:
Right now the settings window utilizes a SimpleXMLRPCServer. I'm hoping to convert that to [grpc](https://grpc.io/). I'm switching from rest to grpc for the API.

comodoro:
Any progress on this? A simple tkinter dialog window via named pipe did not work for me, looking into Caster code there is SimpleXMLRPCServer, but I can wait for grpc if it is the preferred way. Or even help with porting existing and creating documentation windows, if you are not in a hurry.

LexiconCode:
@comodoro besides documentation what would you implement using dialog windows?

comodoro:
Just playing. 1. A general query answer window, where the queries are programmable and assignable to voice commands. 2. Conversation with a chatbot.

LexiconCode:
[...]

PR #451, opened 2019-04-11
Redesign rdescripts

mrob95:
Currently, the accepted formatting for a caster command looks something like this:
```
"function": R(Text("def ():") + Key("left:2"), rdescript="Python: function"),
```

We have the verbal command, R - which creates a registered action, the actual dragonfly objects to be executed, and an rdescript written by whoever wrote the grammar to describe the command.

There are a number of problems with this construction.

1. It's long, often requiring multiple lines to define a single command.
2. It's verbose, only about half of the character count is actually useful code, the rest is verbiage.
3. It's a bug magnet, with a lot of potential for misplaced commas and brackets and misspelled words.
4. rdescripts are inconsistent between different authors and are a pain to write.

The combination of these problems end up making the code base harder to read and harder to work on, with little appreciable benefit.

The aim of this pull request is to radically simplify this construction, pushing stuff which should really be backend into the backend and out of the user's way. 

So far I have:
* Altered the execution of registered action so that dynamic elements can be added into the rdescript.
* Altered the merge rule so that all registered actions which do not have an rdescript will be given one, containing the name of the class, the command spec, and the extras to be dynamically added on execution.
* Deleted all rdescripts from the Python module for testing purposes.

A further improvement would be to have dragonfly objects be automatically converted to registered actions, removing the need for `R` as well. This should be possible but needs some work, as at the moment it causes a nasty dependency cycle.

Happy to get any thoughts.

### DIFF
--- a/castervoice/apps/chrome.py
+++ b/castervoice/apps/chrome.py
@@ -27,130 +27,120 @@ class ChromeRule(MergeRule):
     pronunciation = "google chrome"
     mapping = {
         "new window":
-            R(Key("c-n"), rdescript="Chrome: New Window"),
+            R(Key("c-n")),
         "(new incognito window | incognito)":
-            R(Key("cs-n"), rdescript="Chrome: New Incognito Window"),
+            R(Key("cs-n")),
         "new tab [<n>]":
-            R(Key("c-t"), rdescript="Chrome: New Tab")*Repeat(extra="n"),
+            R(Key("c-t")*Repeat(extra="n")),
         "reopen tab [<n>]":
-            R(Key("cs-t"), rdescript="Chrome: Reopen Tab")*Repeat(extra="n"),
+            R(Key("cs-t"))*Repeat(extra="n"),
         "close tab [<n>]":
-            R(Key("c-w"), rdescript="Chrome: Close Tab")*Repeat(extra='n'),
+            R(Key("c-w"))*Repeat(extra='n'),
         "close all tabs":
-            R(Key("cs-w"), rdescript="Chrome: Close All Tabs"),
+            R(Key("cs-w")),
         "next tab [<n>]":
-            R(Key("c-tab"), rdescript="Chrome: Next Tab")*Repeat(extra="n"),
+            R(Key("c-tab"))*Repeat(extra="n"),
         "previous tab [<n>]":
-            R(Key("cs-tab"), rdescript="Chrome: Previous Tab")*Repeat(extra="n"),
+            R(Key("cs-tab"))*Repeat(extra="n"),
         "new tab that":
-            R(Mouse("middle") + Pause("20") + Key("c-tab"),
-              rdescript=
-              "Browser: when the mouse is hovering over a link open that link in a new tab and then go to that new tab "
-              ),
+            R(Mouse("middle") + Pause("20") + Key("c-tab")),
         "<nth> tab":
-            R(Key("c-%(nth)s"), rdescript="Chrome: nth tab" ),
+            R(Key("c-%(nth)s") ),
         "last tab":
-            R(Key("c-9"), rdescript="Chrome: Last tab"),
+            R(Key("c-9")),
         "second last tab":
-            R(Key("c-9, cs-tab"), rdescript="Chrome: Second last tab"),
+            R(Key("c-9, cs-tab")),
         "go back [<n>]":
-            R(Key("a-left/20"), rdescript="Chrome: Navigate History Backward")*
-            Repeat(extra="n"),
+            R(Key("a-left/20"))*Repeat(extra="n"),
         "go forward [<n>]":
-            R(Key("a-right/20"), rdescript="Chrome: Navigate History Forward")*
-            Repeat(extra="n"),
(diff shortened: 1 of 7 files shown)


LexiconCode:
This is an excellent pull request as it addresses many of the points in dictation-toolbox/Caster#235. 

A number of issues so far.
1. Despite picking a different element choice the print out remains choice element first item for instance `Python: try [<exception>], StandardError`

2. The natlink window goes blank until switching non choice element or a different choice element from previous spoken.  For instance switching from "binary_meth" to "exception" (I've experienced this before but this seems to be exacerbated with this pull request)

3. `meth [<unary_meth>]`  only print out default which is an empty string.

> A further improvement would be to have dragonfly objects be automatically converted to registered actions, removing the need for `R` as well. This should be possible but needs some work, as at the moment it causes a nasty dependency cycle.

Personally pull request is already a large improvement. However I would get further feedback in Gitter from the community before implementing the above suggestion.

LexiconCode:
Two following issues have been resolved.
> 1. meth [<unary_meth>] only print out default which is an empty string.
> 3.Despite picking a different element choice the print out remains choice element first item for instance Python: try [<exception>], StandardError
Clarify 2 is not a bug with the pull request.

4. With default Configdebug.txt removing rdescript from a command causes it to print out to `none`.  most likely due to how the unconventional way configdebug is loaded over the issue below. 

5. Specs that include `Repeat(extra="n")` print out to `None`

```Python
        "Unindent [<n>]":
            R(Key("s-tab")) * Repeat(extra="n"),
```

**Other notes**
Is there a way to handle repeat elements with a number instead of printing out x repetitions?
For instance say "Next Tab 3"

Printout 
```
Core: Next Tab
Core: Next Tab
Core: Next Tab
```

mrob95:
> 1. With default Configdebug.txt removing rdescript from a command causes it to print out to `none`.  most likely due to how the unconventional way configdebug is loaded or related to the issue below.
This may be fixable, but is probably one for a different request since the mechanism would be quite different - currently this way only applies to merge rules when they are instantiated.

> 2. Specs that include `Repeat(extra="n")` without rdescript print out to `None`
> 
> ```python
>         "Unindent [<n>]":
>             R(Key("s-tab")) * Repeat(extra="n"),
> ```
Fixed. With the repeat outside of the R it was creating a dragonfly object instead of a registered action.

> 1. `make fake RegisteredAction` unit test may need to be reworked
Will check it out.


> **Other notes**
> Is there a way to handle repeat elements with a number instead of printing out x repetitions?
> For instance say "Next Tab 3"
> 
> Printout
> 
> ```
> Core: Next Tab
> Core: Next Tab
> Core: Next Tab
> ```
> is it possible to get at Dictation elements?
These should all work now.

LexiconCode:
With the Markdown CCR the element choice is not included.
```
Markdown: insert <element>, backtick:2, left
Markdown: insert <element>, asterisk:4, left:2
Markdown: insert <element>, dollar, dollar, left
```

mrob95:
At the moment that markdown behaviour is as expected - it will print the relevant value from the Choice dict.

rdescripts are only generated for Mergerules, the navigationnon was a mapping rule which was why it wasn't working. Fixed.

For reference, this is the query I'm using to delete rdescripts.

```
,\s?+rdescript=\".*?\"
```

LexiconCode:
So far I haven't had any more issues during testing. I believe we can be pretty liberal about removing most rdescript with the exception of those that are functionally descriptive beyond the command name. an example of this might be 
```
        "garb [<nnavi500>]":
            R(Mouse("left") + Mouse("left") + Function(
                navigation.stoosh_keep_clipboard,
                nexus=_NEXUS),
              rdescript="Core: Highlight @ Mouse + Copy"), 
```

kendonB:
Is it still possible to have this? https://github.com/dictation-toolbox/Caster/issues/234

seekM:
@LexiconCode As you said you haven't experienced more issues, can this be merged? 
In this PR `NavigationNon` is changed from a `MappingRule` to a `MergeRule`. In general all rules listed as a `non` should be changed from `MappingRule` to a `MergeRule`, shouldn't they @mrob95? What else has to be done? I'm looking forward to this because it would solve the issue I have that I can't change specs in  `MappingRule`s like `NavigationNon` without touching non user Caster files (see #466).

mrob95:
Going to merge this before there are too many more merge conflicts. Removing rdescripts can be done gradually. I'm busy with exams at the moment but can make time if there are any bugs with this that need fixing.

kendonB:
@mrob95 Can you summarise here what the new best practice is for rdescripts? And potentially pin the issue?

mrob95:
@kendonB Unless there is a particularly good reason I think best practice would be to not include rdescripts, and I think we should be aiming to remove them from the code base over time.

kendonB:
Can you give an example of "a particularly good reason"? If the spec was very opaque about what was trying to do?

LexiconCode:
> 
> 
> Can you give an example of "a particularly good reason"? If the spec was very opaque about what was trying to do?

That's pretty much it as seen in my example above. Where the command name is not descriptive enough to describe its function.

PR #459, opened 2019-04-17
New bring me folders command for Windows Explorer

alexboche:
Opens the folder in the same window of Windows Explorer that you are currently in including Windows Explorer child Windows.


LexiconCode:
The bring me commands are more like user-defined bookmarks this falls outside of that scope. I think something like this should its own grammar. More commands will be added to control Windows Explorer and they will become increasingly complex.  It could be made as an app grammar for now and once it's developed turned into core.

alexboche:
Okay I have removed the Windows Explorer bring me commands from the bringme.py file and I have added a bring me command into both Explorer.py (the file for Windows Explorer main window) and filedialogue.py (the file for Windows Explorer child window). These use the function paste_string_without_altering_clipboard . Unfortunately, these commands seem to occasionally not work – in particular occasionally nothing is pasted into the address bar. I'm not sure why this is, I'm guessing it might have something to do with the pausing between the execution of different actions, but I could be wrong. This may be different from user to user. Please let me know if you any thoughts.

mrob95:
I think using a `Text` call is probably better than pasting. Far more reliable and the loss of speed is not really a problem unless it is an extremely long string

alexboche:
good idea. I changed it so that it uses a text action now. I left commented out two other options which use pasting and are thus faster. one is somewhat unreliable and the other (based on paperclip) alters the clipboard. I have checked that caster will prioritize the application-specific commands over the global "bring me" command even though they all have the same name.

alexboche:
If I recall correctly, importing the app context from Caster was causing some kind of problem for me when I was using the paste text without altering clipboard approach instead of the text action. But it seems to be working okay now when I try the former approach, and in any case we are using the text action now.

kendonB:
Can we change the spec for this from "bring me" to "navigate to" so that it's clearly distinct?

kendonB:
It would also be amazing if this was clever enough to be extended to "Open" and "Save" dialog boxes also

alexboche:
it is already extended in a separate pull request to open and save dialog boxes as long as those words are in the title of the window (which is how the context works for filedialogue.py). The reason I was thinking "bring me" was so you would only have to remember one command and it would open a new window Windows Explorer if you're not already in one but use the same one year in if you are in one. I guess sometimes you might want to open a new window to navigate to the folder you want in which case it would be good to have a separate name "like navigate to" "versus bring me".

kendonB:
> it is already extended in a separate pull request to open and save dialog boxes as long as those words are in the title of the window (which is how the context works for filedialogue.py).

@alexboche can you link to this pull request and we will work on getting that reviewed first?

alexboche:
@kendonB my mistake, it is not a separate pull request it is included in this pull request.  Let me know if this is the sort of thing you'd like to see in two different pull requests (one for Windows Explorer child window and one for Windows Explorer main window). Though I'd ask that that you make an exception for since this is already here.

castervoice/apps/explorer.py
```diff
@@ -1,15 +1,42 @@
+# import pyperclip
 from dragonfly import (Grammar, MappingRule, Dictation, IntegerRef,
-                       Repeat, Pause)
+                       Repeat, Pause, Function, Choice)
 
-from castervoice.lib import control
+from castervoice.lib import control, context, utilities, settings
+from castervoice.lib.context import AppContext, paste_string_without_altering_clipboard
 from castervoice.lib import settings
```
kendonB: remove paste_string... and associated code commented out below. Always better to have something slow that works.

castervoice/apps/explorer.py
```diff
@@ -1,15 +1,42 @@
...
+def explorer_bring_it(folder_path):
+    Key("c-l/5").execute()
+    Text("{}".format(folder_path)).execute()
+    # Alternate method: is faster but somewhat inconsistent.
+        # if not paste_string_without_altering_clipboard(folder_path):
+        #     print("failed to paste {}".format(folder_path))
+        # Pause("10").execute()
+    # Second alternate method: is both fast and consistent, but alters the clipboard
+        # pyperclip.copy(folder_path)
+        # Pause("5").execute()
```
kendonB: I have found that the current method is quite fast. Let's avoid dependencies where practical. Let's remove this content that is commented out.

castervoice/apps/explorer.py
```diff
@@ -29,11 +56,16 @@ class IERule(MergeRule):
             R(Key("a-left"), rdescript="Explorer: Navigate back"),
         "get forward":
             R(Key("a-right"), rdescript="Explorer: Navigate forward"),
+        "[explorer] bring me <folder_path>":
+            R(Function(explorer_bring_it),
+            rdescript="go to preconfigured folder within currently open Windows Explorer main window"),
```
kendonB: Match the rdescript style as "Explorer: Go to..."

castervoice/apps/file_dialogue.py
```diff
@@ -1,12 +1,42 @@
-from dragonfly import (AppContext, Dictation, Grammar, IntegerRef, Key, MappingRule,
-                       Pause, Repeat, Text)
-from dragonfly.actions.action_mimic import Mimic
+# import pyperclip
```
kendonB: remove comment

castervoice/apps/file_dialogue.py
```diff
@@ -26,8 +56,13 @@ class FileDialogueRule(MergeRule):
             R(Key("a-d, f6:2"), rdescript="File Dialogue: Navigation pane"),
         "[file] name":
             R(Key("a-d, f6:5"), rdescript="File Dialogue: File name"),
+        "[(child | dialogue)] bring me <folder_path>":
+            R(Function(dialogue_bring_it),
+            rdescript="go to preconfigured folder within currently open Windows Explorer child window"),
```
kendonB: Match rdescript style

kendonB:
@LexiconCode wrote:
> The bring me commands are more like user-defined bookmarks this falls outside of that scope. I think something like this should its own grammar. More commands will be added to control Windows Explorer and they will become increasingly complex. It could be made as an app grammar for now and once it's developed turned into core.

The current version of this is within the `explorer.py` and `file_dialogue.py` files.  Does this address your earlier concern?

I also suggested earlier that the spec should be "navigate to" instead of "bring me". I now prefer "bring me". It is consistent with bringme opening a new tab when asking for a website, for example.

LexiconCode:
Yes I think that's a good way to proceed.

LexiconCode:
@alexboche are you going to proceed with this or should I close?

alexboche:
I believe have implemented all the requested changes. I believe this pull request is done.

LexiconCode:
Once the conflicts are I will test and merge.

kendonB:
This one will be superseded by #614, I believe.

kendonB:
Fixed via #614

PR #476, opened 2019-05-02
add folder path commands for command line in gitbash.py 

alexboche:
Also I enabled gitbash.py for powershell.


castervoice/apps/gitbash.py
```diff
@@ -121,9 +129,17 @@ class GitBashRule(MergeRule):
               rdescript="GREP: Search Recursive Filetype"),
         "to file":
             R(Text(" > FILENAME"), rdescript="Bash: To File"),
+        
+        # Folder path commands (not git specific)
+        "[folder] path <folder_path>": 
+            R(Text("%(folder_path)s"), rdescript="GIT: type in folder path"),
+        "(CD | go to) <folder_path>": 
+            R(Text("cd %(folder_path)s") + Key("enter"), rdescript="GIT: go to folder"),
+
```
kendonB: I quite like having "bring me" and "navigate to" as options here too

alexboche: If I call it "bring me" ( as one of the alternatives alongside "CD", "navigate to", "go to" ) it will typically override the usual bring me command which opens up a window in Windows Explorer. That sounds fine to me. Another option would be to make it something like "shell bring me".

kendonB: Once we get bring me to search for open instances and work with terminals we might have to revisit this. I'm happy to leave it in for now.

kendonB:
Need to get this to evaluate Windows environment variables"

```
kmbel@DESKTOP-VUFG094 MINGW64 ~
$ cd %DROPBOX%\projects_ac\Team_folder
bash: cd: %DROPBOX%projects_acTeam_folder: No such file or directory
```
For git bash, also need to swap "\" with "/" I believe.

kendonB:
@alexboche do you still want to work on this one?

alexboche:
> @alexboche139 do you still want to work on this one?

I probably won't have time to work on this for a while. Would somebody else mind taking this up? Sorry about the git bash thing with the slashes in the wrong direction. I use power shell so it works there, but I didn't think of the bash-style terminals.

LexiconCode:
I think will close this out for now we can re-add it later as needed.

kendonB:
I will work on this one.

mrob95:
Would this make more sense as a ContextAction in the "bring me" commands? Pseudocode but the mapping for "bring the <desired item>" could be something like:

```
ContextAction(default=open in new windows explorer,
[(AppContext("gitbash"), type cd ...)])
```

That way the commands are consistent and adapt to the context.

LexiconCode:
I like that idea has it really makes it much more consistent.

LexiconCode:
You also might want to be made aware for terminal/CMD.
Dragonflies[runcommand-action](https://dragonfly2.readthedocs.io/en/latest/actions.html#runcommand-action)
and
Casters [TerminalCommand](https://github.com/dictation-toolbox/Caster/blob/develop/castervoice/lib/terminal.py)

kendonB:
This is basically superseded by #614. We can implement a "paste me" command that just prints the bring me element everywhere to handle the folder path stuff.

PR #478, opened 2019-05-02
Fixed error in section "Template-based navigation"

brunoV:
There was an error in the "Template-based navigation" a section. It said "jump in" instead of "fill <target>".


LexiconCode:
Nice catch and welcome!

PR #479, opened 2019-05-03
add Python dictionaries for keywords etc.

alexboche:


alexboche:
This pull request is just me redoing #434

LexiconCode:
Outside of the Python standard library we are going have to figure out how to manage third-party libraries efficiently. I don't think it's sustainable to Include them in the Python module not just because of grammar complexity but also in the structure of organization.  Eventually we need to come up with a system to break it down by library with a way to load them as needed.


Feedback from the community and how this might be achieved with a modular implementation or if there's a preference towards a monolithic grammar.

kendonB:
I feel like the long lists of commands belong in `.caster` in a `.toml` file. Each user will like/dislike different functions. We can provide a few common items by default and help people get started editing it for their own use.

kendonB:
@alexboche would you like to have a go at trying to set up a `python.toml` in `.caster/data/lists` (or a better name than "lists" if you think of one) and have `python.py` pull those items in?

alexboche:
Yeah I'll do the toml thing good idea.

LexiconCode:
I was originally thinking of breaking it down by library per language. However I realized it to be a monumental task to keep those up-to-date. I think a simple list is a good intermediary for the time being.

kendonB:
@alexboche you can use the `USER_DIR` variable from `settings.toml`

kendonB:
We could also consider adding automated functions that will:
 - add the selected text to a given list, identified via a spec. e.g. "add this to python functions list". The exact function name will be the default spec. I presume there is a good toml writing function in python. Then:
 - open the list toml at the position of the new spec so the user can edit it. Can use Ctrl-F for this bit.

kendonB:
Just so we have it here, this one is on hold while we wait for @mrob95 to move all the language and app files into the use of folder and this will become `python.py.defaults`. We'll then revisit exactly how much we want the default to have and how much we want to share via "dot_caster" repos.

LexiconCode:

PR #480, opened 2019-05-03
Add and adjust keystroke commands in nav.py

alexboche:
Pressing control backspace doesn't work in all application so I changed it to use control shift left/right delete. Also, as it was, "splat lease" was sometimes interpreted as to commands and sometimes interpreted as one. I remove this ambiguity by making "splat" for deleting words to the left and "sprat" for deleting words to the right.

### DIFF
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -232,8 +232,16 @@ class Navigation(MergeRule):
         "spark [<nnavi500>] [(<capitalization> <spacing> | <capitalization> | <spacing>) (bow|bowel)]":
             R(Function(navigation.drop_keep_clipboard, nexus=_NEXUS), rspec="spark", rdescript="Core: Paste"),
 
-        "splat [<splatdir>] [<nnavi10>]":
-            R(Key("c-%(splatdir)s"), rspec="splat", rdescript="Core: Splat") * Repeat(extra="nnavi10"),
+        "splat [<nnavi50>]": 
+            R(Key("cs-left:%(nnavi50)s, del"), rspec="splat",  
+            rdescript="Core: delete words to the left"),
+        "sprat [<nnavi50>]":
+            R(Key("cs-right:%(nnavi50)s, del"), rspec="sprat",  
+            rdescript="Core: delete words to the right"),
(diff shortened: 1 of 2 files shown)


alexboche:
I appreciate the feedback, and I'll try to do a better job with making pull requests into more atomic units with matching descriptions. I hope people can bear with me though. My understanding of git/github is poor and I am currently unwilling to use a mouse or keyboard at all for fear of exacerbating my hand problems, so it is possible that some users may be able to do things (especially relating to git and testing) much much faster than me. Sometimes I spend hours on something in git that somebody knowledgeable could probably do in thirty seconds. I certainly appreciate everyone's work here and realize that many users also have a lot of difficulty using a computer. I am working on my git skills so I can make more competent pull requests.

kendonB:
I'm personally very excited that you're keen to make caster better! Feel free to ask questions (to google first, then gitter) if it feels like something is taking you a long time. If it feels like there's a faster way, there's probably a faster way. I'm also going to get write access soon so that this process could go a bit faster.

alexboche:
Great thanks!

kendonB:
I definitely like the idea of moving from "splat lease" and "splat ross" to two individual words (I do that in my own version using "nope" and "kay", the latter because I had a lot of trouble with other "cuter" alternatives). We would definitely want a bit more input on that decision, though. How would one revert to the old specs? Would words.txt be able to handle going back?

LexiconCode:
"words.txt" would not be able to handle multiple strings. However spec replacement is possible using https://github.com/dictation-toolbox/Caster/blob/de8536eb3c74996bb01c76c9846260f44a1a3e4a/caster/user/filters/examples/scen4.py#L27

Unfortunately the Filter Rules are not nearly as accessible as words.txt.

kendonB:
@alexboche, can you elaborate a little bit on your experience with Dragon sometimes interpreting your command as "splat lease" and sometimes as "splat" then "lease"? Did it just seem random? If you leave a gap between the words in your speech it can get interpreted as two commands.

kendonB:
I think a good general rule is to leave the core specs alone, unless they are buggy.

kendonB:
this one is superseded by #585. closing

PR #481, opened 2019-05-03
move some commands in nav.py to CCR from non-CCR

alexboche:
I wasn't sure if I should make middle click CCR but I was pretty sure about the other ones


LexiconCode:
We will see how this goes in practice and we can always change things at a later date with feedback

LexiconCode:
I accidentally merged the committee prematurely. You'll need to reopen the pull request.

PR #484, opened 2019-05-04
add Talon alphabet but leave it commented out

alexboche:


kendonB:
Should this sort of thing really be included via `words.txt`?

LexiconCode:
I would be hesitant to add it to the main repository because of permissions and commented code as a source of customization is undesirable. However it does not mean I don't think it's useful or that it should not be preserved. 

An alternative would be to document caster user setups. Therefore reducing fragmentation of how Caster users implement customizations.  Caster users could include `.caster` and upload it to a separate repository or branch. We could then link it in the Caster Documentation to show off user customizations and implementations.

kendonB:
One day I'd like for everyone to be able to to all customizations using just the `.caster` folder and not touching, nor ever wanting to touch, the main source. @alexboche would you like to convert this content to a mock `words.txt` and then link a gist here?

alexboche:
I haven't really use words.txt. would it be like this?
arch -> air
brov -> bat

kendonB:
Mine is here and it includes a few alphabet changes:
https://github.com/kendonB/dot_caster/blob/master/data/words.txt

PR #395, opened 2019-02-25
Implemented Accessibility API

LexiconCode:
Info on what this is and what this does can be found on [Hands Free Coding](http://handsfreecoding.org/2018/12/27/enhanced-text-manipulation-using-accessibility-apis/ ). Credit for the goes Wolfmanstout

**Note**
Follow the install instructions on in the link above for the Accessibility API.

Works for Chrome, Firefox, Discord, and Gitter

-  To test ` words <dictation>` Command will highlight/select the dictated text. For example:

   1. Open an application and click on a field that accepts text. Say `this is a test`.

   2. Say the command `words <is>`. In `this is a test` the word `is` will be highlighted.

Natlink Window print out results

- Supported Application

`accessibility: Selecting text: {'end_phrase': 'is'}`
`accessibility: Selected text`

- Unsupported application or nonfunctioning api

`accessibility: Selecting text: {'end_phrase': 'is'}`
`accessibility: Nothing is focused.`

**Known issues - Needs further testing by the community**

It works intermittently for some programs. It's bulletproof for Firefox and Chrome. However for gitter and discord the API seems to be sporadic. When it does work for those two applications works until Dragon is restarted. Functionality is all or nothing. For example, If it works in gitter then likewise discord or not at all in those applications. Firefox and Chrome are unaffected. I need more people to test the Accessibility API to confirm.

**Want to test API compatibility for new programs?**
If you wish to test the API to find new programs use [configdebug.txt.]( https://files.gitter.im/sphinx-dragonfly/Lobby/p2ve/configdebug.txt.) It's a global grammar which will be available on any program.

Instructions for the **development** branch. For **master** branch use `\caster\bin\data` instead of`C:\Users\%USERNAME%\.caster\data`

Instructions to use configdebug.txt on the latest version of Caster Development branch.

1. Edit or replace `configdebug.txt`  in `C:\Users\%USERNAME%\.caster\data`
3. Then say `refresh debug file` to load the debug grammar.

In the event that you find a new program let me know here. I will add it to the grammar.

PR #485, opened 2019-05-04
add text manipulation commands  to nav 

alexboche:
The supporting functions are in navigation.

These could definitely be improved, for example by extending into multiple lines and optimizing the keypress speed and weight times. But I think we should get these commands out there sooner rather than later since they are quite useful and eventually these may be made obsolete by more sophisticated commands like Wolfmanstouts accessibility commands.

I have tested these but I would recommend testing them again.

These work in most applications but not all for example Microsoft Word, probably something to do with the wait times in the caster clipboard functions since straight pyperclip ( which unlike the caster clipboard functions does alter the clipboard ) seems to work better there.

I believe you need the latest version of dragonfly for these to work dragonfly2-0.13.0 because the functions use a new argument-name remapping feature of the Function action. (Prior to that change, I think most people used lambda functions which still works also.)


kendonB:
Please also 1) refer to relevant issues if they exist and 2) provide some explanation of what you did in the PR for all but trivial changes. You wouldn't need to explain what you did if fixing a typo, for example. Thanks for getting stuck in to helping Caster!!

kendonB:
These look very neat! Will need some testing before merging though.

castervoice/lib/ccr/core/nav.py
```diff
@@ -189,94 +186,203 @@ class Navigation(MergeRule):
...
+            R(Function(navigation.copypaste_replace_phrase_with_phrase,
+                       dict(dictation="replaced_phrase", dictation2="replacement_phrase"),
+                       left_right="right"),
+              rdescript="Core: replace text to the right of the cursor"),
+        "remove lease <dictation>":
+            R(Function(navigation.copypaste_remove_phrase_from_text,
+                       dict(dictation="phrase"),
+                       left_right="left"),
+              rdescript="remove chosen phrase to the left of the cursor"),
+        "remove lease <left_character>":
```
kendonB: Need to add "Core: " to the rdescripts

castervoice/lib/ccr/core/nav.py
```diff
@@ -332,6 +438,42 @@ class Navigation(MergeRule):
...
+            "left_character", {
+                "prekris": "(",
+                "right prekris": ")",
+                "brax": "[",
+                "right brax": "]",
+                "angle": "<",
+                "right angle": ">",
+                "curly": "{",
+                "right curlry": "}",
+                "quotes": '"',
```
kendonB: curlry is deliberate?

kendonB:
Is it easy to add one that highlights text?

kendonB:
I'm finding that "go ross/lease <dictation>" do not work correctly within TexStudio, but work fine and notepad++ (Why?).

kendonB:
Do you also think that this could be clever enough to say things like "capitalize the p"? This one would work just like the Dragon command.

kendonB:
I'm finding that "change lease x to y" decapitalizes everything in the line.

castervoice/lib/ccr/core/nav.py
```diff
@@ -23,10 +23,9 @@
 class NavigationNon(MappingRule):
     mapping = {
         "<direction> <time_in_seconds>":
-            AsynchronousAction(
-                [L(S(["cancel"], Key("%(direction)s"), consume=False))],
-                repetitions=1000,
-                blocking=False),
+            AsynchronousAction([L(S(["cancel"], Key("%(direction)s"), consume=False))],
+                               repetitions=1000,
+                               blocking=False),
```
kendonB: Please revert these superfluous changes for now. We can fix style in a separate PR.

kendonB:
I agree with @alexboche that these changes are worth pushing to the top of the list. And they're almost there. I have tagged this as high-priority; @LexiconCode please feel free to untag if you disagree.

kendonB:
@alexboche do you know what's going on with the capitalization issue?

kendonB:
I got this error at some stage:

```
action.exec: Execution failed: Playback failed: No matching rule found for words ['click'].
engine: Grammar MergeRule(Merged143Mh): failed to decode recognition (u'flat', u'go', u'ace', u'flat', u'lease').
action: Exception from function copypaste_replace_phrase_with_phrase:
  File "C:\Python27\lib\site-packages\dragonfly\actions\action_function.py", line 136, in _execute
    self._function(**arguments)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\navigation.py", line 81, in copypaste_replace_phrase_with_phrase
    new_text = replace_phrase_with_phrase(text, replaced_phrase, replacement_phrase, left_right)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\navigation.py", line 57, in replace_phrase_with_phrase
    raise ValueError("{} not found".format(replaced_phrase))
ValueError: did not found
action.exec: Execution failed: Function(copypaste_replace_phrase_with_phrase): did not found
```

alexboche:
I think other changes you are referring to resulted from me applying in autoformater yapf (I tried to do that just for the new stuff,but I guess it accidentally got applied to the entire file.) The text manipulation functions I wrote in navigation.py raise value error if they can't find the substring (see the third to last line of your error). I will work on this a bit more.

castervoice/lib/ccr/core/nav.py
alexboche: that's a typo

kendonB:
Changing the sleep time does not seem to help. 

Is it fast to just restore the clipboard?

kendonB:
Since the main benefit of this is for text navigation and editing while writing, it would be great to get it working in the big text editors that people use for writing. Word already has half decent native support in DNS so I wouldn't worry too much about it working there, though others may disagree. This is my list off the top of my head.

- [ ] TexStudio
- [ ] Texmaker
- [ ] Overleaf
- [ ] Google Docs
- [ ] Word

kendonB:
Is there any downside to making lease the default for "change x to y"? I imagine in most cases people will be editing back something they just said.

kendonB:
Is it easy to limit the options for interpretation to only what is there? Kind of like we do in other files where we limit the options to a specified list of (usually) functions?

alexboche:
The problem with tex studio seems to me that when you have stuff highlighted in it, and then you press the arrow key to unselect, the cursor pops up on the opposite side of where it does in most applications. This is solved by just pasting the same text right over itself rather than using the arrow keys to unselect for the "go lease <dictation>"  commands. I'm going to try to rewrite the commands using regular expressions to avoid issues like "and land island". I'm going out of town for a few days, I might not get to this for a bit. 

As for limiting the options for interpretation to only what is there: That sounds great, I don't know how to do that though. I think that is what Dragon's native commands do

kendonB:
This is a bug in TexStudio. Let's see if they fix it https://github.com/texstudio-org/texstudio/issues/599

kendonB:
> The problem with tex studio seems to me that when you have stuff highlighted in it, and then you press the arrow key to unselect, the cursor pops up on the opposite side of where it does in most applications.

"Ctrl-Shift-Left" instead of "left" etc might fix this

kendonB:
There are enough exceptions to the usual behavior (notably emacs) that it's worth making a solution that works for both cases: https://github.com/texstudio-org/texstudio/issues/599

I believe that switching the <kbd> Ctrl </kbd>+<kbd> Shift </kbd>+<kbd> Left </kbd> --> <kbd> Right </kbd> cycle to <kbd> Ctrl </kbd>+<kbd> Shift </kbd>+<kbd> Left </kbd> --> <kbd> Ctrl </kbd>+<kbd> Shift </kbd>+<kbd> Right </kbd> will work if it's only one <kbd> Ctrl </kbd>+<kbd> Shift </kbd>+<kbd> Left </kbd> to start with. If it's more you just need to add more reversals.

alexboche:
Okay I made some improvements. Should be working in more applications now (e.g. tex studio). @kendonB  Can you test these? still for me to do: put the long choice dictionaries into a toml file and add the alphabet. thanks

alexboche:
Okay I made some changes/improvements. It uses pyperclip now instead of casters clipboard functions, but still puts the thing back on the clipboard at the end so as to not alter it (unless you're using a multi-clipboard in which case it will put some extraneous stuff on the second slot of the multi-clipboard). And now uses regular expressions instead of string methods. I added "grab <dictation>" which is for selecting until a phrase. Everything generally seems to work better and more applications now, though still inconsistent in Microsoft Word and Lyx (a little inconsistent in tex studio but lately is working).

Can somebody please test these on their computer (@kendonB  ?)?

I will try to put the long choice dictionaries into a toml file and add the alphabet. Other things to add would be a general selection command (as opposed to select until) and add functionality for <phrase> through <phrase>. Obviously also multiline functionality. And automatically putting the words in the selected text into a grammar so that Dragon prioritizes them, as discussed above (I think this is much more difficult).

kendonB:
> long choice dictionaries

What are these?

kendonB:
I've tried this a few times using Notepad++, TexStudio, and RStudio and it now almost never works...

There errors I see are:

```
'import' not found
'import' not found
'out' not found
'confirmed' not found
'credits' not found
'directory' not found
'gravitational' not found
'distance' not found
```

kendonB:
I also find myself wanting to use caster commands to set the format of the replacement. So it would be amazing to be able to say "change lease amazing to yell bow superb". Is this possible?

kendonB:
Are you also able to make it an option to say "Replace with" instead of "Change to"?

alexboche:
Hmm, It's working for me in texstudio and notepad++ ,I don't have Rstudio. Maybe we should get a third-party to test it. I used change to instead replace because "replace"is used by the accessibility API commands. A message like "'gravitational' not found" would normally mean that you said something like "go lease gravitational" and then the function that searches the text did not find the word "gravitational" in the text.If everything is spelled correctly this would most likely be because the selected text was not successfully passed into the function to search for the phrase "gravitational".  Will think about the formatting stuff

kendonB:
Yes it appears that the text does not successfully get passed into the function. Could this be a matter of a pause being too short? can you point me to the pause parameter that I should mess around with?

kendonB:
I just noticed an unexpected usage of "wipe". I think to avoid this we need to make everything have "lease" or "ross" (i.e. not optional)

kendonB:
Also just tried to Google "change legend title in ggplot" and it triggered the function from here.

alexboche:
I agree the direction should not be optional. I have updated it to that effect. And for the record ( so we have everything in the same place here), I think the crucial pause for several applications like tex studio is right after pressing control-c and that has been updated above. 50 ( I think that's milliseconds) seems to be enough but more should be added if necessary. I have also added some commented-out print statements that would print the variable 'selected text' to see if it was being properly passed into the function. If the value of the variable 'selected text' is not what it should be, but rather is the thing that was previously on the clipboard, then the pause time should be increased. Ideally the pause time would be dependent on the application so as to avoid unnecessary pauses for applications that don't require a long pause ( for example VS code ).

kendonB:
>  I used change to instead replace because "replace"is used by the accessibility API commands.

@alexboche What does this refer to?

alexboche:
The method of pasting the selected text over itself in order to unselect appears to work in all applications, while the method of unselecting the text by pressing the opposite arrow key does not work in texstudio and I think it also doesn't work in Emacs. However, the method of pasting the selected text over itself in order to unselect is sometimes substantially slower (unless we want to speed up the key pressing so much that it doesn't matter). The reason it can be slower is that the cursor jumps to the opposite side of the line and then has to come all the way back. Depending on the initial position of the cursor and the position of the word/phrase (relative to the end of the line) either method can be faster although I think the method of unselecting by pressing left is most often faster. 

Furthermore, I could easily (and will I think) tweak the functions so that it chooses whether to first jump to the end of the line and then go back or to just go directly to the phrase depending on the the positions mentioned above. In that case, I think the method of the selected text over itself would be dominated by the other method in applications where the latter method works. Therefore it may be a good idea to have two different implementations. Right now I have two different implementations in the same file "navigation" where one is commented out, but we may want to put the "pasting over the selected text" method in its own file application specific to texstudio, Emacs, and any other applications with similar behavior.

> > I used change to instead replace because "replace"is used by the accessibility API commands.
> 
> @alexboche363 What does this refer to?

"replace <dictation> with <dictation>" are commands here for the same purpose https://github.com/dictation-toolbox/Caster/pull/395/files. see the discussion here http://handsfreecoding.org/2018/12/27/enhanced-text-manipulation-using-accessibility-apis/
Those commands are limited to a fairly small number of applications ( the applications I know of where works is chrome, Firefox, discord, texmaker ), but when they work they are better than the commands I've written here.

kendonB:
Found another wee bug. To reproduce (there may be more ways than this):
 - Do "change lease X to Y" where X does not exist from the right of a line of text.
 - Change to a different line of text and do "go ross Z" from the right of a line of text. 

## Expected Behaviour
Nothing should happen because there is nothing to go right to from the right end of the line.

## Actual Behaviour
The clipboard contents from the first line get copied at the right of the cursor.

kendonB:
I'm also finding that "remove ross left prekris" "remove lease right prekris" both do not work.

kendonB:
With "remove lease <dictation>", it looks like you remove the character immediately preceding the word? Can you change this so that it's only going to automatically remove that character if it is a space?

kendonB:
Starting with:
`"<direction> <time_in_seconds>":`
With the cursor at the right. 

I do "go lease before seconds" and I get the error message "seconds not found". I'm guessing that your methods to prevent picking up substrings is too restrictive? Can you get it to assume that "_" also delimits words?

kendonB:
In notepad++, using the above line I do "go lease after direction" and I end up here "d|irection". Why? it seems to work fine in chrome.

kendonB:
Again using:
`"<direction> <time_in_seconds>":`

In notepad++, using the above line I do "go lease after left angle" and I end up here "dire|ction". Why? it seems to work fine in chrome.

kendonB:
Can we also add a <before_after> to the wipe lease/ross dictation?

kendonB:
@alexboche I found a few more bugs that need squashing. I imagine once I believe that there are no bugs left there still will be some so I think we should move this out of navigation and into "experimental navigation" so that we can merge these changes soonish and users can disable them if they wish. I think it would be cool to add a message to the caster loading output that advertises these, because when they're working these can be seriously powerful and a genuine game changer for the quality of caster.

kendonB:
Also, FYI I still haven't retested in TexStudio and I imagine there will be more changes required after that. We might also consider merging these in and announcing in the loading message that we haven't tested them in all editors and to ask users to raise issues for specific editors as they find them.

kendonB:
I believe that these changes make https://github.com/dictation-toolbox/Caster/pull/447 obsolete. "jump in ross" would now be "go ross after left prekris/brax", for example. We might also consider deleting those commands. I wonder if anyone would miss them?

kendonB:
> "replace with " are commands here for the same purpose https://github.com/dictation-toolbox/Caster/pull/395/files. see the discussion here http://handsfreecoding.org/2018/12/27/enhanced-text-manipulation-using-accessibility-apis/
Those commands are limited to a fairly small number of applications ( the applications I know of where works is chrome, Firefox, discord, texmaker ), but when they work they are better than the commands I've written here.

I don't think there's any reason why we shouldn't have common specs for common behavior even if they're from different functions. I'll be able to say "enable experimental navigation" or "enable accessibility API". I think there should be a way to get caster to prefer the latter if there's a conflict.

kendonB:
Let's also make "go to" have lease/ross compulsory. I had it trigger at a time when I was trying to say something completely different.

kendonB:
> Can we also add a <before_after> to the wipe lease/ross dictation?

And `grab`/`remove`?

LexiconCode:
This is shaping up. Before this is merged the commented out code needs to be removed.

castervoice/lib/ccr/core/nav.py
```diff
@@ -319,8 +397,69 @@ class Navigation(MergeRule):
...
+                "[right] brax": "]",
+                "left brax": "[",
+                "[right] angle": ">",
+                "left angle": "<",
+                "[right] curly": "}",
+                "left curly": "{",
+                "quotes": '"',
+                "single quote": "'",
+                "comma": ",",
+                "period": ".",
```
kendonB: I think this should be "dot" to match the caster default spec

kendonB: `(period | dot)`

kendonB:
Another one for the wish list: add a `[<which>]` parameter which defaults to `first` but allows for everything up to `tenth` to let the user choose which, for example, comma they mean. It would be counting from the cursor in whatever direction is specified. This one is not required for this particular merge but let's keep in mind for a future change.

alexboche:
I have uploaded a new draft above. Yeah that [<which>] command will be cool. I can add that soon.

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+        # The solution is to add a longer pause after pressing control see in the supporting functions in text_manipulation_functions.py
+        # For some applications this pause ( and other pauses in the functions for that matter ) is not necessary
+        # and may be removed by the user if they wish to speed up the execution of these commands
+        
+
+
+# How to better consolidate these context actions so I'm not repeating myself so much???
+        
+        # PROBLEM: sometimes Dragon thinks the variables are part of dictation.      
+        "(replace|to) <lease_ross> [<number_of_lines_to_search>] <dictation> (with|to) <dictation2>":
```
kendonB: I presume you meant `(replace|change)`

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+            R(ContextAction(default=Function(text_manipulation_functions.copypaste_replace_phrase_with_phrase,
+                       dict(dictation="replaced_phrase", dictation2="replacement_phrase", lease_ross="left_right"), 
+                       cursor_behavior="standard"), actions=[
+                        # Use different cursor method for texstudio
+                        (AppContext(executable="texstudio"), Function(text_manipulation_functions.copypaste_replace_phrase_with_phrase,
+                       dict(dictation="replaced_phrase", dictation2="replacement_phrase", lease_ross="left_right"), 
+                       cursor_behavior="texstudio"))]),
+              rdescript="Core: replace text to the left or right of the cursor"),
```
kendonB: This repetition is fine for now. Might be worth addressing if the number of contexts explodes.

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+
+from castervoice.lib.context import AppContext
+from castervoice.lib.dfplus.additions import IntegerRefST
+from castervoice.lib.dfplus.merge import gfilter
+from castervoice.lib.dfplus.merge.mergerule import MergeRule
+from castervoice.lib.dfplus.merge.ccrmerger import CCRMerger
+from castervoice.lib.dfplus.state.short import R
+from castervoice.apps import utils
+from castervoice.apps import reloader
+from castervoice.apps import reloader
```
kendonB: repeated

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+
+from castervoice.lib import control, settings, text_manipulation_functions
+
+from castervoice.lib.context import AppContext
+from castervoice.lib.dfplus.additions import IntegerRefST
+from castervoice.lib.dfplus.merge import gfilter
+from castervoice.lib.dfplus.merge.mergerule import MergeRule
+from castervoice.lib.dfplus.merge.ccrmerger import CCRMerger
+from castervoice.lib.dfplus.state.short import R
+from castervoice.apps import utils
```
kendonB: This doesn't exist in castervoice/apps?

castervoice/lib/ccr/core/text_manipulation.py
kendonB: Also doesn't seem to exist

kendonB:
Now that you can both search up and down, what do you think about making sauce a synonym of lease and dunce a synonym of ross here?

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+            Function(text_manipulation_functions.select_until_phrase, dict(left_character="phrase"), 
+            left_right="left", cursor_behavior="texstudio"))]),
+            rdescript="select left until chosen character"),
+        "grab ross [<number_of_lines_to_search>] until  <right_character>":
+            R(ContextAction(default=Function(text_manipulation_functions.select_until_phrase, dict(right_character="phrase"), 
+            left_right="right", cursor_behavior="standard"),
+            actions=[(AppContext("texstudio"), Function(text_manipulation_functions.select_until_phrase,
+            dict(right_character="phrase"), left_right="right", cursor_behavior="texstudio"))]),
+            rdescript="select right until chosen character"),
+        "wipe <lease_ross> [<number_of_lines_to_search>] [<before_after>] <dictation>":
```
kendonB: The interface for wipe and grab should be identical because one just adds a backspace at the end right? Is wipe really necessary when I can just add "clear" after grab (I'm not really sure either way)? Should "select" be an option instead of "grab"? Also need to add "until" to wipe (if we keep it) and is there a reason to make "until" compulsory?

kendonB: There should also be a before/after for grab

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,254 @@
...
+            cursor_behavior="standard"), actions=[(AppContext("texstudio"),
+            Function(text_manipulation_functions.select_until_phrase, dict(dictation="phrase", lease_ross="left_right"), 
+            cursor_behavior="texstudio"))]),
+                 rdescript="select until chosen phrase (inclusive)"),
+        "grab lease [<number_of_lines_to_search>] until <left_character>":
+            R(ContextAction(default=Function(text_manipulation_functions.select_until_phrase, dict(left_character="phrase"), 
+            left_right="left", cursor_behavior="standard"), actions=[(AppContext("texstudio"),
+            Function(text_manipulation_functions.select_until_phrase, dict(left_character="phrase"), 
+            left_right="left", cursor_behavior="texstudio"))]),
+            rdescript="select left until chosen character"),
```
kendonB: Add "Text Manipulation:" to the rdescripts and remove those for which the spec tells you everything you need to know about what it does.

kendonB:
Starting with:
```
"grab ross [<number_of_lines_to_search>] until  <right_character>":|
```
I say "wipe lease brax" I expect:
```
"grab ross |
```
but I get
```
"grab ross [|
```
So for that one I think that wipe/grab should delete/select the  `[`, suggesting that the default for wipe lease should be "before".

kendonB:
We should advertise this in the startup messages as well which I think is just a `print()`. Mention that the interface is subject to change and ask for feedback on interface/bugs/behavior. Mention how to turn it on as well

kendonB:
I also think that the exact character chosen for brax/prekris etc should depend on both lease/ross and before/after. If I say "go lease after brax", the cursor ends up after the `[`, rather than after the "brax" which is after the `]`. I think this one is worth addressing before merging.

castervoice/lib/ccr/core/text_manipulation.py
alexboche: Right now `wipe` pastes the altered tex over the original text; it does not specifically select the text to be deleted. I think it's a little faster than the `grab until` commands and hence faster than "grab until clear". I think when I tried to use "select" in the spec, Dragon prioritized its native selection command.

alexboche:
Keep in mind that you can specify whether you want `[` or `]`by saying "left brax" or "right brax".But I guess you are saying you want better defaults. Relatedly, do you want different before_after defaults for punctuation versus dictation?

kendonB:
"single quote" should be "thin quotes" to be consistent with navigation.

kendonB:
Starting with:
```
|    "go to line <n>":
```
I say "go ross before quotes" and it puts me before the second quotes. It seems to miss the first one.

kendonB:
Starting with:
```
|        IntegerRefST("n", 1, 10000),
```
I say, "go ross after quotes" and it seems to ignore the quotes here and the cursor stays in the same place.

kendonB:
We will also have to add a section to this document: https://github.com/dictation-toolbox/Caster/blob/develop/castervoice/doc/readthedocs/CCR_languages_Quick_Reference.md

kendonB:
> Relatedly, do you want different before_after defaults for punctuation versus dictation?

These definitely should be the same. As in "go ross brax" should have the same default as "go ross dictation". I'm not sure whether it should be before or after but I would lean towards the side which the cursor would hit first. As in, after for lease and before for ross.

alexboche:
I want to clarify something. Here's what I'm thinking would be good, I'm not sure if this is what you're suggesting or not. When you do `lease` (`ross`), it defaults to `before` (`after`). `before` (`after`) determines whether the behavior will be done on the left or right phrase in question. For parable punctuation, the user can specify e.g. `left brax` or `right brax`. If you don't specify left/right but rather just say `brax` it will choose `[` if you're moving towards the left of the cursor (i.e. if you say `lease`) and `]` if you're moving towards the right of the cursor (i.e. if you say `ross`). It sounds like you might be suggesting instead that whether `[` or `]` is used when you just say `brax` would be jointly determined by whether you say `lease` versus `ross` and whether you say `before` versus `after`. If this is what you're saying, please specify the behavior you want for each of the four combinations.

> I also think that the exact character chosen for brax/prekris etc should depend on both lease/ross and before/after. If I say "go lease after brax", the cursor ends up after the `[`, rather than after the "brax" which is after the `]`. I think this one is worth addressing before merging.

kendonB:
I (now) agree that the default should be to always go all the way through whatever is asked for. 

In the "pairable punctuation" case, when I say "brax", "prekris" etc my mind thinks about the pair, so going after "brax" should mean going after the pair, which is the `]`.

Starting with this:
```
|manipulation[two]
```
"go ross after brax" should end up like this:
```
manipulation[two]|
```
"go ross before brax" should end up like this:
```
manipulation|[two]
```
Starting with this:
```
(manipulation[two], text|)
```
"go lease after brax" should end up like this:
```
(manipulation[two]|, text)
```
"go lease before brax" should end up like this:
```
(manipulation|[two], text)
```

alexboche:
Note for the multiple occurrences,there is a bug where the regex does not match properly when there are two of the same word in a row. I will think about/work on the the parentheses stuff now.

alexboche:
Thanks for your explanation regarding the pairable punctuation. I think what you're suggesting sounds good, but there are number of cases to consider and I think it would be best to get more feedback before investing more to change the defaults. As it is, even if the default isn't exactly the way you want for when the user  just says"brax", users can still get whatever behavior they want if they specify whether they want the open or close bracket e.g. "left brax" or "right brax". It might be worth noting that the functions I wrote are more designed with words rather than parentheses in mind; furthermore, many text editors have more sophisticated tools for parentheses management. I have some other work I need to get done now so I'd rather not do more without more user feedback, apologies. But I think this is in a pretty good spot for more people to try it out. If there is something that is totally essential, I can probably do that now but otherwise I think I'm gonna take a step back and wait to get some more feedback. Thanks for all your guidance with this.

kendonB:
Thanks heaps again for all your work. I'm going to try these out in the wild for at least a couple of days bug hunting before merging.

kendonB:
I had initially misunderstood the distinction between "grab" and "grab until". Is there any reason not to also have "wipe" become "remove" and make the "until" compulsory. I feel like the deletion and the selection commands should be as close to each other as possible.

kendonB:
Is it also possible to add "replace ross <character> with <character>". I think a pretty common action in writing is wanting to replace "." with "?", for example.

kendonB:
I also think, keeping in line with the "prekris" refers to the pair, if I say "replace ross prekris with brax", I expect it to remove the pair. 

Because you won't be able to match up the pairs, I wonder if it would be better to enforce that the user specify left or right for "remove" and "replace". What do you think?

alexboche:
I added the replace character command and changed "wipe" to "delete… until". The last commit takes out the separate commands for <left_character> and <right_character> and replaces them with just one command (per function) for <character>. So now you cannot just say e.g. "brax" but I did add alternate names for the left and right characters that are one syllable (you can still say "left brax" and "right brax"). Later I can work on the defaults so that you can just say brax. In the future, the defaults will be handled within the functions rather than through dragonfly. Later I also can add better functionality for pairs of characters e.g. remove both sides of the parentheses.

kendonB:
"remove lease until from" doesn't work here
```
from castervoice.lib.dfplus.merge.mergerule import MergeRule|
```
"replace lease dot with clamor" also doesn't work

kendonB:
To structure the testing burden a bit better let's develop a "script" for people to use which includes the start state of the text with a "|" indicating the start position, then the expected resultant state. These could eventually become unit tests if we figure out how to do that.

kendonB:
Once we merge this let's also create a canonical character dictionary so that those definitions aren't doubled up.

alexboche:
> "remove lease until from" doesn't work here
> 
> ```
> from castervoice.lib.dfplus.merge.mergerule import MergeRule|
> ```
> That works for me. Though now I'm having some problems again texstudio.

kendonB:
Word users might have make these changes to get this to work: https://superuser.com/questions/962710/how-to-make-microsoft-word-selection-behave-like-it-would-in-a-plain-text-editor

kendonB:
```
other planting rates between|
```

"remove lease other" doesn't work here. It's got something to do with it being at the start of the line

kendonB:
```
"hello", "alex"|
```
"go lease before quotes" does nothing here

alexboche:
> "hello", "alex"|

This is because I have some punctuation missing from the punctuation list in text_manipulation_functions.py including double quotes. I will update that and also the dictionary for the choice object which is missing some stuff. I am experiencing the same problem as you with using the "remove" command at the beginning of the line.I will look into it.

alexboche:
I fixed two different bugs at the beginning of lines, one for "remove" and one for "remove until".I also added more punctuation.Note that punctuation should be put into not just the dictionary for the choice object but also `character_list` at the top of text_manipulation.py . Being in `character_list` allows you to match it even when it's in the middle of the word ( using a different regex). So for example you can use "go lease arch" to get to the "a" in "hat".

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,260 @@
...
+                "(left prekris | lay)": "(",
+                "(right prekris | ray)": ")",
+                "(left brax | lack)": "[",
+                "(right brax | rack)": "]",
+                "(left angle | lang)": "<",
+                "(right angle | rang)": ">",
+                "(left curly | lace)": "{",
+                "(right curly | race)": "}",
+                "quotes": '"',
+                "(single quote | thin quote)": "'",
```
kendonB: Need the plurals to be consistent here. I'd have `(quotes | quote)` for the first one and `(single | thin) (quote | quotes)` for the second

kendonB: Actually we really need the language to be consistent here with regular caster. "quotes" refers to a pair and we don't have a word for a single quote. We need all of these in `punctuation.py` and ideally we would just rewrite `punctuation.py` to have this dictionary style and we'd just import it here.

kendonB: I'm going to have a go at refactoring `punctuation.py` to make it usable here

kendonB: You can try out the character dicts here https://github.com/dictation-toolbox/Caster/pull/548

kendonB:
@alexboche you can get my PR altering the punctuation dicts using:

```
git fetch https://github.com/dictation-toolbox/Caster.git pull/548/head
git checkout -b dictation-toolbox/Caster/pull/548 FETCH_HEAD
```

kendonB:
I believe you can then do:
```
from castervoice.lib.ccr.core.punctuation import text_punc_dict, double_text_punc_dict
```

kendonB:
See here: https://github.com/kendonB/caster/blob/what_i_use/castervoice/lib/ccr/core/text_manipulation.py

it works for me

kendonB:
```
.
,|
```
"go lease two after dot" fails here

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,260 @@
...
+            "seventh": 7,
+            "eighth": 8,
+            "ninth": 9,
+            "tenth": 10,
+        }),
+        
+        
+    ]
+    defaults = {"n": 1, "m": 1, "spec": "", "dict": "", "text": "", "mouse_button": "", 
+        "horizontal_distance": 0, "vertical_distance": 0,
```
kendonB: Where do these get used?

kendonB: mouse_button, hor_dist, vert_dist etc

alexboche:
> I'm also a bit concerned about the use of the multi clipboard. I'm sure there are people using the second slot. Can you sneak these into a different much larger slot? Like the 6000th? Or whatever is the highest available slot that won't compromise performance?
 


I don't know how to do that using the regular clipboard. Previously we were using the caster clipboard functions read_selected_text_without_altering_clipboard and paste_string_without_altering_clipboard which do not have this problem but seemed unreliable. I just tried using those again and they seem to work pretty well actually accept with the occasional failure in Lyx and or tex studio. The latest commit replaces the regular copy-paste-pyperclip stuff with those caster functions, leaving the former commented out. I'm curious to know how reliable this method is for you.

I think we're doing with the punctuation is great; I haven't quite gotten to that yet.
Eventually would probably also be good to change the approach used for tex studio. Right now it is pasting the selected text over itself in order to unselect. A safer method would probably be best to just press right and then left to unselect – that would require redoing the stuff for getting the cursor back in the right spot.

Unfortunately, my voice has been feeling increasingly strained so I think I need to take a little break from voice coding for a few days. I'll try to get back to this sooner rather than later. Feel free to make any changes if you want.

kendonB:
@alexboche did you end up asking your question on SO? Ref: https://superuser.com/questions/1443670/is-there-a-way-to-read-selected-text-into-a-python-function-in-windows-without-a

alexboche:
> @alexboche391 did you end up asking your question on SO? Ref: [https://superuser.com/questions/1443670/is-there-a-way-to-read-selected-text-into-a-python-function-in-windows-without-a392](https://superuser.com/questions/1443670/is-there-a-way-to-read-selected-text-into-a-python-function-in-windows-without-a)

https://stackoverflow.com/questions/56414636/reading-selected-text-into-a-python-function-without-altering-the-clipboard-not

alexboche:
It turns out the caster clipboard functions you seem to have the capability to solve the multi clipboard problem if you get the proper sleep time,but this very inconsistent especially with the function that reads the text.

alexboche:
I have reorganized the application dependent behavior so that it will more easy for people to customize the pause times for different applications. Some applications (e.g. VS code) allow very little pause time (Tex studio and Lyx require more). 
I think this is ready for another round of testing.

kendonB:
One more for the wish list: "sauce" and "dunce" should exclude the current line from the search.

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,169 @@
...
+            rdescript="Text Manipulation: select until chosen character"),
+        
+        "remove <direction> [<number_of_lines_to_search>] until [<before_after>] [<occurrence_number>] <dictation>":
+            R(Function(text_manipulation_functions.copypaste_delete_until_phrase,
+                       dict(dictation="phrase")), 
+              rdescript="Text Manipulation: delete until chosen phrase"),
+        "remove <direction> [<number_of_lines_to_search>] until [<before_after>] [<occurrence_number>] <character>":
+            R(Function(text_manipulation_functions.copypaste_delete_until_phrase,
+                       dict(character="phrase")),
+              rdescript="Text Manipulation: delete until chosen character"),
```
kendonB: Move these two removes up with the other two

kendonB:
"go lease after dot" isn't working in Notepad++ or texstudio

kendonB:
```
failed to copy text
ERROR:action:Exception from function move_until_phrase:
  File "C:\Python27\lib\site-packages\dragonfly\actions\action_function.py", line 136, in _execute
    self._function(**arguments)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\text_manipulation_functions.py", line 187, in move_until_phrase
    match_index = get_start_end_position(selected_text, phrase, direction, occurrence_number)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\text_manipulation_functions.py", line 38, in get_start_end_position
    if not re.search(pattern, text.lower()):
AttributeError: 'NoneType' object has no attribute 'lower'
ERROR:action.exec:Execution failed: Function(move_until_phrase): 'NoneType' object has no attribute 'lower'
```

kendonB:
After testing a couple more times I no longer get this error. Might have been fleeting somehow.

castervoice/lib/context.py
```diff
@@ -121,8 +122,9 @@ def read_selected_without_altering_clipboard(same_is_okay=False):
         Clipboard.set_system_text("")
 
         Key("c-c").execute()
+        Pause(pause_time).execute()
         time.sleep(settings.SETTINGS["miscellaneous"]["keypress_wait"]/
-                   1000.)  # time for keypress to execute
+                   1000)  # time for keypress to execute
```
kendonB: Seems like this needs the "." for another function I use.

castervoice/lib/ccr/core/text_manipulation.py
```diff
@@ -0,0 +1,169 @@
...
+        "go <direction> [<before_after>] [<number_of_lines_to_search>] [<occurrence_number>] <character>":
+            R(Function(text_manipulation_functions.move_until_phrase,
+                       dict(character="phrase")),
+              rdescript="Text Manipulation: move to chosen character to the left of the cursor"),
+
+        "grab <direction> [<number_of_lines_to_search>] [<occurrence_number>] <dictation>":
+            R(Function(text_manipulation_functions.select_phrase, 
+            dict(dictation="phrase")), 
+                 rdescript="Text Manipulation: select chosen phrase"),
+        "grab <direction> [<number_of_lines_to_search>] [<occurrence_number>] <character>":
```
kendonB: Remind me of the use case for "grab"? I haven't been using it nor come across a situation where one of the other functions wasn't more suitable.

alexboche:
I tested other commands that use read_selected_without_altering_clipboard or paste_string_without_altering_clipboard : hug commands, bring me commands, alias, jump in. they all seem to work fine with the slight modification of those functions

castervoice/lib/ccr/core/text_manipulation.py
alexboche: "Grab" is for select. Dragon always prioritizes its native commands "select", which is why I called it "grab" instead of "select". I have added some better comments in addition to the rdescript

alexboche:
So I just noticed that previous caster functions that depend on read_selected_without_altering_clipboard have a loop that checks if the clipboard has changed, and if not tries to read the text again. I just tried that and it didn't seem to help so far but I might work on it a bit more. I think that is better left for the second release of these.

castervoice/lib/ccr/core/text_manipulation.py
kendonB: Nm cut and paste is a good enough use-case that I wasn't thinking about.

kendonB:
```
test|
```
say "grab ross until echo"
I get:
```
Text Manipulation: select until chosen phrase
failed to copy text
ERROR:action:Exception from function select_until_phrase:
  File "C:\Python27\lib\site-packages\dragonfly\actions\action_function.py", line 136, in _execute
    self._function(**arguments)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\text_manipulation_functions.py", line 374, in select_until_phrase
    match_index = get_start_end_position(selected_text, phrase, direction, occurrence_number)
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\text_manipulation_functions.py", line 38, in get_start_end_position
    if not re.search(pattern, text.lower()):
AttributeError: 'NoneType' object has no attribute 'lower'
ERROR:action.exec:Execution failed: Function(select_until_phrase): 'NoneType' object has no attribute 'lower'
```
I imagine this will be a pretty common error so we should handle it.

kendonB:
let's change it so that the default number of lines for "sauce" and "dunce" is 1

kendonB:
I know making documentation is the worst part of all this so I'm going to add a placeholder page on readthedocs with the basics of what people need to know to get started using these.

kendonB:
Found another possibly related bug (I haven't been able to work out a way to make this more minimal):

```
- Selecting a single element: `grab <direction> [<number_of_lines_to_search>] [<occurrence_number>] <dictation>`
- Selecting from the cursor to a target: `grab <direction> [<number_of_lines_to_search>] until [<before_after>] [<occurrence_number>] <dictation>`
|
```
I say "go sauce two before from" in Notepad.exe. This one is super weird for me as I can get it to work just by changing a single character (removing the "a" in "dictation" for example). This is definitely a weird bug and may well be environmental somehow. I wonder if it's because of some checking against the clipboard somewhere?

alexboche:
I think I fixed that error were getting. Note that certain applications such as text editors will copy the line if you press control C without having anything selected on that line. That can cause problems if you try to do one of these commands in the wrong direction when you're at the end of the line. I don't know of any easy way to fix that, but I also don't see why anybody would be using the command and that way except by accident (so I don't think it's a serious problem).

alexboche:
I made the default number_of_lines_to_search for sauce and dunce 3 (which is four including the current line).

kendonB:
Is it possible to exclude the current line when using "sauce" and "dunce"?

castervoice/lib/ccr/core/nav.py
```diff
@@ -268,6 +268,7 @@ class Navigation(MergeRule):
         IntegerRefST("nnavi50", 1, 50),
         IntegerRefST("nnavi500", 1, 500),
         Dictation("textnv"),
+        Dictation("dictation"),
```
kendonB: remove this?

castervoice/lib/ccr/core/nav.py
```diff
@@ -323,8 +324,10 @@ class Navigation(MergeRule):
             "lease": "backspace",
             "ross": "delete",
         }),
+        
     ]
 
+
```
kendonB: remove these added spaces?

kendonB:
In RStudio, starting here
```
test <- c(1, 3)
  |
```
I say "go sauce before right prekris" and I end up here:
```
test <- c(1, |3)
  
```
I haven't been able to replicate this behavior in any other apps.

alexboche:
> Is it possible to exclude the current line when using "sauce" and "dunce"?

Yes, but then we would not be able to get the cursor back into the original spot.

kendonB:
> > Is it possible to exclude the current line when using "sauce" and "dunce"?
> 
> Yes, but then we would not be able to get the cursor back into the original spot.

Seems like a solvable problem but probably not easy. Let's leave that for V2.

alexboche:
> > > Is it possible to exclude the current line when using "sauce" and "dunce"?
> > 
> > 
> > Yes, but then we would not be able to get the cursor back into the original spot.
> 
> Seems like a solvable problem but probably not easy. Let's leave that for V2.

It probably won't be that hard if we just get the right regex so it only matches if it sees
character first. but we will leave it for next time– I will look into it

alexboche:
> In RStudio, starting here
> 
> ```
> test <- c(1, 3)
>   |
> ```
> 
> I say "go sauce before right prekris" and I end up here:
> 
> ```
> test <- c(1, |3)
>   
> ```
> 
> I haven't been able to replicate this behavior in any other apps.

I don't totally understand what's going on here. Part of the problem seems to be that if you have two spaces in a row on a blank line (and sometimes even on a non-blank line,  this seems two vary consistent), Rstudio  will move across both of them in a single arrow-key press.
Not sure that's the whole story though. Let's say I have the following (with two or three spaces before the cursor, doesn't matter as long as it's more than one)
 ```
test <- c(1, 3)
  |
```
and I say "go sauce before right prekris", I get the same problem as you with the result
```
test <- c(1, |3)
```
But if I say, "go sauce before echo", I get the desired result:
```
t|est <- c(1, 3)
```
It doesn't seem to be anything specific to "right prekris" since I was able to reproduce the error in another context with "go lease before echo" (I don't remember the exact context).

I also get another bug. If I have
```
tet  st < - c(1,|3)
```
and I say "go lease before sierra" I think I got
```
tet  |st < - c(1,3)
```
and then I started just getting that it didn't move the cursor after selecting and printed
"failed to copy" "no text to select"

**Comment by alexboche** [comment]
> > > Is it possible to exclude the current line when using "sauce" and "dunce"?

> > 
> > Yes, but then we would not be able to get the cursor back into the original spot.
> 
> Seems like a solvable problem but probably not easy. Let's leave that for V2.

It probably won't be that hard if we just get the right regex so it only matches if it sees
character first. but we will leave it for next time– I will look into it

**Comment by kendonB** [comment]
```
specifications are indicated in bold text. The Ro labeled|
```
I say, "go lease after oscar" here and I get the "failed to copy text; no text to select" error. I get this problem both in Microsoft Word as well as in this text box. You have to remove the pipe symbol for it to fail.

File: castervoice/lib/text_manipulation_functions.py [hunk 18]
```diff
@@ -0,0 +1,476 @@
...
+        try:
+            match = match_index_list[occurrence_number - 1] # count from the left
+        except IndexError:
+            print("There aren't that many occurrences of '{}'".format(phrase))
+            return 
+    left_index, right_index = match
+    return (left_index, right_index)
+    
+copy_pause_time_dict = {"standard": "10", "texstudio": "70", "lyx": "60"}
+paste_pause_time_dict = {"standard": "0", "texstudio": "100", "lyx": "20"}
```
kendonB: this doesn't seem to ever get used

alexboche: > this doesn't seem to ever get used

I'm having some error on Github so I can't see the file, but on my version those dictionaries are used in the functions text_manipulation_copy and the text_manipulation_paste.

kendonB:
@alexboche I figured out what was causing the bug I was most recently seeing. Had to make this change: https://github.com/dictation-toolbox/Caster/pull/485/commits/99d2fdd22c6f332412dde86ae359198b6c4f0c05

kendonB:
@alexboche I am ready to merge this. Please review the documentation and let me know what you think.

castervoice/lib/text_manipulation_functions.py
```diff
@@ -75,7 +75,7 @@ def text_manipulation_copy(application):
     ## Pause(copy_pause_time_dict[application]).execute()
     ## selected_text = pyperclip.paste()
     
-    err, selected_text = context.read_selected_without_altering_clipboard(pause_time=copy_pause_time_dict[application])
+    err, selected_text = context.read_selected_without_altering_clipboard(same_is_okay=True, pause_time=copy_pause_time_dict[application])
```
alexboche: Eventually, we may want change this back to same_is_okay=False eventually and use it to make it try copying again (with an increased pause time) if `err` == 1 which I think means that the clipboard hasn't changed.

kendonB: Can you mention this in the "Known issues" section?

kendonB: Is the function able to distinguish between a failed copy and successfully copying the thing that's sitting on the clipboard?

alexboche: It seems that it is not able to distinguish between those based on a quick test I did. Let me know if you have a different experience.

kendonB: I think it's still better to do `same_is_ok=True` and caveat that the pauses might mess stuff up from time to time.

PR #490, opened 2019-05-04
Added Mock dependencies, Travis, Dragonfly text engine tests

comodoro:
The idea is to make tests runnable with the Dragonfly text engine when no engine is running and to eventually get rid  of many dependencies, particularly OS specific. The tests still should not be called unit, but might be possible to run on Travis in the future


LexiconCode:
Thank you as this is something that we desperately need but I haven't had the time or energy to implement. Eventually will get to proper unit tests but this is a great first step!

LexiconCode:
@danesprite As mentioned in the past to use dragonfly's "engine.create_timer()" method instead of the implementation in dragonfly.timer. An error will be raised if you use another engine like the "text" engine.

I'd have to take a closer look at your implementation but I figured I'd pass it along quick.

Danesprite:
This is a good idea. I'm happy to help fix Caster's use of the `dragonfly.timer` module.

@comodoro I see that you are importing `Text` and `Key` from `dragonfly.os_dependent_mock`. This will not work with dragonfly2 versions >= 0.13.0. You could define mock actions in Caster instead. Perhaps in the `castervoice.lib.actions` module if the tests are running?

LexiconCode:
> This is a good idea. I'm happy to help fix Caster's use of the `dragonfly.timer` module.
That's generous of you, thanks

comodoro:
Thanks for the warning, I hope simply emulating it as per the last commit is enough.

Danesprite:
@comodoro No problem. Using `MockAction` should fix the problem.

@LexiconCode I'll do that in a separate PR sometime. Another problem with using `_Timer` is that it will conflict with any other natlink modules that use `engine.create_timer()` and the natlink timer callback will be overridden.

comodoro:
By the way @Danesprite I suppose there is no better way of determining running engine than `_engines_by_name`? I gather `get_engine` initializes one if none is present

comodoro:
OK, there is a lot of mocking and still when I spotted that Travis starts to support Windows, I went for it, That does not mean the tests should not work on Linux by now though. There is a Travis buildfile, the actual Travis connection has to be done by someone with appropriate rights. Passing the build is lenient enough at the moment I would imagine, pylint error check and the tests. I hope it will pass initially as debugging Travis builds has been a nightmare for me in the past.

LexiconCode:
@comodoro let me know when this is ready to merge!

requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+dragonfly2>=0.12.0
```
Danesprite: Just FYI, this should be `dragonfly2>=0.14.0` because Caster uses lists for `AppContext` executables in some files.

comodoro: Thanks! I just updated it in both files

comodoro:
@LexiconCode so now is the time as I just managed to get the Travis build pass:
https://travis-ci.org/comodoro/caster/builds/542818988
However as said before, someone with write access to Caster has to setup Travis on this end.

Looking at it and the changes I would feel better if someone tried it before merging, but waiting too long inevitably leads to merge conflicts and it works on my end without problems.

Danesprite:
@comodoro I think I have set it up correctly (https://travis-ci.org/dictation-toolbox/Caster). It should trigger a build if you commit again.

LexiconCode:
Thanks for setting that up @Danesprite!  @comodoro make another commit as a test to see if the configuration is successful. I'll try to merge as soon as possible.

kendonB:
Would it be easy to add in a page on readthedocs the outlines the process to add a new test?

comodoro:
@LexiconCode I will try, I hope it will be that easy, but I have doubts. I remember having to fiddle with setup in Travis, somehow triggering failed build on master (which does not even have `.travis.yml`) before branches started to build

comodoro:
And it does work. Yay:)

LexiconCode:
`ImportError: No module named mock`
@comodoro I think `mock` will have to be added to requirements.txt and setup.py

comodoro:
@LexiconCode I thought about tests as development feature (hence `requirements-dev.txt`), but I think you are right - in order to preserve previous behavior. I will add it.

LexiconCode:
Yeah that's what I originally thought but due to the way we load dependencies I think it expects it.

Issue #426, opened 2019-03-24
Change default "Redo" to `cs-z`


kendonB:
AFAIK, more text editors on Windows use `cs-z` and not `c-y` for "Redo". Though I may be wrong. However, "Redo" should be included in each App, if the app redo differs from the default in `nav.py`.

So, this issue is about:
 - deciding whether to use `cs-z` or `c-y` as the default
 - including the other one in all apps for which the default does not work

LexiconCode:
That's Reasonable.
Why don't we take the time to test each app. To see if `cs-z` is the preferred method  "Redo". As you test post name of the apps and I'll check them off here.

List of Apps.
- [ ] atom
- [ ] chrome
- [ ] eclipse
- [ ] emacs
- [ ] excel
- [ ] explorer
- [ ] firefox
- [ ] flashdevelop
- [ ] foxitreader
- [ ] gitter
- [ ] jetbrains
- [ ] kdiff3
- [ ] lyx
- [ ] msvc
- [ ] notepadplusplus
- [ ] outlook
- [ ] rstudio
- [ ] sqldeveloper
- [ ] ssms
- [ ] sublime
- [ ] totalcmd
- [ ] typora
- [ ] visualstudio
- [ ] vscode
- [ ] winword

kendonB:
Some apps allow for both, so there should be two lists.

```
rstudio:         cs-z works and c-y fails
chrome:          cs-z works and c-y works
firefox:         cs-z works and c-y works
notepadplusplus: cs-z works and c-y works
outlook:         cs-z fails and c-y works
winword:        cs-z fails and c-y works
excel:           cs-z fails and c-y works
explorer:        cs-z fails and c-y works
```

Danesprite:
There is an [aenea](https://github.com/dictation-toolbox/aenea) class I use in these situations called [ContextAction](https://github.com/dictation-toolbox/aenea/blob/ef8ef3d51df1e58aba6eec3ef5e03400646625de/client/aenea/wrappers.py#L273). It executes different actions depending on the first dragonfly context in a list that matches, falling back on a default action. It isn't dependent on any part of aenea to work. 

An example for "redo" would be:
```Python
redo = ContextAction(default=Key("c-y"), actions=[
    # Use cs-z for rstudio
    (AppContext(executable="rstudio"), Key("cs-z")),
])

```

I find it very useful as an Emacs user. Maybe caster could use something like this?

LexiconCode:
[...]

PR #493, opened 2019-05-05
add ContextAction for redo

kendonB:
Addresses https://github.com/dictation-toolbox/Caster/issues/426.

note that this requires updating dragonfly2 to the latest version.

- [x] @kendonB test exact branch on home computer, including repeats
- [x] Someone else test it too with rstudio and one other app, including repeats


kendonB:
Note that next time I will create the new branch on my own fork. I just need to work out how to do that

kendonB:
Looks like the repeats aren't working. @LexiconCode, can you see what I've done wrong?

LexiconCode:
I'll see if I get this tested tomorrow!

LexiconCode:
This works as expected. You can add the bolded applications in #426.

castervoice/lib/ccr/core/nav.py
```diff
@@ -95,7 +96,10 @@ class NavigationNon(MappingRule):
         "undo [<n>]":
             R(Key("c-z"), rdescript="Core: Undo")*Repeat(extra="n"),
         "redo [<n>]":
-            R(Key("c-y"), rdescript="Core: Redo")*Repeat(extra="n"),
+            R(ContextAction(default=Key("c-y")*Repeat(extra="n"), actions=[
+                # Use cs-z for rstudio
+                (AppContext(executable="rstudio"), Key("cs-z")*Repeat(extra="n")),
```
lexxish: This should be in app grammar. Jetbrains is also cs-z for example

kendonB: Using contextaction is an alternative to using app grammars.

kendonB: Will add jetbrains. Does c-y not work there?

lexxish: I meant from a design perspective since otherwise the code will get polluted with app specific if statements

kendonB: In this case where there are just a few exceptions to the common key mapping, I think I prefer ContextAction to adding the exceptions to App grammars. Let's see how it goes doing it this way and if the core grammars start to engorge we can move stuff into app files.

kendonB: @lexxish, The apps can be specified as a list as well, so adding a new app to a ContextAction is super compact. Now that I know this, I'm skeptical that it's better to have app grammars at all.

PR #514, opened 2019-05-14
Include two commands for right mouse down (lean) and up (hoist).

codingApprentice:
Since many PC games use the right mouse button (holding down and releasing) to move characters around, I thought it would be useful to include these two extra commands.
I've also updated the documentation accordingly to include these commands.

Feel free to edit if you think they need adjusting.


kendonB:
Do you like "press" and "hoist" better than "sigh squat" and "sigh bench"? Similar to "kick" vs "psychic". Is it important that these are single syllable?

kendonB:
If PC gaming is the main use case for this then probably they should be single syllable yeah?

kendonB:
I have tested this and it seems to work just like the real mouse.

kendonB:
I'm happy to merge this once we've settled on the spec.

If it's important that these are single syllable: Let's change the "press" spec as the Dragon command for pressing keys is "press key name" and could get weird possibly get weird in continuous recognition. Possibly "ram" or "duck".

If it's not important that these are single syllable, I vote "sigh squat" and "sigh bench".

codingApprentice:
I feel like since it is a common action in gaming, it should be one syllable to reduce vocal fatigue.
How about "lean" instead of "press"?

kendonB:
Sounds great! Once I work out how to "squash merge" these changes I will merge. Thanks heaps!

PR #515, opened 2019-05-15
Text manipulation 2.0

alexboche:
Sorry I don't know how to put this on the other pull request #485  for text manipulation because this is from a different repository on my computer.I think the problem is that the text does not have enough time to get onto the clipboard after pressing control-c. Accordingly I added a long pause after pressing control-c. If this still doesn't work you may want to make the pause even longer. I also put a debugging print statement in the move_until_phrase function

### DIFF
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -256,6 +256,79 @@ class Navigation(MergeRule):
             R(Function(text_utils.enclose_selected)),
         "dredge":
             R(Key("a-tab")),
+    
+
+        
+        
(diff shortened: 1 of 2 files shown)


kendonB:
Yes this seems to work now. Can you add these changes to the other pull request? What's the barrier? Have you added your fork as a remote? Do you know how to check out a remote branch? Happy to help you get this done here if Google is unhelpful.

alexboche:
Yeah I think I can do it as part of the other pull request later tonight

alexboche:
I pushed the changes to the other will request

PR #518, opened 2019-05-16
add varepsilon to latex

kendonB:


kendonB:
This one should be uncontroversial; cc @mrob95

PR #519, opened 2019-05-16
Remove the '+' from the `dev pause' command.

alexboche:
I first put in the '+' because the Pause action is often used in conjunction with other actions and so often requires the plus before it. But after more experience using these commands, I think it is better just to be consistent and not put the plus in front.
(by the way, this is my first ball request made using the gui as opposed to the command line so it works)


kendonB:
I agree with this change and it is trivial. Merging

PR #521, opened 2019-05-16
Change alt tab command `dredge` to go multiple windows back

alexboche:
Right now the alt tab command `dredge` only lets you press tab once to get to the most recent window. this change will let you go to the second or third (or more) most recent window too.


kendonB:
Can you change the folder names so that they match the repo?

I see: `castervoice25/lib26/ccr27/core28/nav.py`

kendonB:
Tested this and it works. merging

Issue #516, opened 2019-05-15
Caster fails on startup: KeyError: 'PROCESSOR_ARCHITEW6432'


rakeshnbabu:
Hello,

When starting up caster, dragonfly/lib/os.py threw a KeyError: 'PROCESSOR_ARCHITEW6432' The full error output is below:

```
(dragonfly) C:\Users\rbabu01\caster\Caster-master>python _caster.py
Searching Windows Registry For DNS...
Traceback (most recent call last):
  File "_caster.py", line 12, in <module>
    from caster.lib.ccr.standard import SymbolSpecs
  File "C:\Users\rbabu01\caster\Caster-master\caster\lib\ccr\__init__.py", line 3, in <module>
    from caster.lib import utilities
  File "C:\Users\rbabu01\caster\Caster-master\caster\lib\utilities.py", line 27, in <module>
    from caster.lib import settings
  File "C:\Users\rbabu01\caster\Caster-master\caster\lib\settings.py", line 139, in <module>
    "ENGINE_PATH": _validate_engine_path(),
  File "C:\Users\rbabu01\caster\Caster-master\caster\lib\settings.py", line 62, in _validate_engine_path
    return _find_natspeak()
  File "C:\Users\rbabu01\caster\Caster-master\caster\lib\settings.py", line 71, in _find_natspeak
    proc_arch64 = os.environ['PROCESSOR_ARCHITEW6432'].lower()
  File "C:\Users\rbabu01\AppData\Local\Continuum\anaconda3\envs\dragonfly\lib\os.py", line 425, in __getitem__
    return self.data[key.upper()]
KeyError: 'PROCESSOR_ARCHITEW6432'
```
I traced the error back to Caster-master\caster\lib\settings.py", line 71, in _find_natspeak
    proc_arch64 = os.environ['PROCESSOR_ARCHITEW6432'].lower() . The issue was that the environment variable "PROCESSOR_ARCHITEW6432" did not exist.

I am running Windows 10 64bit.
```
(dragonfly) C:\Users\rbabu01\caster\Caster-master>python
Python 2.7.16 |Anaconda, Inc.| (default, Mar 14 2019, 15:42:17) [MSC v.1500 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.environ.keys()
[...]

PR #524, opened 2019-05-17
Handle case when 'PROCESSOR_ARCHITEW6432' is not an environment var

rakeshnbabu:
Addresses #516. Added error handling for case when 'PROCESSOR_ARCHITEW6432' is not set as an environment variable.


LexiconCode:
I must've missed the notification so apologies on making the dupe. Thanks again for the contribution!

Issue #520, opened 2019-05-16
Refactor apps grammars to have standard grammar


lexxish:
Similar to how we have a standard for CCRs (https://github.com/dictation-toolbox/Caster/blob/develop/castervoice/lib/ccr/standard.py), we need to create a similar standard for applications. The different domains of applications would be for example IDE, Browser, WordProcessor, and each implementation would have standard grammar in common. For example, both Chrome and Firefox would be implementations of the browser grammar and they would both contain "back", "forward", etc.

LexiconCode:
I'm in favor of this as it simplifies documentation and makes it easier to utilize filter rules. However this will change the command names across all application grammars which would be disruptive to the community.

kendonB:
I think we should make a call on whether we do this after assessing the extent to which there actually is conflict across apps. I believe that there's not that much conflict

lexxish:
If there ends up being too many conflicts it could be implemented as a refactoring. That's what I intended initially anyway, but a rewrite would be much cleaner and better long-term.

kendonB:
`ContextAction` is very compact: https://github.com/dictation-toolbox/Caster/pull/493/files

Let's consider whether we should even have the standard elements redefined within App Grammars or if they should all be done using `ContextAction`s. I think it would be much tidier to pull all of the common stuff out of the app grammars and just use those for app-specific commands.

kendonB:
The only downside I see is that it makes it a little less transparent to new users how to add a new app. Since that's relatively rare, I still think it's the right way to go.

LexiconCode:
Work in progress but the initial commit is done we will open up other issues as we convert.

PR #530, opened 2019-05-20
Refactor apps grammars to have standard grammar  #520

lexxish:
Plan to standardize "browser" and "integrated development environment". So far have drafted standardization of "integrated development environment" and changed "jet brains" application to use the new standard.


castervoice/lib/ccr/core/nav.py
```diff
@@ -245,7 +245,7 @@ class Navigation(MergeRule):
             R(Function(textformat.clear_text_format)),
         "peek [<big>] format":
             R(Function(textformat.peek_text_format)),
-        "(<capitalization> <spacing> | <capitalization> | <spacing>) (bow|bowel) <textnv> [brunt]":
+        "(<capitalization> <spacing> | <capitalization> | <spacing>) [(bow|bowel)] <textnv> [brunt]":
```
lexxish: Are people are okay with making bowel optional?

_caster.py
```diff
@@ -36,7 +36,7 @@ def _wait_for_wsr_activation():
 from castervoice.lib import control
 _NEXUS = control.nexus()
 
-from castervoice.apps import *
+from castervoice.apps import __init__
```
lexxish: apps/__init__ has different functionality in this patch. It now pulls in all the files looking under the directory /apps. If you call `from castervoice.apps import *` instead it will error out (I believe because it tries to import twice).

castervoice/lib/ccr/core/nav.py
LexiconCode: I would be fine with that but let's open it up in a separate pull request. That way can be discussed separate from this request.

castervoice/apps/ide/jetbrains.py
```diff
@@ -0,0 +1,142 @@
...
+    else:
+        jet_brains_rule = JetbrainsRule(name="jet brains")
+        gfilter.run_on(jet_brains_rule)
+        grammar.add_rule(jet_brains_rule)
+
+        # You will need to change your terminal to bash under settings->tools
+        # C:\Program Files\Git\bin\bash.exe
+        git_rule = GitBashRule()
+        gfilter.run_on(git_rule)
+        grammar.add_rule(git_rule)
```
lexxish: @LexiconCode  thoughts on merging here? I think we can actually create a general git rule and merge it into most editors since most of them have built-in terminals.

lexxish:
standardized chrome, firefox and jetbrains

LexiconCode:
You'll need to edit [Application_Commands_Quick_Reference.md](https://github.com/dictation-toolbox/Caster/blob/develop/castervoice/doc/readthedocs/Application_Commands_Quick_Reference.md) to reflect the the commands for the applications that you've changed.

lexxish:
added docs

LexiconCode:
what's the purpose of `jetbrains.ipynb`?

lexxish:
> what's the purpose of `jetbrains.ipynb`?

Accidental commit. Removed it

_caster.py
LexiconCode: Remind me of the downside of your current implementation? 
The reason we have it set up this way as it checks for errors in Python files and excludes them loading of caster. Otherwise it crashes the whole system.

castervoice/apps/browser/chrome_and_firefox.py
```diff
@@ -0,0 +1,98 @@
+#
```
LexiconCode: I would recommend renaming this to 'browser_standard'.

As people add other browsers this is the basic functionality we can expect them to implement. We can add in the comments and documentation which browsers this currently supports.

lexxish: how is BrowserSharedCommands?

lexxish:
This isn't strictly a re-factor I should add. Overall I added a ton of commands the Firefox and brought it up to par with chrome. Also added a lot of jet brains commands and possibly renamed a few. Lastly it makes it possible to have nested directories under /apps and still load them. Also serves as pre-work for https://github.com/dictation-toolbox/Caster/issues/533 and https://github.com/dictation-toolbox/Caster/issues/534 as I will re-use the grammars

_caster.py
lexxish: No real downside that I'm aware of. Still verifies the rules and will only fail in a single rule at a time. It just is a little strange that I had to change it. I'm thinking it's because for example chrome.py imports browser_shared_commands.py which means if we used a * here we would import it twice, but not really sure. I could test it more and try to "fix" it but not sure it's worth it.

castervoice/apps/browser/chrome_and_firefox.py
LexiconCode: Personally I like standard better but I suppose it works.

lexxish: The standard is in browser.py not. This file is just commands that have the same hotkeys between browsers

LexiconCode: ok

kendonB:
One thing to note here is that there are a bunch of commands in `nav.py`~that naturally fit in a universal app grammar. Things like "save", "find" etc. are not really about navigation. They are:

```
        'save':
            R(Key("c-s"), rspec="save"),
        "find":
            R(Key("c-f")),
        "find next [<n>]":
            R(Key("f3"))*Repeat(extra="n"),
        "find prior [<n>]":
            R(Key("s-f3"))*Repeat(extra="n"),
        "find everywhere":
            R(Key("cs-f")),
        "replace":
            R(Key("c-h")),
        "[show] context menu":
            R(Key("s-f10")),
       "refresh":
            R(Key("c-r")),
         "refresh":
            R(Key("c-r")),
        "maxiwin":
            R(Key("w-up")),
        "move window":
            R(Key("a-space, r, a-space, m")),
        "window (left | lease) [<n>]":
            R(Key("w-left"))*Repeat(extra="n"),
        "window (right | ross) [<n>]":
            R(Key("w-right"))*Repeat(extra="n"),
        "monitor (left | lease) [<n>]":
            R(Key("sw-left"))*Repeat(extra="n"),
        "monitor (right | ross) [<n>]":
            R(Key("sw-right"))*Repeat(extra="n"),
        "(next | prior) window":
            R(Key("ca-tab, enter")),
        "switch (window | windows)":
            R(Key("ca-tab"))*Repeat(extra="n"),
        "next tab [<n>]":
            R(Key("c-pgdown"))*Repeat(extra="n"),
        "prior tab [<n>]":
            R(Key("c-pgup"))*Repeat(extra="n"),
        "close tab [<n>]":
            R(Key("c-w/20"))*Repeat(extra="n"),
        "undo [<nnavi10>]":
            R(Key("c-z"))*Repeat(extra="nnavi10"),
        "redo [<nnavi10>]":
            R(ContextAction(default=Key("c-y")*Repeat(extra="nnavi10"), actions=[
                (AppContext(executable=["rstudio", "foxitreader"]), Key("cs-z")*Repeat(extra="nnavi10")),
                ])),
```
I think that any command that might be different across apps should be in this standard app grammar file. This PR should remove these commands from nav.py.

LexiconCode:
> One thing to note here is that there are a bunch of commands in `nav.py`~that naturally fit in a universal app grammar. Things like "save", "find" etc. are not really about navigation. They are:
> ....
> 
> I think that any command that might be different across apps should be in this standard app grammar file. This PR should remove these commands from nav.py.

They might be defined in the universal grammar but that's what nav.py is in this case. App grammar are defined. The current pull requests purpose isn't to provide a universal grammar commands and key shortcuts but the universal command name. Possibly an exception to this is like Firefox and Chrome which overlap keyboard shortcuts.

Ultimately this allows us to set up an API which were we can change the command and its action.

lexxish:
I had to prioritize getting my environment up and running. Is this close to something that we'd merge as is and can iterate on to reduce conflicts? One thing I would change first is removing above and below for up and down as I've noticed Dragon misinterpreting these as other commands.

lexxish:
> > One thing to note here is that there are a bunch of commands in `nav.py`~that naturally fit in a universal app grammar. Things like "save", "find" etc. are not really about navigation. They are:
> > ....
> > I think that any command that might be different across apps should be in this standard app grammar file. This PR should remove these commands from nav.py.
> 
> They might be defined in the universal grammar but that's what nav.py is in this case. App grammar are defined. The current pull requests purpose isn't to provide a universal grammar commands and key shortcuts but the universal command name. Possibly an exception to this is like Firefox and Chrome which overlap keyboard shortcuts.
> 
> Ultimately this allows us to set up an API which were we can change the command and its action.

My understanding matched with what @kendonB was saying and I do have some (incomplete) code somewhere where I started pulling grammar out of nav.py. If we pull the grammar out one way to reference a given command is by the grammar itself (which is what the code in this pull request does). Even the extras can be pulled out as grammar. The link between the actions and the grammar are the names of the extras. Can you give an example of how you were thinking this would be implemented without pulling grammar out of nav.py @LexiconCode ?

LexiconCode:
> Even the extras can be pulled out as grammar. The link between the actions and the grammar are the names of the extras. Can you give an example of how you were thinking this would be implemented without pulling grammar out of nav.py @LexiconCode ?

Thanks for clarifying my understanding, I don't see any other way of implementing it. To further understand some of the core grammars like 'Nav.py' are going to be subsumed into standard app grammar file. In that case what is the criteria determining what we've traditionally called a "core" versus vs "standard app" grammars?
 
Previously because it was clear cut between `core` and `app` this made documentation and setting the user's expectations for contributing additions straightforward.

lexxish:
@LexiconCode should be ready to merge.

I think we would need to document a grammar architecture with future changes that impact contributing/documenting. For this case, we haven't impacted them too much at least.

LexiconCode:
Sounds like a plan I'm not going to be able to get to it till this weekend.

castervoice/apps/browser/chrome.py
```diff
@@ -0,0 +1,81 @@
...
+from castervoice.lib.dfplus.state.short import R
+from castervoice.lib.temporary import Store, Retrieve
+
+
+class ChromeRule(BrowserSharedCommands):
+    pronunciation = "google chrome"
+
+    _mapping = {
+        browser.OPEN_NEW_WINDOW:
+            R(Key("c-n")),
```
kendonB: This and many others is in `browser_shared_commands`. Why is it redefined here?

castervoice/apps/ide/ide.py
```diff
@@ -0,0 +1,94 @@
...
+PREVIOUS_ERROR = "(%s error|error %s)" % (BACK, LEFT)
+REDO = "redo [<n>]"
+SHOW_SETTINGS = "[show] settings"
+RENAME_CURRENT_FILE = "file rename | rename file"
+
+# window navigation
+NEXT_TAB = "%s tab [<n>]|tab %s [<n>]" % (FORWARD, RIGHT)
+PREVIOUS_TAB = "%s tab [<n>]|tab %s [<n>]" % (BACK, LEFT)
+CLOSE_TAB_N_TIMES = "close tab [<n>]|tab close [<n>]"
+GO_TO_EDITOR = "go [to] editor"
```
kendonB: I think we should keep "go" for text navigation and use "focus" for moving between areas of the IDE.

castervoice/apps/shared/directions.py
```diff
@@ -0,0 +1,10 @@
+#
+# __author__ = "lexxish"
+#
+
+RIGHT = "(right|sauce)"
+LEFT = "(left|lease)"
+UP = "(up|sauce)"
+DOWN = "(down|dunce)"
+FORWARD = "(%s|next|forward)" % RIGHT
+BACK = "(%s|back|prev|previous)" % LEFT
```
kendonB: prior should be an option here

castervoice/apps/browser/browser.py
```diff
@@ -0,0 +1,67 @@
...
+SWITCH_TO_LAST_TAB = "last tab"
+SWITCH_TO_SECOND_TO_LAST_TAB = "second last tab"
+GO_FORWARD_N_TIMES = "go %s [<n>]" % FORWARD
+GO_BACK_N_TIMES = "go %s [<n>]" % BACK
+ZOOM_IN_N_TIMES = "zoom in [<n>]"
+ZOOM_OUT_N_TIMES = "zoom out [<n>]"
+ZOOM_RESET_DEFAULT = "zoom reset"
+FORCE_HARD_REFRESH = "(hard refresh|super refresh)"
+FIND_NEXT_MATCH = "find %s [match] [<n>]" % BACK
+FIND_PREVIOUS_MATCH = "find %s [match] [<n>]" % FORWARD
```
kendonB: back and forward should be switched here

castervoice/apps/ide/ide.py
```diff
@@ -0,0 +1,94 @@
...
+RUN_CURRENT_FILE = "run file"
+NEXT_ERROR = "(%s error|error %s)" % (FORWARD, RIGHT)
+PREVIOUS_ERROR = "(%s error|error %s)" % (BACK, LEFT)
+REDO = "redo [<n>]"
+SHOW_SETTINGS = "[show] settings"
+RENAME_CURRENT_FILE = "file rename | rename file"
+
+# window navigation
+NEXT_TAB = "%s tab [<n>]|tab %s [<n>]" % (FORWARD, RIGHT)
+PREVIOUS_TAB = "%s tab [<n>]|tab %s [<n>]" % (BACK, LEFT)
```
kendonB: There are plenty of commands that should be common between all apps like next tab and previous tab. Should these get pulled out of both into a shared_commands file that browser_shared_commands etc pull from?

castervoice/apps/ide/ide.py
kendonB: "go back" "go forward" in the browser/file dialogue are probably fine.

lexxish: Yes "go" might not be the best here. Do you prefer "focus" over "show"

castervoice/apps/browser/browser.py
lexxish: good catch

castervoice/apps/ide/ide.py
lexxish: yes I would think so

castervoice/apps/browser/chrome.py
lexxish: what do you mean many others? This is the only duplicate right?

castervoice/apps/browser/browser.py
```diff
@@ -61,9 +61,4 @@
         "fourth": "4",
         "fifth": "5",
         "sixth": "6",
-        "seventh": "7",
-        "eighth": "8",
-    }),
-    IntegerRefST("n", 1, 100),
-    IntegerRefST("m", 1, 10)
-]
+        "seventh"
```
kendonB: broken?

lexxish: how did that happen…

LexiconCode:
I will test this and merge tomorrow.

castervoice/apps/browser/firefox.py
```diff
@@ -0,0 +1,55 @@
...
+from castervoice.lib.dfplus.merge import gfilter
+from castervoice.lib.dfplus.state.short import R
+
+
+class FirefoxRule(BrowserSharedCommands):
+    pronunciation = "fire fox"
+
+    _mapping = {
+        browser.PREVIOUS_TAB_N_TIMES:
+            # control shift tab doesn't work and this appears to be an undocumented work new tab that new tab that around
```
kendonB: comment needs tidying

castervoice/apps/ide/ide.py
kendonB: focus is better. Many IDEs have different panes that are all visible (i.e. the editor may already be "shown" but not "focused"

castervoice/apps/ide/ide.py
```diff
@@ -0,0 +1,94 @@
...
+
+# editor management
+SPLIT_WINDOW_UP = "split [pane] %s" % UP
+SPLIT_WINDOW_DOWN = "split [pane] %s" % DOWN
+SPLIT_WINDOW_RIGHT = "split [pane] %s" % RIGHT
+SPLIT_WINDOW_LEFT = "split [pane] %s" % LEFT
+SPLIT_MOVE_UP = "pane %s [<n>]" % UP
+SPLIT_MOVE_DOWN = "pane %s [<n>]" % DOWN
+SPLIT_MOVE_RIGHT = "(pane %s|next pane) [<n>]" % RIGHT
+SPLIT_MOVE_LEFT = "(pane %s|(prior|previous) pane) [<n>]" % LEFT
```
kendonB: why not have all the BACK's here too (why only prior|previous?).

LexiconCode:
@lexxish Everything seems to test out wise and outside of the few comments being addressed it's ready to merge from my perspective.

castervoice/apps/ide/ide.py
lexxish: was intentional (thought they didn't sound right), but changed it

lexxish:
Updated based on comments. I think this is far from perfect btw, but it's at least an incremental improvement

castervoice/apps/ide/ide.py
```diff
@@ -0,0 +1,94 @@
...
+
+# editor management
+SPLIT_WINDOW_UP = "split [pane] %s" % UP
+SPLIT_WINDOW_DOWN = "split [pane] %s" % DOWN
+SPLIT_WINDOW_RIGHT = "split [pane] %s" % RIGHT
+SPLIT_WINDOW_LEFT = "split [pane] %s" % LEFT
+SPLIT_MOVE_UP = "pane %s [<n>]" % UP
+SPLIT_MOVE_DOWN = "pane %s [<n>]" % DOWN
+SPLIT_MOVE_RIGHT = "(pane %s|%s pane) [<n>]" % (RIGHT, RIGHT)
+SPLIT_MOVE_LEFT = "(pane %s|%s pane) [<n>]" % (LEFT, LEFT)
```
kendonB: @lexxish should these be RIGHT, FORWARD and LEFT, BACK?

kendonB: no I get it - these are "pane left" and "left pane"

Issue #541, opened 2019-05-25
HTML elements not written properly in Jetbrains IDE


gntsketches:
Hi, in Jetbrains Webstorm, html elements often are written with additional partial tags. For example, when I say "division", I get this `<div></div>div></div>` (where I just expected `<div></div>`) ... note the second opening div is missing the `<`. The curser is placed after the "i" in the first closing </div>. "Paragraph" will be written as `<p>/n</p>p>` with the cursor at the newline. This behavior is inconsistent depending on the surrounding context. 

By contrast, things seem to work fine in Notepad++

LexiconCode:
This is most likely an issue related to auto complete being triggered in Webstorm.

LexiconCode:
is this true of all commands in Webstorm?

LexiconCode:
https://github.com/dictation-toolbox/Caster/issues/56
@kendonB looks like that is still relevant, Thoughts?

ContextAction would work well for exceptions but this is pretty systemic within the HTML grammar for webstorm.

Maybe the grammar could be Re-factored so it does not trigger auto complete in webstorm.

LexiconCode:
I confirm that this is definitely related to auto complete. Webstorm not only automatically closes HTML tags but moves the cursor as well producing the issue.

 - The following works as expected
`"division": R(Text("<division>"), rdescript="HTML: Division"),`
Outputs
`<division></division>`

- Whereas the original
"division":  R(Text("<div></div>") + Key("left/10:6"), rdescript="HTML: Division"),
Outputs
<div></div>div></div>

gntsketches:
Oh, of course.

I definitely use the autocomplete functionality alot when coding, and continuingly/more it would seem with voice. If it's possible to not autocomplete for HTML code in particular that might be helpful. But disabling it globally wouldn't be good.

LexiconCode:
We deftly don't want to disable auto complete. Preferably I wouldn't want anyone to have to touch their settings to make this work. There are a few suggestions in the above referenced issue that need to be investigated as an alternative.

gntsketches:
Would it be possible to prefix/suffix the command with "open" or "close" and just get part of the tag? That way the autocomplete functionality wouldn't conflict. 

It would also be useful in those times when stray element tags need to be dealt with. This could also be useful for (), [], {}... it would be cool to be able to say "brax closed" in similar situation. 

Apologies if something like this case already exists, if so I may have missed it since I'm new.

kendonB:
[...]

PR #543, opened 2019-05-25
HTML workaround for Jetbrains IDE auto complete #541

LexiconCode:
This will request contextually changes imports `Text `to `Paste` based on AppContext for all supported Jetbrains products.
 `from dragonfly.actions.action_paste import Paste as Text`

### DIFF
--- a/castervoice/lib/ccr/html/html.py
+++ b/castervoice/lib/ccr/html/html.py
@@ -8,12 +8,8 @@ from castervoice.lib import settings
 
 # HTML auto complete workaround for Jetbrains Products
 # If more applications have issues with auto complete may be moved to actions.py
-context = AppContext(executable="idea", title="IntelliJ") \
-    | AppContext(executable="idea64", title="IntelliJ") \
-    | AppContext(executable="studio64") \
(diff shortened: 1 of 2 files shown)


castervoice/lib/ccr/html/html.py
```diff
@@ -1,8 +1,29 @@
...
+from castervoice.lib import settings
+
+# HTML auto complete workaround for Jetbrains Products
+# If more applications have issues with auto complete may be moved to actions.py
+context = AppContext(executable="idea", title="IntelliJ") \
+    | AppContext(executable="idea64", title="IntelliJ") \
+    | AppContext(executable="studio64") \
+    | AppContext(executable="pycharm")  \
+    | AppContext(executable="webstorm64") \
+    | AppContext(executable="webstorm")
```
kendonB: Can pass a list to `AppContext` as in something like `AppContext(executable=["studio64", "pycharm"]) ` etc

LexiconCode: Thanks that would be much cleaner approach.

PR #550, opened 2019-05-29
Make bow optional

kendonB:
@lexxish and I suggested this and @LexiconCode approved. Everything still seems to work.


kendonB:
CasterQuickReference.tex is also updated.

PR #555, opened 2019-06-02
Add short integer

kendonB:
This pull request re-implements the `IntegerRefST` class to use `ShortIntegerRef` by default and adds a settings option to disable it.

@mrob95 Are you able to let me know if this is a reasonable approach to incorporate ShortIntegerRef and tidy up that part of the codebase?


mrob95:
Looks like it should work. This is also probably more elegant than manually changing `IntegerRefST` across the code base.

PR #557, opened 2019-06-03
move all links from synkarius to dictation-toolbox

kendonB:


kendonB:
Also removes a link to the wiki which is no longer there

kendonB:
I opened this PR using:
```
hub pull-request -o -b develop -a kendonB
````
Worked great!

PR #418, opened 2019-03-17
add new commands to Dragon module

alexboche:
I previously had a pull request for adding new command in Dragon but I think it was messed up in certain ways. I think we should just start over with this pull request and close the other one. Now that I'm getting the hang of how to use git properly I should be getting it right the 1st time more often

PR #559, opened 2019-06-03
add new commands to dragon.py

alexboche:
I am redoing previous pull request #418 because I'm having some issues with git.
I have tried to respond to the comments there.
I took out the proposed commands for ignoring what you say if you say a specific word, the reason being that I don't know what word to use since people may say that word in sentences and thus have their dictation ignored accidentally.


kendonB:
new contents are just added below the old contents. See "Files changed".

alexboche:
Sorry about that. Should be fixed now

castervoice/apps/dragon.py
```diff
@@ -36,11 +63,12 @@ class DragonRule(MergeRule):
         'normal mode':
             R(Playback([(["normal", "mode", "on"], 0.0)]),
               rdescript="Dragon: Normal Mode"),
-        'command mode':
+        'com on':
```
kendonB: keep the old one as an option?

castervoice/apps/dragon.py
```diff
@@ -36,11 +63,12 @@ class DragonRule(MergeRule):
...
             R(Playback([(["command", "mode", "on"], 0.0)]),
-              rdescript="Dragon: Command Mode"),
-        'scratch':
-            R(Playback([(["scratch", "that"], 0.0)]), rdescript="Dragon: 'Scratch That'"),
+              rdescript="Dragon: Command Mode (On)"),
+        'com off':
```
kendonB: Good to have a verbose version of this too.

castervoice/apps/dragon.py
```diff
@@ -55,24 +83,77 @@ class DragonRule(MergeRule):
...
+        "cory <text>": 
+            R(Mimic("correct", extra="text") + WaitWindow(title="spelling window") + Mimic("choose", "one"),
+                rdescript="Dragon: brings up the correction menu for the phrase spoken in the command and chooses the 1st choice"),
+        "cory that": 
+            R(Mimic("correct", "that") + WaitWindow(title="spelling window") + Mimic("choose", "one"), 
+                rdescript="Dragon: brings up the correction menu for the previously spoken phrase and chooses the first choice"),
+
+        "make that <text>": R(Mimic("scratch", "that") + Mimic(extra="text"), 
+             rdescript="Dragon: deletes the dictation generated by the previous utterance and replaces it with what you say next"),
+        'scratch [<n10>]': R(Playback([(["scratch", "that"], 0.03)]),
```
kendonB: double quotes for consistency

kendonB:
Still needs testing (by me) but looks pretty good so far. We will add docs right before merging as well.

alexboche:
So this pull request is done and tested by me. (maintainers may want to tested). I would remove the "waiting for user reply" tag unless I'm missing something.

PR #567, opened 2019-06-05
add most essential click by voice commands

alexboche:
I am leaving out some of the other commands we had before that lead to misfires.
I have also changed it so you have to say the option e.g. 'tab' before the number rather than vice versa.
I am also requiring a specification of the options rather than making it optional and doing 'click' by default. Users may want to change that although I find that often the number by itself is not recognized for some reason.


kendonB:
This one has the numbers in the file path again

LexiconCode:
Due to that this requires an extension it would be best not to include it. However there is no current method to add application-specific rules. Some thought will have to be put on how to handle that.

LexiconCode:
closing to tidy up.

Issue #554, opened 2019-06-02
Update some engine-specific code


Danesprite:
Some of Caster's code for WSR and natlink is no longer necessary with recent changes to dragonfly. Changing a few things should make it easier to test with the "text" engine (#490) and use Caster with other engines.

- [x] Update *_caster.py* to use `engine.recognize_forever()` instead of the pump waiting messages loop if WSR is being used. The `recognize_forever()` engine method pumps waiting messages, calls timer functions set via `create_timer()` and updates grammar state on window changes (to fix an annoying bug).
- [ ] Use the normal `RecognitionHistory` class from dragonfly instead of the class in Caster's *wsrdf.py* file.
- [x] Use the `engine.create_timer()` method for all engines instead of dragonfly's `_Timer` class for natlink and the `TimerForWSR` class for WSR.

@LexiconCode Let me know what you think. It might take me a while to get to these and there may be other problems. Feel free to edit the post.

LexiconCode:
Sounds good, thank you for opening an issue. With my clinical rotation I just don't have the time right now.

Danesprite:
No problem. That's okay, I know you're very busy. I can probably submit a PR to fix the first two soon.

LexiconCode:
Update `_caster.py` to use `engine.recognize_forever()` completed by https://github.com/dictation-toolbox/Caster/commit/0c8dddfed00f857d4159c41134ff026ab9fe8830

Danesprite:
Thanks. The loop isn't necessary for that method. I'm about to fix the only remaining issue in a new PR, so I'll just remove the while loop in there.

Danesprite:
The `_wait_for_wsr_activation()` function only really applies to `sapi5shared`, not `sapi5inproc`. Should it be removed?

LexiconCode:
I would say remove it.

PR #577, opened 2019-06-12
Use Dragonfly's new timers and update Sikuli code

comodoro:
In order to remove the ugly deprecation warning. @Danesprite could you please look at the usage in `stackitems.py` at the bottom and confirm that this is the intended usage?

Sikuli code was made into a class to avoid globals. By the way "terminate sick server" without having Sikuli commands raises an exception for me on develop, which should be solved here.

Looking at Dragonfly git log, this change was not released yet, so this should wait until then and dependencies will have to be bumped.


Danesprite:
@comodoro The usage looks good to me. I should be able to release *dragonfly2* version 0.15.0 today or tomorrow. I think the `TimerForWSR` class can be deleted in this PR too.

This is related to issue #554.

comodoro:
@Danesprite Thank you. I will remove the old timer class, thanks. As for the new release, whenever you want, I don't think this is pressing.

Danesprite:
No problem. I agree this isn't pressing, there are just enough unreleased changes now that there should be a new release version.

LexiconCode:
is this considered complete?

Danesprite:
@LexiconCode Although I should be releasing dragonfly2 version 0.15.0 soon (took longer than expected), this PR doesn't really require it. @comodoro mentioned an error with the "terminate sick server" command. Perhaps that could also be fixed here?

This PR makes significant changes to the Sikuli code, so I'll edit the name.

comodoro:
@Danesprite @LexiconCode Can be merged then, the sikuli problem was just an exception  on exiting, I don't think being a problem in itself, and does not happen ( was resolved ) in this branch/PR.I wanted to increase the version in requirements as well, but it can be done anytime after.

Danesprite:
@comodoro Ah, okay then :+1:

LexiconCode:
Why freeze the requirements '=='?

setup.py
```diff
@@ -59,12 +59,12 @@ def run(self):
         "Operating System :: OS Independent"
     ],
     install_requires=[
-        "dragonfly2>=0.11.1",
-        "wxpython>=4.0.4",
-        "pillow>=5.3.0",
-        "toml>=0.10.0",
+        "dragonfly2>=0.14.2",
```
LexiconCode: Shouldn't this be 0.15.0?

comodoro:
This is after reading some discussions, so that setup.py does contain only restrictions on packages that do not work and we know that and requirements.txt contains packages we know do work. I only know there was something in Dragonfly fairly recently that Caster uses. Feel free to adjust this.

LexiconCode:
> 
> 
> This is after reading some discussions, so that setup.py does contain only restrictions on packages that do not work and we know that and requirements.txt contains packages we know do work. I only know there was something in Dragonfly fairly recently that Caster uses. Feel free to adjust this.

That makes a lot of sense.

LexiconCode:
I'm running into a few issues and trying to tease out whether it's related to this pull request or upstream in Sikulix

LexiconCode:
Once conflicts are resolved like get this merged.

comodoro:
I'm on it. I admit I am not convinced about the mass removal of rdescripts, even where spec is not clear about what the command does. As for star imports, I'm not a fan, but there are positive aspects.

LexiconCode:
Thanks for your help with this! 

It's good to have feedback on these matters so thank you. 

1. Regarding removal of rdescripts. Overall it simplifies maintaining code. As we go along and if you have any suggestions we will be adding back some to clarify commands that are ambiguous. Let me know if there's anything in particular or feel free to make pull requests.

2. As for the * or [standard imports](https://caster.readthedocs.io/en/latest/readthedocs/Examples/Rule_Construction/#standard-imports). It has its advantage in making easier to add new grammars but it does take away readability. Fortunately if this doesn't pan out it's a relatively minor change. Let's run with it for a bit and see how it goes overall.

mrob95:
Regarding star imports, I realise that they are frowned upon in Python and I understand why. In our case though, where we are basically importing the same things over and over again, and where easy experimentation is desirable, I think it's the right choice.

Another thing to consider is that this makes it a lot easier to change the way objects are loaded. For example when Aenea support was added, dragonfly imports had to be manually changed in ~50 files, whereas now the same change would just need to be done in one place.

PR #582, opened 2019-06-16
Revert clipboard to use json

### LINKED ISSUE #357: Bugs in handling non-ASCII characters in clipboard over restarts
1. If I select "závorka", do "stoosh three", restart Dragon and do "spark three", I get `zxe1vorka`. It should be encoded as `z\xe1vorka`. Probably affects characters 128-255. Confirmed by @mrob95 

2. Certain saved clipboard items give
```
<class 'toml.decoder.TomlDecodeError'>
Reserved escape sequence used (line 1 column 1 char 0)
<traceback object at 0x0EF9EB48>
  File "C:\workspaces\dns_userdir\caster\lib\utilities.py", line 85, in load_toml_file
    result = toml.loads(f.read())

  File "C:\Python27\lib\site-packages\toml\decoder.py", line 458, in loads
    raise TomlDecodeError(str(err), original, pos)
```

A clipboard file is pasted here: https://pastebin.com/jGeV5jvE

comodoro:
Finally resolve #357 the least inventive way. Reasons in the issue, unable to fix toml. By the way if more exotic languages where Unicode is needed for dictation are supported, this will become an issue again with aliases, macros etc.

### DIFF
--- a/castervoice/lib/navigation.py
+++ b/castervoice/lib/navigation.py
@@ -49,7 +49,7 @@ def get_direction_choice(name):
 
 def initialize_clipboard(nexus):
     if len(nexus.clip) == 0:
-        nexus.clip = utilities.load_toml_file(
+        nexus.clip = utilities.load_json_file(
             settings.SETTINGS["paths"]["SAVED_CLIPBOARD_PATH"])
 
 
@@ -100,7 +100,7 @@ def _text_to_clipboard(keystroke, nnavi500, nexus):
                 # time for keypress to execute
                 time.sleep(settings.SETTINGS["miscellaneous"]["keypress_wait"]/1000.)
                 nexus.clip[key] = unicode(Clipboard.get_system_text())
-                utilities.save_toml_file(
+                utilities.save_json_file(
                     nexus.clip, settings.SETTINGS["paths"]["SAVED_CLIPBOARD_PATH"])
             except Exception:
                 failure = True
@@ -155,7 +155,7 @@ def duple_keep_clipboard(nnavi50):
 
 def erase_multi_clipboard(nexus):
     nexus.clip = {}
-    utilities.save_toml_file(nexus.clip,
+    utilities.save_json_file(nexus.clip,
                              settings.SETTINGS["paths"]["SAVED_CLIPBOARD_PATH"])
 
 
--- a/castervoice/lib/settings.py
+++ b/castervoice/lib/settings.py
@@ -195,7 +195,6 @@ _DEFAULT_SETTINGS = {
         "SIKULI_SCRIPTS_PATH": _USER_DIR + "/sikuli",
         "GIT_REPO_LOCAL_REMOTE_PATH": _USER_DIR + "/data/git_repo_local_to_remote_match.toml",
         "GIT_REPO_LOCAL_REMOTE_DEFAULT_PATH": BASE_PATH + "/bin/share/git_repo_local_to_remote_match.toml.defaults",
-        "BRINGME_DEFAULTS_PATH": BASE_PATH + "/bin/share/bringme.toml.defaults",
         
         # REMOTE_DEBUGGER_PATH is the folder in which pydevd.py can be found
         "REMOTE_DEBUGGER_PATH": "",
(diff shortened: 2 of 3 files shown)


kendonB:
#485 still works after this change, FYI

kendonB:
Saw this error just now:
```
<type 'exceptions.ValueError'>
Extra data: line 1 column 4 - line 4 column 1 (char 3 - 56)
<traceback object at 0x12E19FA8>
  File "C:\Users\kmbel\Documents\caster\castervoice\lib\utilities.py", line 137, in load_json_file
    result = json.load(json_file)

  File "C:\Python27\Lib\json\__init__.py", line 291, in load
    **kw)

  File "C:\Python27\Lib\json\__init__.py", line 339, in loads
    return _default_decoder.decode(s)

  File "C:\Python27\Lib\json\decoder.py", line 367, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
```

comodoro:
@kendonB Could you please specify more closely when does the error occur and contents of `clipboard.json`?

kendonB:
I see it upon startup and I don't see `clipboard.json` in there at all. Only toml. Do you need to send it to `.caster` if it's not already there?

comodoro:
Oh, (facepalm), I forgot that everyone will still have clipboard.toml in their settings.toml. If you change `SAVED_CLIPBOARD_PATH` to clipboard.json, it should work.

kendonB:
Shouldn't the PR try to do that itself?

comodoro:
It does with settings.py, however it cannot touch your already generated user settings. It would be possible to create new constant in settings.py for this, but the old one would still hang in users' files, so I did not (it is the same as with moving to toml before)

comodoro:
Although after some thought I can see its merit in reducing users' disturbance. What do you think?

kendonB:
I think it's fine to change the path in `settings.toml` via `settings.py` if it's `USER_DIR + "data/clipboard.toml"`

comodoro:
@kendonB Do you mean like adding this specific check (probably in the `_init` function)? It looks a bit suspicious, but with clear indication that it is temporary I have no problem with that.

kendonB:
I was thinking it would be in `settings.py`? Because that already writes new settings elements into `settings.toml` and this would just instead write over an existing settings element.

comodoro:
@kendonB Sorry for the delay. Is the migrate commit what you had in mind? It is ugly and duplicates I/O (because of impirting order; utilities have to be refactored), but I am hoping we will be able to remove it before merging into master.

kendonB:
shouldn't it also remove the `clipboard.toml`?

kendonB:
I am no longer seeing the error though

comodoro:
@kendonB Sorry, yes, thanks, added. Note though that if you already have the path changed from trying this PR, it will not be touched.

kendonB:
I updated this PR locally, changed my clipboard path back to the toml in settings, and now see the old clipboard file disappear. Great! I'm going to merge this.

LexiconCode:
This error keeps popping back up. Looks like it's trying to read clipboard.json as Toml file.
@comodoro 
```
<class 'toml.decoder.TomlDecodeError'>
Found invalid character in key name: '}'. Try quoting the key name. (line 1 column 2 char 1)
<traceback object at 0x1C43C170>
  File "D:\Backup\Library\Documents\Caster\castervoice\lib\utilities.py", line 113, in load_toml_file
    result = toml.loads(f.read())

  File "C:\Python27\lib\site-packages\toml\decoder.py", line 227, in loads
    original, i)


```

comodoro:
@LexiconCode Can happen now if developers switch branches with and without this (and the opposite, if you remake toml clipboard on an old branch). I think for users it should be safe.

LexiconCode:
That makes total sense.

LexiconCode:
@comodoro should we keep using json for clipboard?
Then remove the following temporary code for migrating https://github.com/dictation-toolbox/Caster/blob/5aecbf0ff7c292bf31af7fe9f21390db588f2fcd/castervoice/lib/settings.py#L440

PR #584, opened 2019-06-18
add new "tell" commands

alexboche:
Add new commands for going to the end of the, line putting some punctuation, and then pressing enter

### DIFF
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -270,7 +270,7 @@ class Navigation(MergeRule):
             R(Function(navigation.left_up, nexus=_NEXUS)),
 
     }
-    tell_commands_dict = {"dock": ";", "doc": ";", "sink": ""}
+    tell_commands_dict = {"dock": ";", "doc": ";", "sink": "", "com": ",", "deck": ":"}
     tell_commands_dict.update(text_punc_dict)
 
     extras = [
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -298,7 +298,9 @@ class Navigation(MergeRule):
         Choice("semi", {
             "dock": ";",
             "doc": ";",
-            "sink": ""
+            "sink": "",
+            "(com | comma)": ",",
+            "(deck | deckle | cole)": ":",
         }),
         Choice("word_limit", {
             "single": 1,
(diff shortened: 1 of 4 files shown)


LexiconCode:
Patch 15?

castervoice/lib/ccr/core/nav.py
```diff
@@ -298,7 +298,9 @@ class Navigation(MergeRule):
         Choice("semi", {
             "dock": ";",
             "doc": ";",
-            "sink": ""
+            "sink": "",
+            "(com | comma)": ",",
+            "(deck | deckle | cole)": ":",
```
kendonB: Can you pull in and invert the punctuation dictionary here? I should be able to change the punctuation dictionary and have it work here. to invert a dict do something like: 

```
inv_tpb = {v: k for k, v in text_punc_dict.iteritems()}
```

castervoice/lib/navigation.py
```diff
@@ -243,9 +243,9 @@ def curse(direction, direction2, nnavi500, dokick):
 def next_line(semi):
     semi = str(semi)
     Key("escape").execute()
-    time.sleep(0.25)
+    time.sleep(0.05)
     Key("end").execute()
-    time.sleep(0.25)
+    time.sleep(0.05)
```
kendonB: This is probably fine.

castervoice/lib/ccr/core/nav.py
alexboche: Why would I need to invert the dictionary? Isn't it already in the proper orientation for this purpose?

alexboche:
I tried rebasing which I have never done before. I think I mess it up a bit so it picked up some of the maintainers commits.

castervoice/lib/ccr/core/nav.py
kendonB: You're adding single syllable versions here, which is fine because they're prefixed with "tell" so ambiguity is unlikely. However, the elements "deckle", "comma" etc are in the punc dictionary and should be referenced from there. To access "deckle", we'd need to do `inv_tdb[":"]` so the new command would be `"(deck | cole | " + inv_tdb[":"] + ")"`

kendonB:
> I tried rebasing which I have never done before. I think I mess it up a bit so it picked up some of the maintainers commits. I think it should be okay,but let me know if I need to do it over.

as long as the diff is clean, we can do a squash merge to get rid of any weirdness.

castervoice/lib/ccr/core/nav.py
kendonB: I now see what you did in your latest update. Yes that approach works nicely.

kendonB:
Can you resubmit this PR clean? The rebase is messing with the diff and showing changes that have already been made in develop. I don't know why.

alexboche:
Because I made some mistakes with git when making this pull request, I am closing this pull request and have made a new one in its place:  #593

PR #585, opened 2019-06-19
Modifier keys

alexboche:
I added functionality for `<modifier_key> <button>` for all the modifier keys and nearly all possible buttons. the modifiers can be spoken as the regular English name or as the caster name (e.g. "control "or "fly"). To press a key without modifiers, say "hit <button>". The original caster master navigation command `# "(<mtn_dir> | <mtn_mode> [<mtn_dir>]) [(<nnavi500> | <extreme>)]":` causes some ambiguity with these new commands so I commented it out. All the functionality is preserved except for the ability to say "fly", "shin", "que" _without_ specifying "lease" or "ross" (if you specify the direction those still work fine). I added six new one syllable commands for that purpose (one for each in each direction) (bird, fird, brick, frick, blitch, flitch). I also added "latch", "ratch" which can be used instead of "ross/lease wally" in a stand-alone use or also in cases like "shin ross/lease wally" (you can still say ross/lease wally if you want).

Currently the keys must all be prefaced by a modifier; this could be changed.

Right now I made them all these CCR (and allow some of them to be repeatable to varying degrees). Because the modifier commands have to be prefaced by the modifier I think the risk of misfires is lower than usual. It seems that grammar complexity is a limited resource that CCR commands use up. Given that some of these commands like "control alt insert" may never be used, I think we should not use that resource on them. Thus I think some of these should be made CCR before merging. I have not done so yet because I don't know yet which ones to make CCR and it is simpler to have it this way while I am still getting feedback from the maintainers and the community. In particular, I don't have a good understanding of the grammar complexity error.

To make it easier for people to give feedback on this, I'm including the full key dictionaries and new command names in this comment. The number after button_dictionary_ indicates the number of times the keys there which must be prefaced by modifiers can be repeated:
```
button_dictionary_500 = {"(tab | tabby)": "tab", "(backspace | clear)": "backspace", "(delete|deli)": "del", "(escape | cancel)": "escape", "(enter | shock)": "enter",
    "(left | lease)": "left", "(right | ross)": "right", "(up | sauce)": "up",
    "(down | dunce)": "down", "page (down | dunce)": "pgdown", "page (up | sauce)": "pgup", "space": "space"}
    button_dictionary_10 = {"function {}".format(i):"f{}".format(i) for i in range(1, 10)}
    button_dictionary_10.update(caster_alphabet)
    button_dictionary_10.update(text_punc_dict)
    button_dictionary_1 = {"(home | lease wally | latch)": "home", "(end | ross wally | ratch)": "end", "insert": "insert", "zero": "0",
    "one": "1", "two": "2", "three": "3", "four": "4", "five": "5", "six":"6", "seven": "7", "eight": "8", "nine": "9"}
```
Note: the F keys are called "function <number>".
new command names (old names are preserved, see above):
```
"(lease wally | latch) [<nnavi10>]": R(Key("home:%(nnavi10)s")),
        "(ross wally | ratch) [<nnavi10>]": R(Key("end:%(nnavi10)s")),
        "bird [<nnavi500>]": R(Key("c-left:%(nnavi500)s")),
        "fird [<nnavi500>]": R(Key("c-right:%(nnavi500)s")),
        "brick [<nnavi500>]": R(Key("s-left:%(nnavi500)s")),
        "frick [<nnavi500>]": R(Key("s-right:%(nnavi500)s")),
        "blitch [<nnavi500>]": R(Key("cs-left:%(nnavi500)s")),
        "flitch [<nnavi500>]": R(Key("cs-right:%(nnavi500)s")),
```


castervoice/lib/ccr/core/nav.py
```diff
@@ -268,15 +270,68 @@ class Navigation(MergeRule):
...
+        Choice("button_dictionary_10", button_dictionary_10), 
+        Choice("button_dictionary_500", button_dictionary_500), 
+        Choice("modifier", {
+            "(control | fly)": "c",
+            "(shift | shin)": "s",
+            "alt": "a",
+            "(control shift | que)": "cs",
+            "control alt": "ca",
+            "(shift alt | alt shift)": "sa",
+            "(control alt shift | control shift alt)": "csa", # control must go first
```
kendonB: allow composition with fly, shin, alt as well. ie "fly alt shin" should be an option

kendonB:
right from when I started using caster I added `"hum": "home"`, `"end": "end"`, `"sun": "pageup"`, `"doon": "pagedown"` as options for direction. this change has only been helpful for me and I seldom get misrecognitions because of it.

What do people think about making this change here as well? Of course we can still keep Wally as an option but it gives some nice single syllable options for these powerful navigation commands.

@LexiconCode, @mrob95 input on this please

castervoice/lib/ccr/core/nav.py
```diff
@@ -268,15 +270,68 @@ class Navigation(MergeRule):
...
+        
+        Choice("button_dictionary_1", button_dictionary_1),
+        Choice("button_dictionary_10", button_dictionary_10), 
+        Choice("button_dictionary_500", button_dictionary_500), 
+        Choice("modifier", {
+            "(control | fly)": "c",
+            "(shift | shin)": "s",
+            "alt": "a",
+            "(control shift | que)": "cs",
+            "control alt": "ca",
```
kendonB: control alt | cult

castervoice/lib/ccr/core/nav.py
```diff
@@ -268,15 +270,68 @@ class Navigation(MergeRule):
...
+            "(shift alt | alt shift)": "sa",
```
kendonB: | salt

castervoice/lib/ccr/core/nav.py
kendonB: Add "cass" as an option?

kendonB:
I think this PR is going to help a lot for those of us who switch between typing and voice control!

castervoice/lib/ccr/core/nav.py
alexboche: The shortened names for the rarer modifiers

kendonB: Don't feel strongly about it either way. People can add them

kendonB:
@alexboche are you able to write out all the functional changes in this PR in your first comment? I was about to solicit comments on the Caster channel but realised that people would have to dig into the code to work out the changes.

alexboche:
> @alexboche178 are you able to write out all the functional changes in this PR in your first comment? I was about to solicit comments on the Caster channel but realised that people would have to dig into the code to work out the changes.

I have done so. Let me know if further elaboration is necessary.

kendonB:
> > @alexboche178 are you able to write out all the functional changes in this PR in your first comment? I was about to solicit comments on the Caster channel but realised that people would have to dig into the code to work out the changes.
> 
> I have done so. Let me know if further elaboration is necessary.

flitch etc as well

LexiconCode:
This can be tested and merged once documentation and conflicts are resolved.

alexboche:
I tried to do the documentation in the caster quick reference. Even without making any changes, when I try to compile the latex, I get the error message pointing to line 436. "! Missing $ inserted." On that line,the code is:
```
\end{minipage}
}
```
Any suggestions?

LexiconCode:
> 
> 
> I tried to do the documentation in the caster quick reference. Even without making any changes, when I try to compile the latex, I get the error message pointing to line 436. "! Missing $ inserted." On that line,the code is:
> 
> ```
> \end{minipage}
> }
> ```
> 
> Any suggestions?

A bit about my set up for latex. I found Miktex and Texmaker to be plug-and-play which none of the other editors or latex compilers seem to provide.
[Miktex](https://miktex.org/)
[Texmaker](https://www.xm1math.net/texmaker/)

In terms of troubleshooting does your error happen with the vanilla `CasterQuickReference.tex`? I verified that CasterQuickReference compiles from the Development branch.

alexboche:
Not sure what changed,it's working now. As I add things, things get cut off at the end of the mini page. Is there a way to get everything to line up nicely instead of having to manually adjust the mini pages all the way through the document every time I add something?

LexiconCode:
I'm still learning a bit myself with latex. Therefore I think the easiest solution would be to upload your working file I will take a look.

alexboche:
[Here](https://paste.gg/p/anonymous/6ecf9448dcf54d928570c94d10d0c2db) is what I have. The issue seems that if I move one section to a different many page that messes up the next section and so on and I end up having to shift everything over by a page which is a huge pain, there must be a better way.
Am having some issues with git, which is not allowing me to push the changes,which is why I'm using a paste service for now until I get that figured out. The stuff I'm adding is on line 232-256.  thanks

LexiconCode:
Thanks. Regarding git you might have to do `force push` to your current branch `modifier_keys`. Force push is okay when you don't have anyone else working on the same branch.

alexboche:
If you click on the latest commit, which is called, "first draft add to quick reference",you should be able to see the changes to the quick reference. Thanks

LexiconCode:
@mrob95 I can't quite get figure out how to make the latex play nicely without running out of margins. Any tips?

alexboche:
I just added a new command for making keystroke commands. The command is "keystroke [<modifier>] <button>". So now saying "keystroke control shift delta" should type out `Key("cs-d")`. One could certainly argue this should go in the voice dev command files,but since this is so similar to the modifiier key commands and everything required is sitting right there in nav.py, I thought it would be simplest to put it there. When testing this new command, I was having some trouble getting it to work with some of the keys in `button_dictionary_500`. I think this is some conflict with Unimacro,but I'm not totally sure. can somebody test this new command to make sure it works. Once this is given the green light, I can add a line to the voice dev commands documentation.

 I also fixed the typo with the home key.

kendonB:
> One could certainly argue this should go in the voice dev command files

You are right! It should go there and not in `nav.py`.

alexboche:
Okay I moved the new command for creating keystroke commands into the voice dev command file and called it dev key. If you just say dev key, then it will move the cursor inside the key action,but if you specify the keys,the cursor will be left outside the key action.

LexiconCode:
@alexboche I'm fine merging this without the documentation and we can remove the edits fromCasterQuickReference.tex and CasterQuickReference.tex.

We can include the documentation in https://github.com/dictation-toolbox/Caster/issues/629 as that issue matures.

alexboche:
I undid the changes to the caster quick reference. 
By the way, some of the shorter command words like "latch" will have to be monitored to see if they cause misfires. I did have one misfire where "sauce latch" was interpreted as "slash". If we stick with these, it's key to stick the landing on the "ch" sound.

alexboche:
I just added the ability to say "hit <button>" for when you don't want to modify the button with a modifier key.  I also added the F keys which can be used by saying e.g. "function 6". by the way, git Saying something about " rewrite castervoice/lib/dll/tirg-dll.dll".

LexiconCode:
> I just added the ability to say "hit " for when you don't want to modify the button with a modifier key.

That's brilliant! For some reason github won't let me resolve the conflicts online so you'll need to update it on your end.

kendonB:
@LexiconCode if you want to resolve the conflicts you could try out my "git push back to pull request alex" function:

https://github.com/kendonB/Caster/blob/b8160f73b4eb1212346966c29abb1351a4f2efab/castervoice/apps/gitbash.py#L132-L135

Full workflow:
1. Have a clean caster working tree
2. say "checkout pull request" on this webpage
3. Do a `git merge master` and resolve conflicts
4. say "git push back to pull request alex" then add `modifier_keys` as the branch name

alexboche:
okay I merged with the latest caster and added the windows key commands which I forgot

Issue #531, opened 2019-05-20
Voice dev commands "commander" function buggy in Notepad++


kendonB:
## To reproduce
Say "commander text"

## Actual output
```
        "": R(Text(""), rdescript="MyGrammar: "
,        
```          

## Expected output
```
"|": 
    R(Key(""), rdescript="MyGrammar: ")
```

LexiconCode:
@lexboche

PR #586, opened 2019-06-19
Voice dev commands tweak

alexboche:
Remove unwanted comma.


alexboche:
Currently the command for the rdescript prints out `rdescript="MyGrammar: "` . Lexicon code said that the rdescripts are no longer as important. Should we change the command so that it just prints out `rdescript=""),` ?

kendonB:
I would just remove the first space but keep it. 
```
"dev descript":
    R(Text(' rdescript="MyGrammar: "') + Key("left"), rdescript="CasterDev: Add the rdescript")
```
you could also fix up the problem in #531

kendonB:
you might also want to add an optional "with r descript" to the command phrases for when users want to add rdescripts

kendonB:
lots of these functions also expect the editor to auto complete parentheses and curly braces. the default should be that they work without needing to use an editor's autocomplete

PR #587, opened 2019-06-19
make some variables non-optional in Python.py

alexboche:
These commands were causing some misfires so I made the variables required rather than optional

### DIFF
--- a/castervoice/lib/ccr/python/python.py
+++ b/castervoice/lib/ccr/python/python.py
@@ -120,10 +120,10 @@ class Python(MergeRule):
             # going into the command pallet (cs-p) and typing in "insert snippet"
             # then press enter and then you have choices of snippets show up in the drop-down list.
             # you can also make your own snippets.
-        "try [<exception>]":
+        "try <exception>":
             R(Text("try : ") + Pause("10") + Key("enter/2")
             + Text("except %(exception)s:") + Pause("10") + Key("enter/2")),
-        "try [<exception>] as":
+        "try <exception> as":
             R(Text("try :") + Pause("10") + Key("enter/2") + Text("except %(exception)s as :")
             + Pause("10") + Key("enter/2")),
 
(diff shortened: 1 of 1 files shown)


kendonB:
We should get agreement from someone who uses caster for python programming. @mrob95?

alexboche:
> We should get agreement from someone who uses caster for python programming. @mrob95?

The commands I'm changing are commands that I added pretty recently, so there's a good chance that no one even uses them yet.

LexiconCode:
Since these are commands you recently added go ahead and make the changes.

castervoice/lib/ccr/python/python.py
```diff
@@ -120,19 +120,19 @@ class Python(MergeRule):
             # going into the command pallet (cs-p) and typing in "insert snippet"
             # then press enter and then you have choices of snippets show up in the drop-down list.
             # you can also make your own snippets.
-        "try [<exception>]":
+        "try <exception>":
             R(Text("try : ") + Pause("10") + Key("enter/2")
```
kendonB: no space here after try. There should be two enters (why a 0.02 second pause??). The cursor should end up on the line below try

kendonB: jason above also needs fixing. Can you do that here?

castervoice/lib/ccr/python/python.py
```diff
@@ -120,19 +120,19 @@ class Python(MergeRule):
...
-        "try [<exception>] as":
+        "try <exception> as":
             R(Text("try :") + Pause("10") + Key("enter/2") + Text("except %(exception)s as :")
             + Pause("10") + Key("enter/2")),
 
         # class and class methods
         "subclass": R(Text("class ():") + Key("left:3")),
         "dunder": R(Text("____()") + Key("left:4")),
         "init": R(Text("__init__()") + Key("left")),
-        "meth [<binary_meth>]": R(Text("__%(binary_meth)s__(self, other):")),
```
kendonB: Add enter characters after these?

LexiconCode:

PR #590, opened 2019-06-22
Add imports.py to reduce boilerplate

mrob95:
Following on from #451 and my mission to reduce the amount of *stuff* required to create a rule. All rules should be able to load everything they need just with:
```
from castervoice.lib.imports import *
```

There may well be things that I have forgotten to add to the `imports.py` file, and these can be added as necessary.


kendonB:
Is this going to slow down loading substantially?

mrob95:
I would be very surprised, since Python caches imported modules in `sys.modules`, so it will only load each one once no matter how many times it is imported in the code base

mrob95:
Okay I think I've done what I wanted to do in this pull request, specifically:
* Add `imports.py`
* Replace import statements in all language rules and core where appropriate
* Remove rdescripts from language and core rules

I am planning to do the same for apps as well but that can wait until #596 is ready.

kendonB:
Looks ready to me but note that I haven't tested it (I'm away from my talkie computer).

PR #593, opened 2019-06-23
Add tell commands for all punctuation

alexboche:
This is just redoing pull request #584 which I'm closing because I messed it up in terms of the git stuff. The "tell" commands go to the end of the line, put a character of punctuation, and then go to the
next line. This pull request extends the command to work with every character in the punctuation dictionary `text_punc_dict` . It preserves all the original functionality for example "tell doc" which puts a semicolon at the end of the line and then goes to the next line and adds a couple other shortcuts like "tell deck" which is short for "tell deckle",which also works (since its in the dictionary `text_punc_dict`).

It also reduces the wait time in the function that executes the command. I don't see why there should be any way time; in my own version of the command has no wait time and has worked fine for months. I have left some wait time just in case. It seemed to me to be necessary to reduce the wait time at least somewhat because the slowness was significant.


kendonB:
tested and these work. I'm guessing the pauses are important for some editors but let's see if anyone complains. since it's using the whole punc dictionary it would probably be tidier to have this in the punctuation.py file so if you wanted to do that you could.

You can go ahead and make the changes to casterquickreference.tex. You should just need to put an extra footnote in the punctuation section. It's a bit squished at the moment so you may need to move something around.

LexiconCode:
As soon as is has documentation we can merge.

Issue #589, opened 2019-06-21
Proposal for reducing boilerplate, folder structure and loading rules.


mrob95:
I haven't been particularly active recently, but have been bouncing ideas around my head for a while regarding the direction and structure of caster. Going to do a brain dump here in the hope that it prompts some discussion and hopefully consensus.

# Where we are
I'm going to refer in this section to *old caster* and *new caster*. 

Old caster is anything before the move to pip, and is basically a known entity. You install it by manually downloading the git repo and then manually setting the natlink user directory, and all grammars are stored and loaded from that directory. It has the advantage of being reasonably conceptually simple (everything in one directory), but is a little bit cumbersome to set up and does not make a clear distinction between user space and backend.

New caster is anything after the move to pip. The first thing to say about this is that however new caster was supposed to look, it is unfinished and has a few issues. Firstly, if installed using pip we have gone from caster being in one place (wherever the user downloaded it) to three places - `_caster.py` in the Natlink directory, most of the code in the Python packages directory, and the `.caster` user space containing settings and filters. I think this is really confusing. New users need to be able to easily find where the code is and be able to see how they can change it to do what they require. Secondly, because the instructions in the develop branch describe installation using pip, but we are not yet able to recommend this, develop and master are basically unmergeable and are beginning to diverge significantly in functionality.

At the moment I would say that we are stuck somewhere between old caster and new caster. So the question is: what do we have to do to new caster to get it to the point where it can replace old caster?

# Proposals
## Structure, Rule loading
- [x] Tidy up the folder structure. I think this is possibly the most important thing to do as it is a mess at the moment and it's hard to think clearly about changing things when some of the rules are in cv/apps all together, others are in cv/lib/ccr in their own subdirectories, etc. Moving everything into one cv/rules folder with a consistent structure will make it much easier to load core rules alongside user rules. Unfortunately this will probably break most of the active pull requests.

- [x] #616  Simplify and expand loading of rules from user space. There are a couple of pieces of this I think:
	* Simplify the process of creating user rules in user space so that they have exactly the same syntax as any other rule, removing the `get_rule` formulation.
	* Allow rules in user space with the same name to override caster defaults. So for example if you copy `python.py` into user space the Python rule will be loaded from that file instead of the default. You can then make changes to that file while still allowing the repository to be updated through pip or git.
	* Simplify this process further by adding voice commands to copy the file and open in a text editor, e.g. `edit Python grammar`.

[...]

PR #596, opened 2019-06-23
Rule loading functions in control.py

mrob95:
As mentioned in #589  I have implemented a series of functions in `control.py` to simplify and standardise the API for adding rules.

The only question I have at the moment is about lines like:
```
if settings.SETTINGS["apps"]["firefox"]:
```

Which only load app rules if the relevant value is set to true in the settings. Is there a good reason why this is necessary? Does anybody need the ability to disable particular app rules (perhaps RDP users)? @LexiconCode

### DIFF
--- a/castervoice/lib/control.py
+++ b/castervoice/lib/control.py
@@ -18,10 +18,10 @@ Commands for easily loading different types of rules, e.g.:
 control.non_ccr_app_rule(FirefoxRule(), AppContext("firefox"))
 '''
 def non_ccr_app_rule(rule, context=None):
-    grammar = Grammar(context=context)
     if settings.SETTINGS["miscellaneous"]["rdp_mode"]:
         nexus().merger.add_global_rule(rule)
     else:
+        grammar = Grammar(str(rule), context=context)
         gfilter.run_on(rule)
         grammar.add_rule(rule)
         grammar.load()
--- a/castervoice/lib/control.py
+++ b/castervoice/lib/control.py
@@ -1,10 +1,36 @@
+from dragonfly import Grammar
+
 from castervoice.lib.ctrl.nexus import Nexus
+from castervoice.lib import settings
+from castervoice.lib.dfplus.merge import gfilter
 
 _NEXUS = None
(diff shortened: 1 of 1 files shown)


LexiconCode:
I use RDP quite a bit in remote work. I've not really use disabling apps. Other people I know on the past have done that as they've had a preferred grammar that the choose to use over casters implementation.

Perhaps we can automate writing to app name toml using control.py. Perhaps it can pull from the grammar name and populate the entries without user intervention.

seekM:
> Other people I know on the past have done that as they've had a preferred grammar that the choose to use over casters implementation.

When app rules are copied to user space in the future, the copies are used automatically, aren't they? Wouldn't this make this value irrelevant? (I haven't used RDP and can't speak for this)

LexiconCode:
> > Other people I know on the past have done that as they've had a preferred grammar that the choose to use over casters implementation.
> 
> When app rules are copied to user space in the future, the copies are used automatically, aren't they? Wouldn't this make this value irrelevant? (I haven't used RDP and can't speak for this)

I think that is true and shouldn't have any impact on RDP.

mrob95:
@seekM this is my instinct, I'm not sure I can think of a situation where somebody would really need to disable particular app grammar, that if they needed to they could just make a blank file in their user space with the same name, or if this got cumbersome then we can come up with a better way.

Having to add lines to the settings file for each app when the vast majority of users will just want to leave them all on all the time seems like a waste. Can we just drop it?

LexiconCode:
> Can we just drop it?

let's drop it

seekM:
> that if they needed to they could just make a blank file in their user space with the same name 

Maybe deleting them could also work? Don't know how copying rules to user folder is supposed to work exactly, but if everything is copied, only apps rules in the user folder could be considered for loading? Then simply having no file would work.

Personally I have some of them disabled even though I don't use custom ones for them just because it feels unclean to me to have them loaded if they will never be active. This probably isn't a good reason though, so I'm fine with dropping it to.

PR #598, opened 2019-06-24
add command named Dick for right-click

alexboche:
I'm assuming that the vertical bars don't mess up the latex I did not test it

### DIFF
--- a/castervoice/doc/CasterQuickReference.tex
+++ b/castervoice/doc/CasterQuickReference.tex
@@ -337,6 +337,7 @@ Capitalisation and spacing can be combined into a single command. They may be su
 
 \command{kick}{left click}
 \command{kick mid}{middle click}
+\command{kick double|double kick|dick}{double click}
 \command{psychic}{right click}
 \command{shift right click}{shift right click}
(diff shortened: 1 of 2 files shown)


kendonB:
What's the use case for making this single syllable?

alexboche:
typically selecting a word so something like "dick stoosh/deli/hug-quotes" . it's debatable whether it's worth a single syllable

kendonB:
garb is already one of those. Given we already have "kick double" I don't really think it warrants a change. Others may disagree and I will try it out for a couple days

LexiconCode:
One thing to note as well "Dick" is slang for male anatomy in America.  It just seems a bit awkward dictating.

kendonB:
> One thing to note as well "Dick" is slang for male anatomy in America. It just seems a bit awkward dictating.

It would even be regarded as a curse word in some circles.

LexiconCode:
Well besides the term being used are you finding functionality in a single syllable word?

LexiconCode:
For now we will closes as should be able to be changed in words.txt

Issue #568, opened 2019-06-05
Add code to check for required dragonfly version


kendonB:
As per the gitter chat.

kendonB:
There should probably be dependency checking near the top of _caster.py so that no Caster modules load unless the required version of a package is installed. E.g. accepting lists for AppContext titles and executables is new in dragonfly2 version 0.14.0.
This would make it easier for people to see what's going on.
Something like this:
import pkg_resources
pkg_resources.require("dragonfly2 >= 0.14.0")

LexiconCode:
All right @kendonB let's see if it works on bounty source. I think it will trigger when you close the issue.

kendonB:
I have tested your fix and it forced me to install the new version of dragonfly. Great! closing

PR #599, opened 2019-06-24
Minimum version checking for dragonfly.

LexiconCode:
Caster updates sometimes require a mandatory upgrade to the latest dragonfly. The purpose of this pull request is to add minimum version checking to dragonfly. Implements https://github.com/dictation-toolbox/Caster/issues/568

### DIFF
--- a/_caster.py
+++ b/_caster.py
@@ -8,6 +8,18 @@ import os
 import logging
 logging.basicConfig()
 
+
+def version_minimum():
+    try:
+        import pkg_resources
+        version = "0.15.0"  # Version needs to be manually updated Caster requires a certain version of Dragonfly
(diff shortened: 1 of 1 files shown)


LexiconCode:
A few thoughts when a person has time to review 
- does this code placed higher in _caster if there's an ambiguous error due to dragonfly? The concern is the function may not execute before a error is raised.

- Preferably middle be bundled in the DragonflyCheck class review. My brain which is limited with object-oriented programming thus far didn't come up with a clean way of doing that.

mrob95:
As you say, I think this needs to be much higher up in the file, otherwise new dragonfly stuff will cause errors when grammars et cetera are imported.

PR #601, opened 2019-06-25
add sequences of characters for the move commands only

alexboche:
So far I only did it for the move/go command (which is probably the most useful). would be easy to extend to the others.
I made it so that you can say (and thus have the command search for) sequences of characters up to length three (could be easily changed). 
The commands are CCR. To avoid ambiguity when following this command by character commands (e.g. "deckle" or "arch"), if the sequence of characters has length two or greater, you must the word "stop" after uttering the move/go command. 
Of note, this change uses dragonfly's `Repetition` which is not often used in caster.


kendonB:
Game changer!

"stop" gets misinterpreted as "dot" every time the second character is "ace" for me. I quite like the word "over" as someone suggested on the chat.

kendonB:
Yes over seems to work a lot better

LexiconCode:
if there's a strong preference here we can always change in words.txt

alexboche:
So I'll just do it for all the commands now?

LexiconCode:
Sure what you mean?

Whichever way you guys choose to go forward one of you can use the word.txt to make the change. Not every command name will serve everyone's personal preference or utility.

alexboche:
Having some trouble getting the sequences to work for the other commands. Also I need to make an adjustment so that the functions know whether there searching for characters or words. Simply using a different regular expression for short character sequences  is in good enough because it will get confused with some words like "and" if you allow sequences of characters up to length three.
I probably won't have time to finish all the stuff,but all post a little more information so others can take over if they want

alexboche:
Having some trouble getting the sequences to work for the other commands. You can see what I tried here https://paste.gg/p/anonymous/0d52f05bff8c478394017c5d0b774ebf 

 To extend longer sequences, one must do two things:

* First, one must increase the Repetition object's max parameter to 4 or higher like this:
`Repetition(character_choice_object, min=2, max=4, name="character_sequence"),`   My understanding is that the max number is not included (so the above would be sequences of length two and three).

* Second, one must adjust the regular expressions matching stuff.  
     *increase the parameter taking the value 2 in `character_list = make_list_of_permutations_up_to_length(character_list, 2)` 
     * passing in a parameter that depends on whether the character_sequence or dictation command is used and using this parameter in the function `get_start_end_position` so that it matches sequences of characters that are next to other characters but not words e.g. so it doesn't match land when searching for and; this already works in the original release of these commands but now is getting messed up by the addition of character sequences which include all permutations including character sequences like "to" "and".  Right now it will still accidentally match sub words of length two, but that's not so bad.


 I won't have time to do this stuff, but this pull request should work pretty well as it is right now, and if somebody wants to extend to longer character sequences the information on how to do so is above.

kendonB:
I have made a couple of changes to this pull request.

First, I have made the two character commands into one. The continue CCR/finisher word "over" is now optional so that "go lease before oscar prime" works as a single command. Anyone who wants to use this command with CCR with character sequences of length 1 and 2 must use the finisher word "over".

Second, I made a minor change to the code for going up and down at the initial selection stage. The second change is not functional and is just tidying up.

Can someone else test this and make sure that it works as intended. Please test the following:
```
this is a test sentence|
```
Say "go lease after arch ace" and you should end up here:
```
this is a |test sentence
```
Then say "go ross after char echo over dot" and you should end up with this:
```
this is a test sentence.
```

LexiconCode:
> Then say "go ross after char echo over dot" and you should end up with this:
> 
> ```
> this is a test sentence.
> ```
the last command produces `this is a .test sentence` The cursor `this is a |test sentence`

kendonB:
@alexboche was there a reason you reversed the `[<number_of_lines_to_search>]` and `[<before_after>]` in the first commit of this PR?

kendonB:
@alexboche can you review my documentation changes?

alexboche:
@kendonB documentation looks good to me thanks!

kendonB:
@alexboche I'm a bit confused. Why can't you just use a dummy variable depending on whether you're using the dictation call vs the character sequence call? Apologies for no links as I'm on my phone.

alexboche:
@kendonB Yes that should work,I'm embarrassed that I did not think of that.

alexboche:
Okay I just added the `dictation_versus_character` which takes value `"dictation"` if the target object is dictation and `"character"` if the target object is a character or character sequence. This is done for all the functions, though only the move function supports character sequences. Using this new variable, I updated the regular expression stuff so as to resolve any issues with things like "land" versus "and". 

Going forward, this will make it easier to implement different functionality for dictation on the one hand and versus characters (and character sequences) on the other. For example, increasing speed by using the control arrow keys instead of regular arrow keys for moving towards a word but just using the regular arrow keys when moving towards a character since that's more complicated. I don't think I'll have time for that control arrow key stuff , but this pull request should be good to go. The new variable dictation_versus_character

kendonB:
I retested this with the latest changes and it looks good to go.

PR #602, opened 2019-06-25
new legion dll

kendonB:
recompiled this with the updated VS toolchain. It seems to work marginally faster now. Would be good to get some other "Legion" users to test though.


LexiconCode:
Fantastic, I have a 4K monitor. Legion takes a bit more time with higher resolutions.

LexiconCode:
I'm not really seeing the speed difference at least at a higher resolution but it works as expected. I suppose we could put timeit functions to empirically test.

kendonB:
@seekM does this change make legion work for you?

LexiconCode:
@kendonB Does it change which version of Visual C++ Redistributable packages is required? Historically that is been Visual Studio 2013 (x86).

It would be nice to bump the version [Microsoft Visual C++ Redistributable for Visual Studio 2015, 2017 and 2019. T](https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads) which I believe comes with Windows 10 nowadays.

kendonB:
I'm not sure of the answer to that. I can test it out in a few days by selectively deleting VS packages.

kendonB:
@LexiconCode I've deleted all of the VS Redistributable packages and legion still works regardless. It's possible that the underlying dependency is built into Windows 10 now. I will also check on my Windows 7 machine.

kendonB:
I can't get any of these to work on my Win 7 machine. I tried the old dll, the new one in this PR, and also built a new one on my Win 7 machine. None worked. Win 7 is not long for this world, so this is only a short-term problem.

LexiconCode:
> 
> 
> I can't get any of these to work on my Win 7 machine. I tried the old dll, the new one in this PR, and also built a new one on my Win 7 machine. None worked. Win 7 is not long for this world, so this is only a short-term problem.

Excellent I think we can live with that. It should be documented somewhere...

kendonB:
I have also confirmed that the old .dll works just fine on my Windows 10 computer. So installing the redistributable packages may help some people who aren't otherwise up-to-date but the current pull request doesn't seem to change the requirement (because there is no requirement any more). I've added some documentation to the installation file. 

Those who want to compile this for themselves can do:

- `git clone https://github.com/kendonB/tirg-dll`
- Build the project using Visual Studio Community 2019 (free download) choosing "Release" rather than "Debug" as the build setting. Ctrl-Shift-B is the build command.

PR #604, opened 2019-06-25
Major app rules clean up

mrob95:
~1k LoC down. Functionality should not be affected at all (except in areas where I have fixed bugs), but it will need a bit of testing to make sure everything is still working.

**Major changes**
* Replace import statements at the top of files to import everything from imports.py
* Replace grammar loading code at the bottom of files with calls to the functions in control.py
* Remove handwritten rdescripts as standardised versions of these can now be generated automatically

**Minor changes**
* Removed `cmd.py`, as it is just a subset of the functionality and context of `gitbash.py`.
* Removed a few unnecessary comments
* Fixed a few bugs with CCR apps like eclipse

My hope is that this makes the code much more readable and abstracts away stuff which the user doesn't really need to worry about (like the implementation of RDP mode), therefore allowing quicker iteration and easier creation of new grammars.

One question: I think that because commands are now a lot shorter, if we run yapf it will try to put them on one line, so instead of
```
"switch user":
	R(Key("cs-m")),
```
We will get
```
"switch user": R(Key("cs-m")),
```

Which format do people prefer?


LexiconCode:
As a personal preference I prefer one line. It makes it much easier for find and replace and general editing. The downside on smaller screens it might be an issue.  I'm willing to however I'm willing to be flexible on this.

mrob95:
Think I agree. Most commands should now fit comfortably on one line without wrapping, and it allows a lot more commands to fit on the screen at once.

mrob95:
Ah, I did that manually. I think it makes more sense visually if all the commands in a rule are either on one line or two.

LexiconCode:
That makes a lot of sense. Is it something we want to encourage for all commits?

mrob95:
I'm not sure, it's really just a matter of taste and probably doesn't make much difference. I'm happy to let people format things as they think best.

LexiconCode:
I have to say this makes it much easier to read, great idea!

PR #605, opened 2019-06-25
Further tidying of rule loading

mrob95:
Trying to do changes which touch a lot of files all in one go so that it doesn't cause too much disruption. Mainly just removing the last manual grammar loadings and removing direct calls to the CCRMerger.

I have also begun the task of moving unnecessary stuff out of `_caster.py`, moving the loading of self modifying rules into their respective files to enable easier modification in the future.

### DIFF
--- a/_caster.py
+++ b/_caster.py
@@ -65,9 +65,7 @@ from castervoice.lib.dfplus.additions import IntegerRefST
 
 from castervoice.lib.dfplus.merge.mergepair import MergeInf
 from castervoice.lib.ccr import *
-from castervoice.lib.ccr.recording.again import Again
-from castervoice.lib.ccr.recording.bringme import bring_rule
-from castervoice.lib.ccr.recording.alias import Alias
+from castervoice.lib.ccr.recording import again, bringme, alias
 from castervoice.lib.ccr.recording import history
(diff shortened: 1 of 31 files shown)


_caster.py
```diff
@@ -270,26 +253,8 @@ def generate_sm_ccr_choices(nexus):
     ]
     defaults = {"n": 1, "nnv": 1, "text": "", "volume_mode": "setsysvolume", "enable": -1}
 
+control.non_ccr_app_rule(MainRule(), context=None, rdp=False)
 
-grammar = Grammar('general')
```
LexiconCode: feature switches for `alias` and `again`have been removed?

_caster.py
```diff
@@ -270,26 +253,8 @@ def generate_sm_ccr_choices(nexus):
...
-main_rule = MainRule()
-gfilter.run_on(main_rule)
-grammar.add_rule(main_rule)
-
-gfilter.run_on(bring_rule)
-grammar.add_rule(bring_rule)
-
-if settings.SETTINGS["feature_rules"]["again"]:
```
LexiconCode: feature switches for `alias` and `again` have been removed?

PR #606, opened 2019-06-25
Improve rule naming in rdescripts

mrob95:
So that instead of e.g. "VSCodeNonCcrRule: command" it will just print "VSCode: command"


LexiconCode:
Nice improvement and it tests out.

PR #608, opened 2019-06-26
Fixed Selfmodrule filter bug

mrob95:
To recreate:
1. Add a filter rule in words.txt e.g. "tango -> trap"
2. Add an alias with the same name e.g. "Alias tango"
3. restart

Produces: 
```
  File "C:\Users\Mike\Documents\GitHub\casterfork\castervoice\lib\dfplus\merge\gfilter.py", line 221, in spec_override_from_config
    rule._mcontext, rule._mwith)
TypeError: __init__() takes exactly 2 arguments (11 given)
-- skip unchanged wrong grammar file: C:\Users\Mike\Documents\GitHub\casterfork\_caster.py
-- skip unchanged wrong grammar file: C:\Users\Mike\Documents\GitHub\casterfork\_caster.py
```
I assume this has pretty much been the case since forever as the chances of finding it are fairly small.

Fixed for the moment by not running filters on self modifying rules. It doesn't really make sense to filter them anyway as they are set by the user, and filters are mainly used for modifying caster defaults.

### DIFF
--- a/castervoice/lib/ccr/recording/alias.py
+++ b/castervoice/lib/ccr/recording/alias.py
@@ -91,7 +91,7 @@ class ChainAlias(AliasRule):
 _NEXUS = control.nexus()
 
 if settings.SETTINGS["feature_rules"]["alias"]:
-    control.non_ccr_app_rule(Alias(_NEXUS), context=None, rdp=False)
+    control.non_ccr_app_rule(Alias(_NEXUS), context=None, rdp=False, filter=False)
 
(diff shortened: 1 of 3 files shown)


LexiconCode:
At first I thought you meant that you disabled it for all filters on self modifying rules.
Looking over the code you disabled it specifically for `alias` and 'bring_rule' in a simple but clever way. 

In the future self modifying rules could be used in scenarios were filters would be appropriate which your implementation allows.

PR #610, opened 2019-06-26
Classic install: Enhancements to check dependencies for required versions and missing dependencies.

LexiconCode:
When developing Caster sometimes dependencies change between releases or require a locked version. This whole request implements a method of informing users of the change in dependencies.

- Allows maintainers to define minimum and specific version of dependencies. To be used between releases as a method of informing contributors of changes with dependency.
- Checks for missing dependency reading `requirements.txt`
- Enhanced message that guides the user to install dependencies.
- Suppresses traceback
- If a dependency is missing the main thread sleeps for 15 seconds with both WRS and Natlink
- Moves all dependency related functions out of `_caster.py` to `dependencies.py` except for the grammar.

### DIFF
--- a/_caster.py
+++ b/_caster.py
@@ -8,24 +8,6 @@ import os, time, sys
 import logging
 logging.basicConfig()
 
-
-def version_minimum():
-    try:
-        import pkg_resources
-        version = "0.15.0"  # Minimum Version of Dragonfly2 need for Caster
-        pkg_resources.require("dragonfly2 >= %s" % (version))
-    except Exception:  # pylint: disable=broad-except
-        pass
-        print("\nCaster will not start! \
-            \nCaster: Requires at least Dragonfly version %s\
-            \nUpdate Dragonfly 'pip install --upgrade dragonfly2' from cmd\
-            \n\nThis window will close in 30 seconds" % (version))
-        time.sleep(30)
-        os._exit(1)
-
-
-version_minimum()
-
 import time, socket, os
 from dragonfly import (Function, Grammar, Playback, Dictation, Choice, Pause, RunCommand)
 from castervoice.lib.ccr.standard import SymbolSpecs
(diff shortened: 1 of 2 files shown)


LexiconCode:
Potentially this could grow into a check for any missing/required version dependencies. For example `ImportError: No module named mock` for the scenario when new dependencies have been added using the classic install during contributors updating Caster with git.

LexiconCode:
To test
1. `pip install dragonfly2==0.14.1`
2. restart Dragon
3. observed output in the `status window`

After testing revert back to the latest version of dragonfly.
- `pip install --upgrade dragonfly2`

mrob95:
Not 100% sure breaking out of start-up if the user has the wrong version is a good idea. Partial functionality may be preferable to no functionality for running terminal commands or if the user is away from Internet. Also it may not pick up dragonfly installations updated through git.

LexiconCode:
> Not 100% sure breaking out of start-up if the user has the wrong version is a good idea. Partial functionality may be preferable to no functionality for running terminal commands.

That's a good point. I should reword the use case. 
- Only in the circumstances were Caster would crash ~~or cause strange errors~~ if they're using an old dependency or the project requires a new dependency.
- `PIP` install should never have an issue with dependencies and this means it's for `classic` only.

Caster would not load anyway without upgrading/added the dependency. Therefore utilizing the terminal commands for upgrading would not be an option. As for situations without Internet if they are upgrading Caster then they might as well upgrade Dragonfly as well since it will crash without the minimum version.

> Also it may not pick up dragonfly installations updated through git.

I have tested with `python setup.py install`. Either the repository keeps the same version number and if it's bumped up still considered higher than the minimum.

All that being said partial functionality could be enabled.

LexiconCode:
If there was a way to cancel caster from loading yet leave the background engine running that would be preferable.

LexiconCode:
I am sure the code could be much more elegant and opp but it is functional.

PR #612, opened 2019-06-27
App settings removed in settings.py

LexiconCode:
The ability to toggle application stock grammars through settings.toml has been removed.
Therefore the configurations in settings.py To generate the setting should be removed as well.

Related to #596


LexiconCode:
For some reason it's including _Caster.py. however get does not show it being included locally. I will re-create the issue and branch.

Issue #575, opened 2019-06-11
file to bring me as name not working


kendonB:
Saying "file to bring me as file name" currently adds the folder name in the file dialog to the file section of the `bringme.toml` file. I'm actually not sure if there is a feasible way to allow bring me add to work for files. 

Unless someone has an idea of how to achieve this, we should disable `file` as an option for bring me add. Perhaps a nice middle ground would be to have the `bringme.toml` file open at the location of the new line.

Other ideas are welcome.

comodoro:
What it did was take highlighted text, assume it is a file path and add it to things to launch. I suppose this functionality could be restored, plus check if file exists, plus maybe actually detecting current file if in Windows Explorer

kendonB:
I have a "Copy as path" in the context menu on mine with keyboard shortcut "a". So fixing this could be as simple as doing `Key("s-f10, a")` which also looks like it works for folders.

PR #614, opened 2019-06-27
Bring me updates

mrob95:
I've tidied up this code following the basic pattern of #577 , using a single object to keep track of config rather than calling a bunch of separate functions.

I've also implemented context dependent folder actions for terminal and Windows Explorer as mentioned in #476, so "bring me my documents" will "cd" if a terminal window is active, and will use a focused file explorer window instead of creating a new one.

Happy to get feedback on whether or not people think this is the right approach.


LexiconCode:
I think it's a very reasonable way to go as it reduces duplicate commands and provides a standard workflow. It would be interesting to see what other areas we can brainstorm to expand this functionality further for the roadmap. 

Did you run into this issue while working on the Bring me updates?
https://github.com/dictation-toolbox/Caster/issues/575

LexiconCode:
This is a good example of a `SelfModifyingRule` it's very easy to follow I appreciate that as many of Caster existing examples are convoluted.

LexiconCode:
A significant improvement and definitely much easier to read code. I do not experience the issues described in #575

Everything works the expected except for "restore bring me defaults" producing no error message or restoring the default bringme keys. A not as clean workaround is to overwrite the bringme.toml with the default instead of trying to load the defaults from the source location. Overall I prefer the second method.

One suggestion is to include the context for windows powershell as well.

mrob95:
Regarding defaults, I think it makes sense to define them in a dictionary in the same file rather than a separate toml file. Mainly because it keeps everything in one place and easily readable.

I've also modified the functionality so that restoring defaults also overwrites the config file. Thoughts?

LexiconCode:
> 
> 
> Regarding defaults, I think it makes sense to define them in a dictionary in the same file rather than a separate toml file. Mainly because it keeps everything in one place and easily readable.
> 
> I've also modified the functionality so that restoring defaults also overwrites the config file. Thoughts?

Perfect a bit of clean up the `bringme default path`/`bringme.toml.defaults`/ `the function that restores `bringme.toml.defaults` to the data directory`.

LexiconCode:
After further testing https://github.com/dictation-toolbox/Caster/issues/575 however I'm not sure if that's what you want in the scope of this pull request.

Second thoughts on a spec change ` "<launch> to bring me as <key>":` is difficult for me to say. "bring me <launch> as <key>" seems much more natural. thoughts?

LexiconCode:
Another concept that I thought interesting. `alias <launch> as <key>`
The command `bring me` would be used across caster to retrieve `x` .

mrob95:
I'll think about it, can probably use a context action for this as well

LexiconCode:
This is debatable but possibly functional for those that manually edit the `bringme.toml` manually by preference or testing purposes.
```
            "refresh bring me":
                R(Function(self.refresh)),
```

kendonB:
All of these additions seem to work well. What do you think about adding the following:

```
"bring me <desired_item> new"
```
This one would always open a new window to open the desired item. If the current context is a terminal, folders would open in a new terminal window. 

```
"bring me <folder_item> terminal"
"bring me <folder_item> [windows] explorer"
```
These would operate by 1) looking for an open instance of one of these types of windows, 2) navigating to that folder in an open instance if one exists; if one does not exist, opening a new window at that folder.
```
"bring me <folder_item> new terminal"
"bring me <folder_item> new [windows] explorer"
```
Like the above but it always opens a new window.

kendonB:
Would also be good to incorporate all of this into the `file_dialogue` context.

mrob95:
Okay I've done another rearrange and implemented some of the above. Websites and programs/files work the same but for folders:
* `bring me <folder>` is as before, context-dependent. If you are in an Explorer or terminal window it will use that, otherwise it will open a new Explorer.
* `bring me <folder> in explorer` opens in a new Explorer window regardless
* ditto for `bring me <folder> in terminal`

I'm not a powershell user so not sure how well this will work for that.

I think having a separate method for each item type is neater and will make for easier scaling if we need to add more categories.

LexiconCode:
@mrob95 I made a few tweaks and if you don't feel of the relevant they can be removed.
- Updated hyperlink for caster and dragonfly. 
- Tweaked order of keys
- For Caster and dragonfly modified "<project> documentation" to call documentation per project.

LexiconCode:
A possibleweakness of `"%USERPROFILE%`?
`My Documents` on my system is pointed to an alternate location.
image

Therefore with `command prompt/powershell` context with `bring me my documents` produces unexpected file path when it should be `D:\Backup\Library\Documents`
```
C:\Users\Main>cd "C:\Users\Main\Documents"
The system cannot find the path specified.
```

When dictating `bring me my documents` out of the above context the command opens up my documents folder as expected.

LexiconCode:
You may not be aware but you could leverage these as well.

Dragonflies [runcommand-action](https://dragonfly2.readthedocs.io/en/latest/actions.html#runcommand-action)
and
Caster's [TerminalCommand](https://github.com/dictation-toolbox/Caster/blob/master/castervoice/lib/terminal.py)

mrob95:
Hmm, I'm surprised that it works in one context but not in another. Maybe try adding the following to `_rebuild_extras` to see what is going on:
```
print({key: os.path.expandvars(value)
            for key, value in section.iteritems()}
            for header, section in self.config.iteritems())
```

I'm aware of `RunCommand` but `Popen()` is actually cleaner than `RunCommand().execute()`. Happy to change it though if we want to get rid of Popen across the code base.

LexiconCode:
> _rebuild_extras
I assume you mean `_rebuild_items`
```
    def _rebuild_items(self):
        # E.g. [Choice("folder", {"my pictures": ...}), ...]
        print({key: os.path.expandvars(value)
            for key, value in section.iteritems()}
            for header, section in self.config.iteritems())
        return [
            Choice(header, {key: os.path.expandvars(value)
            for key, value in section.iteritems()})
            for header, section in self.config.iteritems()
```


> I'm aware of `RunCommand` but `Popen()` is actually cleaner than `RunCommand().execute()`. Happy to change it though if we want to get rid of Popen across the code base.

I would say it's not as clean, However think there's also utility and advantage to leveraging `RunCommand` Therefore if your happy to change Popen across the code base go for it.

PR #616, opened 2019-06-27
Redesign rule loading

mrob95:
As outlined in #589 , this is at least a first attempt at allowing rules in user space to automatically override caster defaults. To test, copy (for example) `lib/ccr/python/python.py` into `.caster/rules` and make some alterations. Rules in the user directory are imported first, and then rules with clashing pronunciations are ignored.

A few points:
* This needs thorough testing, I am almost certain that it will be broken in multiple ways.
* In a few places I have commented old stuff without removing it.
* If anybody knows a better way of recursively globbing for filenames then I'm all ears. `recursive` argument was only added in py3.
* At the moment this only works for rules, not for example the app or symbol standards.
* At the moment both the file in user space and the default file are imported, this may slow things down if the user moves a lot of rules so some sort of filename filtering will probably be necessary.
* `rules` and all its subdirectories will need init files so that the rules can be imported. This seems like it may turn out to be a major annoyance and a FAQ-factory. It may be worth generating these automatically if they don't already exist.

Also if anybody has any better ideas about how rules are loaded and how user space/caster space/git pulls/pip installs should coexist then fire away. Theoretically I think this is a decent approach but we will see.


LexiconCode:
> 
> 
>   * `rules` and all its subdirectories will need init files so that the rules can be imported. This seems like it may turn out to be a major annoyance and a FAQ-factory. It may be worth generating these automatically if they don't already exist.

This doesn't have logic to check if they already exist but it's a start.

```Python
from os.path import isdir, walk, join

rootFolder = "C:\\Users\\Main\\Desktop\\\inittest"
init = '__init__.py'
def generateInit(arg, dirname, fnames):
    fnames = [fname for fname in fnames if isdir(fname)]
    print "Created %s in : %s" %(init, dirname) # debug
    with open(join(dirname, init), 'w') as file_: file_.write('')

walk(rootFolder, generateInit, None)
```

Output
```
PS C:\Users\Main> & C:/Python27/python.exe c:/Users/Main/Desktop/inittest/test.py
Created __init__.py in : C:\Users\Main\Desktop\\inittest
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1\level 2 folder 1
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1\level 2 folder 2
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1\level 2 folder 2\level 3 folder 1
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1\level 2 folder 2\level 3 folder 1\level 4 folder 1
Created __init__.py in : C:\Users\Main\Desktop\\inittest\level I folder 1\level 2 folder 2\level 3 folder 2
```

LexiconCode:
Note once this pull request is complete https://github.com/dictation-toolbox/Caster/pull/619 I'll need to document based on this implementation.

mrob95:
Okay it seems to be working reliably. Next two pull requests will likely be:
* Restructuring rules folders
* Docs

LexiconCode:
A few thoughts.

I think `copy_rules/copy_filters` should be under UserContentManager class. I left them as separate functions originally to simplify merging as you're still working on this request pull. In addition I'm still working on reading object-oriented Python.

Thinking of the future of cross-platform instead of concatenating strings `os.path.join()` and `os.path.normcase(path)`/`os.path.normpath(path)` instead of `.replace()`.

One thought is to continue the function of checking the rules/filters loaded correctly (not a syntax error issue) if not to provide feedback. Reference https://github.com/dictation-toolbox/Caster/blob/5cd51087577420794a2b5bb4ef7f507ff920b736/castervoice/lib/ctrl/user.py#L67

I will work on testing it out today.

LexiconCode:
In addition a typo on my part pull. 
https://github.com/dictation-toolbox/Caster/blob/68e7763e56580430c13942dc8a3084b6981c55da/castervoice/lib/ctrl/user.py#L27

Should be renamed `USER_RULES_DEFAULTS_PATH`

castervoice/lib/ctrl/user.py
```diff
@@ -35,17 +35,17 @@ class UserContentManager(object):
     def __init__(self):
         # self.rules = self.import_user_dir("get_rule", settings.SETTINGS["paths"]["USER_DIR"] + "/rules")
         # self.filters = self.import_user_dir("get_filter", settings.SETTINGS["paths"]["USER_DIR"] + "/filters")
-        self.caster_dir = os.path.realpath(__file__).rsplit(os.path.sep + "castervoice", 1)[0].replace("\\", "/") + "/"
-        self.user_dir = settings.SETTINGS["paths"]["USER_DIR"] + "/"
+        self.caster_dir = os.path.realpath(__file__).rsplit(os.path.sep + "castervoice", 1)[0].replace("\\", "/")
```
LexiconCode: I see a variation of this all over CodeBase. Why aren't we utilizing something like `CASTER_ROOT_DIR = "Root castervoice Path"` or/and included as an import?

mrob95: Sheer laziness ^^

LexiconCode: Good to know. In the near future all go through and try to correct these into something more standardized. I wouldn't worry about it here.

LexiconCode:
`__init__` successfully generated although the printed file paths aren't correctly slashed
```
Created __init__.py in : C:/Users/Main/.caster\rules
Created __init__.py in : C:/Users/Main/.caster\rules\examples
```

LexiconCode:
These might be you might be aware already but "development" rule in `dev.py` is not loaded.

LexiconCode:
After searching for MergeRule and MergeRule in caster there's a few more.

It's not quite right but an attempt
```Python
self.import_dir(join(self.caster_dir, "asynch"), "castervoice.asynch")

self.import_dir(join(self.caster_dir, "asynch", "sikuli"), "castervoice.asynch.sikuli")
```

Relative file paths
```
castervoice\asynch\sikuli\sikuli.py # sikuli commands

castervoice\asynch\_hmc.py # UI commands
```

LexiconCode:
Tweaking the way the Example rules filters are copied. We should be able to update examples rules/filters and subfolders. The should not be an issue as nothings loads from the example folders.

castervoice/lib/ctrl/user.py
```diff
@@ -25,6 +25,8 @@ def load_rules(self):
         self.import_dir(join(self.caster_dir, "apps"), "castervoice.apps")
         self.import_dir(join(self.caster_dir, "lib", "ccr"), "castervoice.lib.ccr")
         self.import_dir(join(self.caster_dir, "lib", "dev"), "castervoice.lib.dev")
+        self.import_module("castervoice.asynch.sikuli", "sikuli")
```
LexiconCode: This loads an individual file versus a directory.

LexiconCode:
I'm not convinced with the changes made in "Fixed context loading confusion". My thought would be context should always be placed with applications or grammars that there belong to instead of buried in core code.

While it does present a nice clean look it takes away readability and accessible expandability.

mrob95:
> I'm not convinced with the changes made in "Fixed context loading confusion". My thought would be context should always be placed with applications or grammars that there belong to instead of buried in core code.
> 
> While it does present a nice clean look it takes away readability and accessible expandability.

I agree I am also not happy with it. The problem is that as it was contexts like the terminal one were being imported from the gitbash grammar in e.g. bringme.py. This is not necessarily a problem but if the user is loading their own gitbash.py then the caster one will also be loaded to get the context. This produces a message on the terminal saying that a rule with that pronunciation also exists and may lead to the wrong rule being loaded depending on the ordering.

Solution for now is probably just to define the same context in a couple of places, I don't think it's a huge deal.

LexiconCode:

PR #618, opened 2019-06-30
Add automatic installation script for classic install

mrob95:
Batch script to do a couple of things automatically during the classic caster installation:
* Store the location of the current directory (wherever the user has installed caster)
* Move to the python installation directory (in case the user has not added python to their path)
* Install dependencies from requirements.txt
* Set the Natlink user directory to the installation location

I think this should work fine but don't have a good way of testing it from scratch. I will think about how to make it more robust to directories being differently named.


LexiconCode:
Current I should be able to test by morning late tomorrow morning. 

I see what you mean by alternate paths for the python install and natlink.
We could use install.py which executes .bat and allows to pull strings from the install.py file. That way we could obtain the paths Python functions.

LexiconCode:
This works as expected although it should be noting I have nothing unusual about my Python on or Natlink install location.

- it sets the user directory regardless if Natlink is enabled or not.
- it does not enable Natlink.

LexiconCode:
In order to enable natLink the command has to be executed as with elevated privileges. See `start_natlinkconfigfunctions.py `

```
do action: Enabling NatLink:
  File "C:\NatLink\NatLink\confignatlinkvocolaunimacro\natlinkconfigfunctions.py", line 2201, in <module>
    _main()
  File "C:\NatLink\NatLink\confignatlinkvocolaunimacro\natlinkconfigfunctions.py", line 1442, in _main
    func(None) # dummy arg
  File "C:\NatLink\NatLink\confignatlinkvocolaunimacro\natlinkconfigfunctions.py", line 1821, in do_e
    self.config.enableNatlink()
  File "C:\NatLink\NatLink\confignatlinkvocolaunimacro\natlinkconfigfunctions.py", line 813, in enableNatlink
    if not self.isElevated: raise ElevationError("needed for enabling NatLink")
__main__.ElevationError
```

mrob95:
Yep, I don't think this is a huge problem for a few lines of batch script. The user can easily verify that it isn't doing anything malicious.

LexiconCode:
> 
> 
> Yep, I don't think this is a huge problem for a few lines of batch script. The user can easily verify that it isn't doing anything malicious.

Leveraging a little bit of knowledge from stackoverflow I added a permissions check.

LexiconCode:
We should probably distinguish this as an installer for DNS versus WSR or any other engine.

LexiconCode:
@mrob95 would you think of a companion py file? that allows to return arguments from the e.g install.py file. That way we could obtain the paths from Python functions.

LexiconCode:
A few ways to get at the Python install location without knowing path

-  From Python
	- `python -c "import sys; print sys.executable"`
- The registry   
	- HKLM\SOFTWARE\Python\PythonCore\versionnumber\InstallPath
    - HKLM\SOFTWARE\Wow6432Node\Python\PythonCore\versionnumber\InstallPath

LexiconCode:
Another interesting method is executing python directly from `.bat`

```
0<0# : ^
''' 
@echo off
echo this is batch code
goto check_Permissions

:check_Permissions
    echo Caster: Administrative permissions required to setup Natlink. Detecting permissions...

    net session >nul 2>&1
    if %errorLevel% == 0 (
        echo Success: Administrative permissions confirmed.
    ) else (
        echo Failure: Current permissions inadequate restart install.bat with administrator privileges...
	pause >nul
    )


python %~f0 %*
exit /b 0
'''
print("This is python code")
```

mrob95:
I agree that using Python to find directories would be nice, but it would rely on already knowing where Python is installed, since I don't want to rely on the right version of Python being in the user's PATH.

I'll write up some installation instructions for using this file and then my instinct would be to see how many people have trouble with it due to non-standard directories. I suspect most people will just let Python and Natlink install where they want, in which case this script should work fine.

mrob95:
Okay, I have rejigged the installation instructions to emphasise the classic install with the new automatic batch configuration. I think it is good to emphasise only one installation method rather than offering new users a confusing choice, and I think this is the best method we have. No need to do anything on the command line, and the user knows exactly where all the files are.

I have also added a check in the installation file to see if Natlink is present, if not then it will just install dependencies, so this works for WSR users as well.

PR #620, opened 2019-07-01
Create CODE_OF_CONDUCT.md

kendonB:
This is the standard code of conduct recommended by GitHub. The only edits from the original are to add "political affiliation" to the list of groups and to define our project spaces.


kendonB:
It still requires an email before merging.

Issue #630, opened 2019-07-03
Windows Explorer and folders grammar


comodoro:
Detection of current directory and selected files in Windows Explorer can actually be done via Shell.Application COM object. The code is really simple once you know it. My problem currently is I do not know where to put it. `explorer.py` in apps seems like an obvious choice, but this can also be used in bring me (#575, #499?) and I would like to avoid depending on app grammar. Directly in `lib/` is getting crowded; a new directory under `lib/`?

Note also that this simplifies making Unimacro-like "folders" grammar as suggested some time ago by @alexboche (I actually did not get this working, but the biggest complication seemed to be reliable system independent folder detection).

quintijn:
@comodoro : I fixed a few bugs in this _folders grammar. Due to a migration of Natlink and Unimacro to github (by Jungar), I will pause on this grammar until this is finished. After that, I will build in the "remember" idea, that was posed here. Hope the grammar will then work and be useful to others as well!

I think getting the correct current folder is most important. All other file and folder getting things is standard programming. With "messagesfunctions.py" I managed to get the current folder both when the current window is top and Class CabinetWClass and when window is child and Class #32770.

I also needed to do quite a bit of work to get the so called Library folders (like Documents, Downloads, but also for example Dropbox) correct. The functions for this are in the Natlink core directory.

comodoro:
Unfortunately I did not get all of Unimacro working, or only without most grammars - for example the nice grammar explorer window was possible to invoke by voice, but didn't contain any grammars.

I initially went the path of window hierarchy as well, but as you write, for library folders it is quite complicated and I did not see a simple way. I hope the COM approach, which uses "shell scripting API", or how they call it, does not have serious drawbacks.

quintijn:
These two functions get the folder from an explorer window (CabinetWClass) and a Dialog (#32770):
```
def getFolderFromCabinetWClass(hndle):
    """extract folder from a CabinetWClass window
    
    (Unimacro grammar _folders)
    """
    controls = findControls(hndle, selectionFunction=selFuncExplorerAddress)
    if controls:
        hndle = controls[0]
        text = getwindowtext(hndle)
        ## result is probably u"Address: C:\Documents" or so
        folder = text.split(": ", 1)[1]
        return folder
            

def getFolderFromDialog(hndle, className):
    """dialog for open etc from office etc
    
    only print details if Adres: of Address: is found... (Dutch English windows versions currently)
    """
    if className == '#32770':
[...]

PR #637, opened 2019-07-05
Bring me for files and folders

### LINKED ISSUE #575: file to bring me as name not working

comodoro:
Adding files and folders should work from Windows Explorer (and everywhere files can be copied). Closes #575. Path detection in the end with copying and `win32clipboard` Remarks:

- There was a bug, adding anything did not work?
- Uses Dragonfly's clipboard as well as pywin32's win32clipboard
- The link "caster discord" displays an expired Discord invitation - is there a Discord channel?

### DIFF
--- a/castervoice/lib/ccr/recording/bringme.py
+++ b/castervoice/lib/ccr/recording/bringme.py
@@ -13,7 +13,7 @@ class BringRule(SelfModifyingRule):
             "bring me <folder> [in <app>]": R(Function(self.bring_folder)),
             "bring me <file>": R(Function(self.bring_file)),
             "refresh bring me": R(Function(self.load_and_refresh)),
-            "<program> to bring me as <key>": R(Function(self.bring_add)),
+            "<launch> to bring me as <key>": R(Function(self.bring_add)),
             "to bring me as <key>": R(Function(self.bring_add_auto)),
             "remove <key> from bring me": R(Function(self.bring_remove)),
             "restore bring me defaults": R(Function(self.bring_restore)),
@@ -80,8 +80,12 @@ class BringRule(SelfModifyingRule):
             if not path:
                 # dragonfly.get_engine().speak("program not detected")
                 print("Program path for bring me not found ")
-        # elif launch == 'file':
-        # no way to add file via pyperclip
+        elif launch == 'file':
+            files = utilities.get_selected_files(folders=False)
+            path = files[0] if files else None # or allow adding multiple files
+        elif launch == 'folder':
+            files = utilities.get_selected_files(folders=True)
+            path = files[0] if files else None # or allow adding multiple folders
         else:
             Key("a-d/5").execute()
             fail, path = context.read_selected_without_altering_clipboard()
--- a/castervoice/lib/utilities.py
+++ b/castervoice/lib/utilities.py
@@ -7,6 +7,7 @@ import json
 import os
 import re
 import sys
+import time
 import traceback
 from __builtin__ import True
 from subprocess import Popen
@@ -14,6 +15,9 @@ import toml
 
 import win32gui
 import win32ui
+import win32clipboard
+
+from castervoice.lib.clipboard import Clipboard
 
 from _winreg import (CloseKey, ConnectRegistry, HKEY_CLASSES_ROOT,
     HKEY_CURRENT_USER, OpenKey, QueryValueEx)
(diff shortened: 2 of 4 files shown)


quintijn:
Hi @comodoro, if this has to do with the "remember" function I am working on in the Unimacro grammar _folders, please elaborate. Otherwise leave it. Thanks.

comodoro:
This detects the file or folder by copying it (to saved and cleared clipboard) and then reading it with `win32clipboard` (the code added to `utilities.py`) and restoring the clipboard. The reading (`get_clipboard_files`) includes reading available clipboard formats because 1)win32 api is stupid and 2) in Explorer some weird data objects got copied as well as the actual file. If no files are present, at least it looks at the text there and if it is file or folder path, adds that.

comodoro:
The copying part is however dependent on Dragonfly's `Clipboard` and `Key`, which I do not know if Natlink provides (for handling clipboard and pressing CTRL-C)

quintijn:
1. Natlink only has natlink.getClipboard(), provided for testing purposes. In Unimacro I use the win32clipboard functions.

2. What would you like to copy in this case? The whole file, the file name, the path of the file?

comodoro:
This handles both if the user is in Windows Explorer on a file, then copies the file to clipboard and reads its path from the pointer in clipboard, or has selected text, then it checks that it is really an existing path

quintijn:
Why like this? In the Explorer you get the file name via the COM function, shell.Dispatch or so, in a #32770 dialog, you have the folder, and get the filename by copying the file with {f2}{ctrl+c}. Then with python functions you get all the information you want.

LexiconCode:
>     * The link "caster discord" displays an expired Discord invitation - is there a Discord channel?

We do have a discord room `https://discord.gg/9eAAsCJ` the URL should not expire.

comodoro:
@quintijn This works in more places like file managers. By the way if I copy the file from the #32770 dialog anyway, I will not save that much code.

comodoro:
@LexiconCode OK, I have joined. VS Code must have added the ending quote character when activating it with ctrl+click.

quintijn:
@comodoro : thanks, I will try that! where can I find the utilities code? I cannot however imagine what other file managers there could be. Please give me some examples.

comodoro:
All changes in the pull request can be viewed if you click on the "Files changed" tab near the top of this page. Utilities.py is down there. Or you can view the whole file in the branch: https://github.com/comodoro/caster/blob/bringme-1/castervoice/lib/utilities.py. Added functions are at the bottom.

I personally often use Total Commander as file manager, but there are many others I have not tried like QDir, fman. Google will tell you more.

quintijn:
Quite right!! I am going to try the clipboard approach! Thanks, Vojtech!

LexiconCode:
@comodoro let me know when you feel this is ready to be complete. At that time I'll take a closer look and run through some testing.

quintijn:
Seems to work perfect! Always the CF_HDROM flag of the clipboard.

quintijn:
I think, when making a general "utilities" and "actions" library independent of either of the packages (Natlink raw, Unimacro, Dragonfly, Caster), an improved Clipboard class should also be made. I am impressed by the insight Christo gives in his Clipboard class. But additional useful methods could be made. For example, save/restore as automatically as possible. 

My top question is, how long to wait after a ctrl+c before the new clipboard can be accessed. Reducing this, possibly by testing the state  in a loop with shorter wait intervals, in order to proceed as quick as possible, could be incorporated in such a class. I did several of this in Unimacro, and this code can also contribute to unifiy more stable and quick clipboard operations.

How to detect as "cheap" as possible if the clipboard has changed. If it has changed, is is also filled complete?

My third question (now) is, whether and how the reaction time increases if the clipboard becomes larger. Eg data in Word or larger portions of text. That is especially useful to know when doing dictation/selecet&say base upon information of the clipboard, as Alex Boche has recently written, I believe. Can you just copy the whole buffer or should you be economic with only 10 lines on either side of the cursor.

quintijn:
@comodoro : I now used the clipboard for getting folders and files from a dialog. For CabinetWClass (explorer) still took the COM method (is it com? I think so).

Please see two short videocaptures, made for Alex Boche, which demonstrate. The first one in CabinetWClass, the second (smaller) one in a dialog window. 

https://we.tl/t-sK7jCAp0LJ

Greetings, Quintijn

LexiconCode:
> 
> 
> I think, when making a general "utilities" and "actions" library independent of either of the packages (Natlink raw, Unimacro, Dragonfly, Caster), an improved Clipboard class should also be made. I am impressed by the insight Christo gives in his Clipboard class. But additional useful methods could be made. For example, save/restore as automatically as possible.

I would be in support of this and willing to start the process. Thoughts @comodoro?

comodoro:
@LexiconCode I think you can merge it, I have tried to add a topic to readthedocs, but did not try to build it on my side, feel free to change anything.

comodoro:
@quintijn My answer to these questions is perhaps universally "I don't know". As you correctly suspect, the 0.1s waiting time is arbitrary and overkill, but the main time consumer here is still Dragon and user reactions. Problem with setting this is it is system dependent. You could I suppose periodically check the clip[board with increasing time intervals, if you feel it is worth the effort. As for the last question, it is possible that with large chunks of text the function will be slower.

We can make a separate package, sure, but we will have to maintain it and in case of Caster I fear (as it would probably be Windows-only) that it would be one step further from platform independence. I am afraid I am not sure what you have in mind with the Clipboard class,.

quintijn:
@comodoro: the 0.1 waiting time is overkill in general, but maybe with large selections it is not. If it is used all the time, it could be worth mimimizing this time. And when it is used in the gotBegin of a grammar, it is done at every utterance. We should know, after emptying the clipboard, how to test if the clipboard is filled again. That's the technical questions at least. Irrespective of platform indepence and how to organise things in how to maintain.

comodoro:
Isn't it done after recognizing the grammar, i.e. once after the utterance in response to the user? And if the user has selected too much text and the function fails, I am inclined to call it his problem:)

It looks like there are clipboard monitoring APIs, but polling after some time intervals is easiest/most reliable, e.g. on stackoverflow.

quintijn:
When grammar lists have to be filled, for example the sub files and sub folders of a "current" directory, this is done in gotBegin, so at every utterance. Fortunately this can be done without the clipboard.

Only the "this folder" command of my grammar needs the clipboard.

But... when implementing select and say with help of the clipboard, you need to get the contents of the clipboard after a control-c at every utterance.

I do not agree with your "his problem" opinion. The functionality should just work as good and as efficient as possible.

But in short:
- for actions that need only to be done after a recognition of a command speed is not crucial
- for things that need to be done at every utterance, speed is crucial.

If you found a link about polling for the clipboard, please share it here!

comodoro:
Yes, for folders grammar it could be a problem. To be frank I am already not impressed with it either, since it sends the ctrl+c keystroke regardless of where the user is, so if "adding a folder to bring me" in a terminal window with a running computation, he could be nastily surprised. This can be avoided by checking that we are in a supported window.

Checking for clipboard data, here is simply (along the lines of) what people do: https://stackoverflow.com/a/25900882/999198, here are some Windows APIs: https://docs.microsoft.com/en-us/windows/win32/dataxchg/using-the-clipboard#monitoring-clipboard-contents

quintijn:
This was very helpful. Let us continue the discussion in thread 630 (https://github.com/dictation-toolbox/Caster/issues/630)

PR #640, opened 2019-07-11
fixed punctuation

kendonB:
This pull request reverses some incorrect changes to `punctuation.py` in #590 where some repeats disappeared. @mrob95 can you think of any other places that this type of error could have occurred?


LexiconCode:
Good catch. Let me know when you're ready for merge.

kendonB:
I'm going to merge this now. I had a quick look through the pull request that caused this problem and it looks like the only place it happens.

PR #642, opened 2019-07-11
Make tests discoverable and not check for updates

comodoro:
Makes unittests discoverable by unittest and simplifies loading and adding them. Fixes update messages with every test.

This shows in my opinion that 1. initializing variables as a side-effect of loading a module (and the `global` keyword in general most of the time) should be avoided, and 2. the God-Emperor `nexus` object should be gradually toned down instead of added to


LexiconCode:
Thank you for the feedback!

A few things 

1. The `nexus` is pretty new to me. I know it acts like a controller of sorts passing around variables and states.
2. I agree that global variables should be avoided but what would you recommend for  `pip_path` and `update`? Ultimately I think `pip_path` should be recorded in the settings.toml paths. 
3. Alternatively update checks can be disabled with `settings.SETTINGS["miscellaneous"]["online_mode"]` set to `False`. Use your implementation at your discretion. 
4. Have we changed the location of where DependencyMan is initialized? `castervoice\lib\ctrl\nexus.py` via `self.dep = DependencyMan()` to `_NEXUS.dep.initialize()` `in _caster.py`?

comodoro:
Let me express my opinion, anyone is welcome to discuss and/or correct me.

> 1. The `nexus` is pretty new to me. I know it acts like a controller of sorts passing around variables and states.

Yes, but can't stuff just be passed around whenever needed? One big class depends on lots of stuff, which basically means that code importing it needlessly depends on lots of other  code.

> 2. I agree that global variables should be avoided but what would you recommend for  `pip_path` and `update`? Ultimately I think `pip_path` should be recorded in the settings.toml paths.

I would probably put it in a class. By the way cannot the search for pip be done without relying on default Python path?

> 3. Alternatively update checks can be disabled with `settings.SETTINGS["miscellaneous"]["online_mode"]` set to `False`. Use your implementation at your discretion.

OK

> 4. Have we changed the location of where DependencyMan is initialized? `castervoice\lib\ctrl\nexus.py` via `self.dep = DependencyMan()` to `_NEXUS.dep.initialize()` `in _caster.py`?

Yes, to allow running tests without running the DependencyMan constructor all the time, but the other nexus code is still run with every test, even though not printed out.

LexiconCode:
> Yes, but can't stuff just be passed around whenever needed? One big class depends on lots of stuff, which basically means that code importing it needlessly depends on lots of other code.

That is an interesting discussion however I think we should probably pursue it in a separate issue rather than in a pull request.


> By the way cannot the search for pip be done without relying on default Python path?

I didn't write it but I don't think it's using the default path. It searches for an environmental variable that has "Python27\Scripts" in the path. While it does allow for alternate drive letters it does not account for a change in the file path during installation. 


> Yes, to allow running tests without running the DependencyMan constructor all the time, but the other nexus code is still run with every test, even though not printed out.

I checked and this doesn't change the expected result of missing dependencies. Thanks for clarifying!

LexiconCode:
> That is an interesting discussion however I think we should probably pursue it in a separate issue rather than in a pull request.

That discussion can take place in: https://github.com/dictation-toolbox/Caster/issues/646

PR #643, opened 2019-07-11
Remove unnecessary WSR-specific code

### LINKED ISSUE #554: Update some engine-specific code

Danesprite:
This should close #554.

I've also fixed the `recognize_forever()` call in `_caster.py` (while loop not required).


LexiconCode:
@Danesprite Testing this would WSR and the `record from history` command.
It doubles the entry of the history per command.

For instance if I say the command `maximize window` once it's listed twice `record from history` GUI.

Danesprite:
Thanks for testing it. WSR behaves weirdly with recognition observers. It's a dragonfly problem which I've neglected to make a GitHub issue for. Will do that tomorrow.

I believe this behaviour would also occur with the `RecognitionHistoryForWSR` class, although I'm not sure if it was actually receiving any recognition history because it wasn't registered.

LexiconCode:
`record from history` command is implemented in `castervoice\lib\ccr\recording\history.py`. Looking under `def record_from_history(self)` at `self.nexus.history[:]` it grabs the history from the nexus which is `RecognitionHistory(20)` Therefore I think it must be an issue with the observers.

Danesprite:
Okay I can't reproduce the error I was talking about. Using git blame, I think I solved it a while ago in https://github.com/dictation-toolbox/dragonfly/commit/e7392d21b34c8627a99f209d4a995bd5d718ec3b, although I didn't mention it.

In any case this is actually a Caster issue. The `ContextStack` class is adding the history for WSR again here:

https://github.com/dictation-toolbox/Caster/blob/c5a524e597546bdf33e68511b76288df26615a9e/castervoice/lib/dfplus/state/stack.py#L66-L67

This is no longer necessary. There is another problem I've seen though: the `again do` / `again <n> times` commands aren't working.

LexiconCode:
> In any case this is actually a Caster issue. The `ContextStack` class is adding the history for WSR again here:
> 
> https://github.com/dictation-toolbox/Caster/blob/c5a524e597546bdf33e68511b76288df26615a9e/castervoice/lib/dfplus/state/stack.py#L66-L67

I've tested removing the offending code. The record from history command works with both WSR and DNS. I've also verified that the `ContextStack` is still valid according to the unit tests.

LexiconCode:
Can you see if the main pull allows for edit from maintainers? I'm trying to figure out how to submit changes via push pull in git instead of the github UI.

Danesprite:
Well that's good.

'Edits from maintainers is ticked' for me. Is your local repository pointed to `git@github.com:Danesprite/caster.git` or `https://github.com/Danesprite/caster.git` for Push/Fetch URLs? Should be able to test with `git remote show origin`.

It's no big deal if you can't get it working; I should be able to fix this in the next few days, it's just that these sort of changes to Caster take a while to test because of Caster's lengthy load time.

LexiconCode:
Thanks for explaining because this is kind of a pain point when contributing to anyone's pull request. I feel like I'm missing something obvious however here is the output of my attempts.

```
"C:\Program Files\Git\bin\git.exe" push --recurse-submodules=check --progress "Danesprite" refs/heads/Danesprite/remove-wsr-recog-history:refs/heads/Danesprite/remove-wsr-recog-history
Enumerating objects: 125, done.
Counting objects: 100% (111/111), done.
Delta compression using up to 4 threads
Compressing objects: 100% (87/87), done.
Writing objects: 100% (87/87), 19.54 KiB | 1.08 MiB/s, done.
Total 87 (delta 60), reused 0 (delta 0)
remote: Resolving deltas: 100% (60/60), completed with 17 local objects.
To https://github.com/Danesprite/caster.git
 ! [remote rejected]   Danesprite/remove-wsr-recog-history -> Danesprite/remove-wsr-recog-history (permission denied)
error: failed to push some refs to 'https://github.com/Danesprite/caster.git'
Done

```

```
"C:\Program Files\Git\bin\git.exe" push --recurse-submodules=check --progress "Danesprite" refs/heads/Danesprite/remove-wsr-recog-history:refs/heads/Danesprite/remove-wsr-recog-history
fatal: remote error:
  You can't push to git://github.com/Danesprite/caster.git
  Use https://github.com/Danesprite/caster.git
Done
```

I've added your repository as a remote named "Danesprite" as I reserve origin at least in my mind to my repository.
```
git remote show Danesprite
* remote Danesprite
  Fetch URL: git://github.com/Danesprite/caster.git
  Push  URL: git://github.com/Danesprite/caster.git
  HEAD branch: master
  Remote branches:
    gh-pages                 new (next fetch will store in remotes/Danesprite)
    master                   new (next fetch will store in remotes/Danesprite)
    remove-wsr-recog-history tracked
  Local ref configured for 'git push':
    master pushes to master (fast-forwardable)
```

Danesprite:
Hmm okay then. I'm not sure why the `git://` URL isn't working. Maybe try the following commands:

```shell
# Set the remote URL to use HTTPS.
git remote set-url Danesprite https://github.com/Danesprite/caster.git

# Push to Danesprite/remove-wsr-recog-history
git push Danesprite remove-wsr-recog-history

```

If you use SSH, then you use `git@github.com:Danesprite/caster.git` instead of the `https://` URL. You also must have your public key in the GitHub settings: https://github.com/settings/keys

LexiconCode:
The issue I was having was related to the default push repository and branch. A bit confusing as it was expressly defined in the git command so that shouldn't matter. However I now know for the method for making this work at least within Git Extensions.

Danesprite:
Ah, okay then. No worries, I'm glad you got it sorted out :+1:

PR #647, opened 2019-07-14
Testing a test - NOT TO BE MERGED

kendonB:
This pull request is just trying out a basic text test using travis.


kendonB:
The new test did not pass on Travis. Looks like we don't have access to keyboard inputs on Travis. The only way for us to be able to test this functionality would be locally (unless anybody has any bright ideas).

PR #648, opened 2019-07-16
Add back command descriptions and add pause time for "train word" command

alexboche:
These commands need descriptions for users to know what they do. @kendonB suggested this be put in the command documentation. Hopefully I'll get to that eventually but since I'm pretty busy right now and so we don't forget I'm going to temporarily put it inside the file since that's where I already have it. 
Also I don't see much downside to having the documentation inside the file in addition to in the read the docs except the extra work of creating the documentation inside the file which is already done in this case. 

This pull request also fixes a bug with the command "train word" by adding a pause. The purpose of this command is to more quickly train a word/phrase that's inside of full text control environment. The word must be either selected or the cursor must be next to (or inside) the word. For non-full text control environments, use the command "word train".

This is the same branch I initially made the pull request on which was already accepted and merged into dictation toolbox/Caster. Now I made more commits on this. Sorry that's probably not how I'm supposed to do it. I just merged the latest caster commits into this branch.


castervoice/apps/dragon.py
```diff
@@ -71,44 +72,56 @@ class DragonRule(MergeRule):
...
+            R(Mimic("correct", "that") + WaitWindow(title="spelling window") + Mimic("choose", "one"), 
+                rdescript="Dragon: brings up the correction menu for the previously spoken phrase and chooses the first choice"),
+
+        "make that <text>": R(Mimic("scratch", "that") + Mimic(extra="text"), 
+             rdescript="Dragon: deletes the dictation generated by the previous utterance and replaces it with what you say next"),
+        "scratch [<n10>]": R(Playback([(["scratch", "that"], 0.03)]), 
+            rdescript="Dragon: delete dictation from previous n utterances") * Repeat(extra="n10"),
+
+            # Users may want to adjust the wait time on the next few commands     
+        "train word": R(Mimic("train", "that") + Pause("75") + Key("a-r/250, s"),
```
LexiconCode: I don't think "train that" is a valid DNS command as it prints "train that".

LexiconCode:
> This is the same branch I initially made the pull request on which was already accepted and merged into dictation toolbox/Caster. Now I made more commits on this. Sorry that's probably not how I'm supposed to do it. I just merged the latest caster commits into this branch.

I keep branches with the same name for frequently worked on items. An example being "Documentation". I'm not sure what standard practices but it's not an issue from a maintainer perspective.

castervoice/apps/dragon.py
alexboche: You have to be in a full text control application for the command "train that" to work

PR #663, opened 2019-07-26
Use testrunner for Travis

comodoro:
Playing with Travis, I think a better way is to 1.use testrunner, since it initializes the text engine, which I do not know how to do via commandline with test discovery, and 2. fail the script commands early


LexiconCode:
This may or may not be helpful documentation with dragonfly's [CLI](https://dragonfly2.readthedocs.io/en/latest/cli.html)

PR #667, opened 2019-08-01
Grammar documentation window

comodoro:
If grammar documentation cannot be extracted from dragonfly ( only already merged ), the only other way I can think of doing it is compiling other sources, like existing modules and user settings. Like for example here, moving aside esthetics and technical details like rdescripts. This is just a draft. Thoughts?

### DIFF
--- a/_caster.py
+++ b/_caster.py
@@ -148,6 +148,9 @@ class MainRule(MergeRule):
 
 control.non_ccr_app_rule(MainRule(), context=None, rdp=False)
 
+from castervoice.lib.rules.doc_rule import DocumentationRule
+control.non_ccr_app_rule(DocumentationRule(), context=None, rdp=False)
+
 if globals().has_key('profile_switch_occurred'):
     reload(sikuli)
 else:
--- a/_caster.py
+++ b/_caster.py
@@ -4,17 +4,31 @@ main Caster module
 Created on Jun 29, 2014
 '''
 
-import os, time, sys
+import datetime
+start_time = datetime.datetime.now()
+
+import os, time, sys, subprocess
+from castervoice.lib import settings
+from castervoice.lib import message_window
+subprocess.Popen([settings.SETTINGS["paths"]["PYTHONW"], message_window.__file__])
+from socket import socket, AF_INET, SOCK_STREAM
+sock = socket(AF_INET, SOCK_STREAM)
+sock.connect(('localhost', 10011))
+file = sock.makefile('w', 0)
+sys.stdout = file
+sys.stderr = file
+
 import logging
 logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO )
+logger.info("Starting Caster")
 
-import time, socket, os
 from dragonfly import (get_engine, Function, Grammar, Playback, Dictation, Choice, Pause,
                        RunCommand)
 from castervoice.lib.ccr.standard import SymbolSpecs
 
 _NEXUS = None
-from castervoice.lib import settings  # requires nothing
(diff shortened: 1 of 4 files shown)


LexiconCode:
First, I've also been working on planning out a GUI. Noting this limitation I've worked with @synkarius with the CCR the merger rewrite. One of the things he's implemented this event hook to grab information during the merge process about Grammars and rules before merger. This allows us to grab information from the engine rather than imports which includes changes from Filter Rules soon to be called Transformers.

Second I was actually hoping to use [PySimpleGUI27](https://pypi.org/project/PySimpleGUI/) instead of native TK. PySimpleGUI utilizes a number of frameworks one of which is tkinter.

PySimpleGUI offers a number of advantages if it's flexible enough for our purposes. 
- Painless transition from Python 2.7.x to 3.x 
- Painless transition to different frameworks tkinter to Qt or WxPython.
- Simplified abstraction of the frameworks making it much easier to extend the GUI.
- PySimpleGUI is actively developed and maintained.
- A wonderful cookbook/coding examples to demonstrate features of the UI.

Not that we have to use PySimpleGUI but if it's of interest to you let me know what you think. It would be nice to use one framework as we move forward.

comodoro:
That's good to hear. Information about grammars being tracked before merging is certainly superior, means more changes, so good that it's already being done. By the way, a thought just occurred to me, if these hooks were non-intrusively available, like a method to add a custom callback function, I think translations could be possible this way. Not that I have a Czech engine yet:(

PySimpleGUI looks interesting. I will check it out and maybe reimplement a simple logging window I have somewhere among branches. Which is not much, but just to get the look and feel.

synkarius:
@comodoro "A method to add a custom callback function" is basically what these event hooks are going to be. Information about, for example, merges and NodeRule node changes, will be available for you to do whatever you want with. The specifics of what information exactly is still TBD.

comodoro:
Just added a PySimpleGUI window, which is even rougher POC (does not handle sockets timeouts, Dragon restarts etc.). My reasoning for using a socket is that standard output can be really frequent and using something like XML-RPC might start to have significant overhead(?).

PySimpleGUI is a really good for simple things done quickly, but has quite a few disadvantages:

- The ability to support several frameworks with a single interface means that you cannot use advanced features of any of the frameworks. With it being relatively new it means that some features are lacking, even for Casters usage. For example, I do not really know how to make advanced text formatting inside a text widget

- It is basically a one-man show at the moment, if you look at the GitHub repository. The person is really dedicated and sympathetic, but if there are not more developers, one might wonder about the stability of its future.

Seeing recent developments in the general Gitter channel, if we were to eliminate Natlink (sorry, Quintijn), we would be able to eliminate this socket RPC business, which is sometimes very cumbersome and some behaviors are hard to eliminate, like if you exit Dragon while some Caster GUI windows are still visible, they are not automatically closed.

comodoro:
Forgot to mention, needs `pip install PySimpleGUI27`

LexiconCode:
All right have gotten a chance to try this out!
I'm going to try to if there's a way so that the window contents is not editable.

LexiconCode:
@comodoro It sounds like tkinter would be a better framework overall. So you think we should stick with it I'll learn to adapt.

comodoro:
Yes, it is not too complicated too and supposedly should be capable of more stuff than the defaults people use. I might have a look at wxPython in the coming weeks as I should have more time, as it is already present, another option is PyQT, although probably unnecessarily complex for Caster

LexiconCode:
Looking at this it would be pretty easy to get it working with the rewrite so I will leave it open.

LexiconCode:
Closing as some other GUI stuff is in the works. Thanks for the inspiration though!

PR #669, opened 2019-08-02
consolidate double punctuation, hug, and formatting e.g. snake

alexboche:
## Description

This pull request allows you to say things like `"quotes <dictation>"` and puts the cursor at the end with a space. Other spacing options (e.g. before and/or after) are provided. The commands are in nav.py
It also combines this with the hug commands (e.g. for in enclosing something that is already there) and the formatting commands like snake (e.g. "hug quotes snake") . I removed the dictionary called `double_text_punc_dict` in favor of a more flexible dictionary called `enclosure` In punctuation.py.  the values of `enclosure` are either tuples representing the left and right boundaries of the enclosure or a single string which will occur on both the left and right side of the enclosure. I also add the ability to add spaces before and/or after dictation (enclosed and/or formatted or otherwise). Because functionality is consolidated, some old code is deleted. 

Note: it appears that the ability to say "cap" to get capitalization is now built into dragonfly `Dictation` objects?
I modified the capitalization/spacing commands a bit it so that you can capitalize words by prefixing with "cap"
This works with all the regular dictation commands but if you want to get that when using a spacing command like "snake" you must
preface "snake" with the word "rarb". ideally you wouldn't have to preface it that way but it's a tricky to 
get that without making considerable changes to the original capitalization/spacing functions.

I try to optimize the spacing defaults so for example a lot of things have a space afterwards but not the hug commands
and if you just say something like quotes, the cursor will be inside with no space. 
But the spacing is not perfect in every case. For example, saying something like "tickris angle hello" produces `"<hello> |"` (where the quotes are supposed to be backticks I just don't know how to escape them)
which is probably not what you want (where the bar represents the cursor). if you use, "tickris nurk angle hello", `"<hello>|"` is a bit better but you still want the cursor outside.
To solve this problem the user can create a new enclosure for this as I have already done and called it "varib" which produces the desired result `"<hello>" |` (again, I am using `"` instead of backticks because the escaping is tricky, but "variab" prints out backticks around angle brackets). 

Since all these commands basiccally run through the same function `enclose_format` in text_format.py, I made them all be printed out with the same method, namely text action.This is a change and that previously the hug commands used a clipboardd paste. There is a variable in the function `enclose_format` for whether a hug command is being used or not,so if desired it can be changed back to use the clipboard paste for hug commands.

It is possible that having all these different dictation commands or even just a couple with a bunch of different variables and optional arguments may slow down the recognition of these commands (but I don't think that's too much of an issue). 

## Motivation and Context
Without this you have to say things like `quotes format <dictation> ace arch`. now you can just say things like `quotes <dictation> arch`.   

## Possible future features
Ideally we would be able to have the ability to say something like "quotes arch brov long equals arch"
and have the cursor come out on the right side of the quotes without saying ross. 
This requires some kind of command recursion scheme which is something that caster probably should think more about anyways.
This can be accomplished using the underlying dragonfly API using Repetition objects, though I don't have a good understanding of it.
See the approach taken by @mrob95 in Mathfly for example (though there may be better ways)
[code](https://github.com/mrob95/mathfly/blob/master/mathfly/lib/merge/nestedrule.py) ,  [description](https://mathfly.org/implementation.html)
I might open issue about this at some point.

## How Has This Been Tested
I tested the commands and punctuation.py to make sure there is still work.
The new commands are all in nav.py . There are commands that just put a pair of punctuation marks `<enclosure>` 
and then there are commands that do dictation possibly with enclosures in her capitalization and/or spacing formatting
and/or outer spacing adjustment (e.g. before and/or after). I have tested various combinations of these.
My testing is in VS code if I see no bugs currently (though I may well have missed something).
Other editors may pose issues due to the auto pairing of punctuation. In writing these commands,
I did not take account of this auto pairing.  

### Bugs
I have found a bug when testing "tickris" in sublime
where unwanted single quotes are added when typing backticks. I don't know what to do about that. Within get markdown on Google Chrome, I once got the behavior where saying "prekris" but the cursor on the right of the parentheses instead of inside, but then that stopped happening I don't know why. 

## Types of changes

- [ x] Docs change recommended but not done
- [ x] New feature (non-breaking change which adds functionality)
- [x ] Breaking change: fix or feature that __could__ (but doesn't unless I made a mistake) cause existing functionality to change

## Checklist

- [x ] I have read the CONTRIBUTING document.
- [ x] My code follows the code style of this project.
- [ x] I have checked that my code does not duplicate functionality elsewhere in Caster.
- [ x] I have checked for and utilized existing command phrases from within Caster (delete if not applicable). 
- [ x] My code implements all the features I wish to merge in this pull request.
- [ x] My change requires a change to the documentation.

### DIFF
--- a/castervoice/apps/dragon.py
+++ b/castervoice/apps/dragon.py
@@ -12,16 +12,6 @@ def fix_dragon_double(nexus):
         utilities.simple_log(False)
 
 
-def cap_dictation(dictation):
-    input_list = str(dictation).split(" ")
-    output_list = []
-    for i in range(len(input_list)):
-        if input_list[i] == "cap":
-            input_list[i + 1] = input_list[i + 1].title()
-        else:
-            output_list.append(input_list[i])
-    Text(" ".join(output_list)).execute()
-
 
 # extras are common to both classes in this file
 extras_for_whole_file = [
@@ -47,7 +37,6 @@ class DragonRule(MergeRule):
     pronunciation = "dragon"
 
     mapping = {
-        "format <text>": Function(cap_dictation, extra={"text"}),
         '(lock Dragon | deactivate)':
             R(Playback([(["go", "to", "sleep"], 0.0)])),
         '(number|numbers) mode':
--- a/castervoice/lib/ccr/core/nav.py
+++ b/castervoice/lib/ccr/core/nav.py
@@ -232,22 +232,24 @@ class Navigation(MergeRule):
         "peek [<big>] format":
             R(Function(textformat.peek_text_format)),
        
-        "[<big>] format <textnv>":
+  
+        # the next three commands put a space at the end
+        "([<enclosure>] (<capitalization> <spacing> | <capitalization> | <spacing>) | <enclosure> [<capitalization>] [<spacing>]) [bow] <textnv> [(brunt | over)]": 
+            R(Function(enclose_format)), # combination enclosure, spacing, and capitalization command
+        "spay ([<enclosure>] (<capitalization> <spacing> | <capitalization> | <spacing>) | <enclosure> [<capitalization>] [<spacing>]) [bow] <textnv> [(brunt | over)]": 
+            R(Text(" ") + Function(enclose_format)), # same as above, puts space at the beginning
+        "phrase <textnv> [(brunt | over)]": R(Function(enclose_format)), # mid utterance dictation command like "format"
+        # the next four commands do not put a space at the end
+        "nay ([<enclosure>] (<capitalization> <spacing> | <capitalization> | <spacing>) | <enclosure> [<capitalization>] [<spacing>]) [bow] <textnv> [(brunt | over)]": 
+            R(Function(enclose_format) + Key("backspace")), # combination enclosure, spacing, and capitalization command  
+        "<enclosure>": R(Function(enclose_format)), # this is to e.g. type a pair of parentheses and put the cursor in between them
+        "[<big>] format <textnv> [(brunt | over)]":
             R(Function(textformat.prior_text_format)),
-        "<word_limit> [<big>] format <textnv>":
+        "<word_limit> [<big>] format <textnv> [(brunt | over)]":
             R(Function(textformat.partial_format_text)),
         "hug <enclosure> [<capitalization>] [<spacing>]":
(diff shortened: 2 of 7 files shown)


kendonB:
Hi Alex, are you able to follow the template for this PR?

alexboche:
yes I will follow the template later today.

alexboche:
@kendonB Thank you for pointing me to the template. This is ready for your review. Thank you.

kendonB:
Does the template not automatically come up for you?

alexboche:
It does, I just didn't use it the first time. sorry.

kendonB:
I'm not really sure about this new behavior. I see the utility of adding spaces when you do and moving outside the enclosures automatically in cases where that's what I want to do. However, it presumes that users _usually_ want that behavior and I'm not sure that they do. Even just dictating this paragraph, I've had to reverse several spaces that the new behavior puts in. There are plenty of use cases where you want to stay inside enclosures after an utterance and I'd argue that those use cases are pretty damn common in programming.

It's also not a massive problem that this PR solves and it adds new commands adding complexity. It's not that hard to say "ross ace" when you want to exit enclosures.

So, not a rejection but you'll have to make a bit more of a case for moving outside of enclosures and adding spaces being the default behavior.

You should also split this into several PRs as lots of the changes don't need to be together. The new enclosures you're proposing should be a single PR, for example.

A few minor points:
 - The "spay" command seems redundant when you can already say "ace" before these commands.
 - It's worth having a engine-formatting option in Caster and I like using "phrase" not "rarb"
 - The PR has too many changes in behavior to easily review without clear and complete instructions for testing. e.g. why do have "rarb" as well as "phrase" which seem to have the same behavior?? What're the different use cases that each solves?

alexboche:
Thanks for reviewing this. maybe after we have  instant reloadability I will revisit this. I  realize now that I should use things for at least a week on my own system before doing a pull request, though it's a bit tricky because I haven't pulled from caster in over six months  so things are a bit different  on my version and also I'm not using caster that much anyways at the moment.   "rarb"  is for e.g. if you want to say things like "snake hello Alex"  and have it still capitalize Alex  with output hello_Alex.

LexiconCode:

Issue #625, opened 2019-07-02
Overhaul grammars to use plain English based on Human-Machine Grammar.


LexiconCode:
Once we've evolved filter rules and grammar loading from user were it's much easier to include command profiles This issue be pursued. Profiles such as such as legacy caster, a narrow scope like the alphabet NATO phonetic or even another language such as Dutch.

As it stands currently Caster has relied on phonetically distinct and obscure words in its grammars. Traditionally this is led to higher accuracy of command recognition. However that is a little less relevant than it was in 2015 when the majority of Caster was written. 

Drawbacks of the current grammar set

- Foremost its simply It's harder to learn these obscure words and increases cognitive loads.
- It's hard to guide users to standardize a method for adding new commands that utilize obscure words somewhat less phonetically distinct. 
- It's difficult to write documentation because words like "sauce" means "up" to be explained.
-  It would be difficult to translate into another language

**The proposal**

- Utilize plain English in all casters grammar

- Let end-user customize their grammars through the user directory when the need arises for phonetically distinct words as its highly individual based on accident and preference

- Adopt [Human-Machine Grammar - The Rules](http://redstartsystems.com/human-machine-grammar-dictionary) for developing the new plain English grammars and as a template for future grammars.

- Discuss the avenue to retire to legacy profile or drop the current grammar.

- Pursue permissions from Kim the author to utilize [Human-Machine Grammar Dictionary](http://redstartsystems.com/human-machine-grammar-dictionary) as a basis for redeveloping caster Grammars. - A third-party [review](http://handsfreecoding.org/2018/09/04/utter-command-why-i-rewrote-my-entire-grammar/) of grammar system

LexiconCode:
> I am probably behind, could you please explain what you mean? I don't have any command `up`,Is there a context to it?

No you're not behind I forgot I have some filter rules running `sauce -> up` when dictating.  

> My concern is simply that if you want to automatically document grammars, you have to have something that describes them. Almost all current specs, `sauce` or `dunce`, for example, are in themselves just meaningless words.

That's a valid point. The majority of obscure words are contained in core less so in the app grammars. As a temporary measure I propose that we add back some rdescripts. The guiding rule being simplifying complex commands as demonstrated above or explaining ambiguity. Overall were still left with the benefit of not having that document self-explanatory commands in the code as rdescripts. However there's any other thoughts on how to mitigate these issues feel free to broaden my horizons.

[...]

PR #670, opened 2019-08-03
Key commands tweaks

alexboche:
(I consider this ready for review now.)
## Description

* This adds the ability to say shin/que dunce/sauce wally to press control shift end/home. shin/que do exactly the same thing as each other in this case. Technically this is unnecessary, because you can already say que lease/ross wally to get this but it is nice cognitively to have this option.
* I also change latch-> lairch (home) ratch->rairch (end) because of a pattern of misrecognition. 
Edit: 
* I also removed four commands blitch/flitch/brick/frick which I previously added without proper testing, two of which caused misrecognitions (brick/frick).
* I removed unnecessary repeats after the commands "sauce wally": R(Key("c-home")),
        "dunce wally": R(Key("c-end")),
        "(shin | que) sauce wally": R(Key("cs-home")),
        "(shin | que) dunce wally": R(Key("cs-end")),
though kept the repeat after [shin] lease/ross wally because that can be used in Lyx and less frequently in VS code.
 Newly added/removed commands have not yet been documented and (I assume) have not been adopted by other users so there should be no harm done. 

## Related Issue
Regarding the names of commands, #625 is related.

## Motivation and Context
Shorter commands are useful instead of Ross/lease wally and that's what lairch/rairch are for. Ideally I think most command should have a version that's easy to remember and a shorter version (usually one syllable). A while back, I commented out the original master command for lease/ross/que/shin/fly/wally and replaced it with individual commands attempting to preserve the original commands as well. I am not sure if shin sauce/dunce wally was part of that or not since they are somewhat redundant but they are included there now in any case. 

## How Has This Been Tested
I have been using the commands lairch/rairch (as well as the previously added bird/firch) heavily for several days with no misrecognitions. I was using these when testing the commands from #669 and did not notice any misrecognitions. I only tested the shin/que sauce/dunce wally a couple times. 

## Types of changes

- [ ] (possibly) documentation change

## Checklist

- [x] My change requires a change to the documentation (but maybe not yet).


kendonB:
Hi Alex, are you able to use the PR template? My view is that all PRs should use it except for trivial changes. For example, a typo fix or a single line change is trivial but something that needs more than a sentence to explain is not trivial and should use the template.

mrob95:
```
        "(lease wally | lairch) [<nnavi10>]": R(Key("home:%(nnavi10)s")),
        "(ross wally | rairch) [<nnavi10>]": R(Key("end:%(nnavi10)s")),
        "sauce wally [<nnavi10>]": R(Key("c-home:%(nnavi10)s")),
        "dunce wally [<nnavi10>]": R(Key("c-end:%(nnavi10)s")),
        "(shin | que) sauce wally [<nnavi10>]": R(Key("cs-home:%(nnavi10)s")),
        "(shin | que) dunce wally [<nnavi10>]": R(Key("cs-end:%(nnavi10)s")),
        "bird [<nnavi500>]": R(Key("c-left:%(nnavi500)s")),
        "firch [<nnavi500>]": R(Key("c-right:%(nnavi500)s")),
        "brick [<nnavi500>]": R(Key("s-left:%(nnavi500)s")),
        "frick [<nnavi500>]": R(Key("s-right:%(nnavi500)s")),
        "blitch [<nnavi500>]": R(Key("cs-left:%(nnavi500)s")),
        "flitch [<nnavi500>]": R(Key("cs-right:%(nnavi500)s"))
```

A few points.

1. There is no need for repeat values on "wally" commands, they are idempotent.
2. Is any of this stuff documented anywhere?
3. Don't these cause all sorts of misrecognitions, particularly combined with others like #669? (I'm not using caster master so haven't tested, but the fact that they keep needing to be tweaked suggests that this is a problem)

This may be a discussion better had in #625, but I think it needs to be had. I'm not sure it's fair or reasonable to expect new users (who may have poor quality audio setups or be using less accurate backends) to learn a whole load of made up words which all sound similar and aren't documented.

alexboche:
The "wally" are not always itempotent (i.e. repeating them sometimes does do something different than just doing them once) for example in Lyx. These are not documented yet, because there still kind of experimental, but it is nice to build a say things like "shin lairch" instead of "shin lease wally" since its shorter. Yeah maybe a bunch of new shorter commands should all be added at the same time and documented rather than a few at a time like I've been doing. I have had success with these without misrecognitions so far but I've only tested for a few days.

LexiconCode:
I think we need to be careful about adding new commands that have made of words so perhaps we should have a discussion about this issue in #625. Overwhelming amounts of made up words can be a deterrent for adoption as well.

Then we can address this issue after the CCR rewrite? BTW which is very close to being In a testable usable state.

LexiconCode:

PR #372, opened 2019-01-18
Implements tomlkit in place of toml dependency

LexiconCode:
A work in progress is not ready for testing.

Fixes  non-ASCII characters in clipboard
https://github.com/dictation-toolbox/caster/issues/357 

Implements sorting keys
https://github.com/dictation-toolbox/caster/issues/323

Issue #323, opened 2018-11-14
Implement Sort keys when saving Toml files.


LexiconCode:
Json has `sort_keys=True`  That does not exist within toml specifications.
An equivalent method needs to be implemented for toml files.

Issue #244, opened 2018-07-02
Alias Commands are nonfunctional


LexiconCode:
The command `alias <something>`seems to be broken. Supposedly you can create commands on-the-fly by highlighting text on the screen and say  `alias <something>`. [Alias Documentation](http://caster.readthedocs.io/en/latest/caster/doc/readthedocs/Alias/)

Nothing was added to `aliases.json` after attempting to  `alias <something>`.

gerrish:
Do you get any errors from natlink?

LexiconCode:
Excellent! For reference  https://github.com/synkarius/caster/pull/245

gerrish:
Thank you I will do so in the future

PR #676, opened 2019-08-15
Implements tomlkit in place of toml dependency

LexiconCode:
## Description

Implements [tomlkit](https://github.com/sdispater/tomlkit) place of [uiri/toml](https://github.com/uiri/toml) dependency.

A lot of credit goes to @comodoro for testing and breathing some life into the old issue.

Contains temporary a patch in place of https://github.com/dictation-toolbox/dragonfly/pull/125.

## Related Issue

Original Pull Request https://github.com/dictation-toolbox/Caster/pull/372

Fixes non-ASCII characters in clipboard 
https://github.com/dictation-toolbox/caster/issues/357 

Implements sorting keys
https://github.com/dictation-toolbox/caster/issues/323

## Motivation and Context

[uiri/toml](https://github.com/uiri/toml) is not being maintained enough for the purposes of our project.
`
String dump with correct escaping #244` pull request by **comodoro** has been open for 111 days without comment from the author. Thus the motivation to move to an alternative toml parsing library. Along with the related issues mentioned above.
 
## How Has This Been Tested

- Deleting and regenerating all toml files
- Pasted built-in unit tests
- creating/using aliases
- launching bring me commands
 

## Types of changes

- [x] New feature (non-breaking change which adds functionality) (sorting keys)

## Checklist

### DIFF
--- a/castervoice/lib/dfplus/merge/ccrmerger.py
+++ b/castervoice/lib/dfplus/merge/ccrmerger.py
@@ -421,7 +421,7 @@ class CCRMerger(object):
         ORIGINAL, SEQ, TERMINAL = "original", "caster_base_sequence", "terminal"
         alts = [RuleRef(rule=rule)]  #+[RuleRef(rule=sm) for sm in selfmod]
         single_action = Alternative(alts)
-        max = settings.SETTINGS["miscellaneous"]["max_ccr_repetitions"]
+        max = int(settings.SETTINGS["miscellaneous"]["max_ccr_repetitions"])
         sequence = Repetition(single_action, min=1, max=max, name=SEQ)
         original = Alternative(alts, name=ORIGINAL)
         terminal = Alternative(alts, name=TERMINAL)
--- a/castervoice/lib/github_automation.py
+++ b/castervoice/lib/github_automation.py
@@ -7,7 +7,7 @@ import sys
 import traceback
 from __builtin__ import True
 from subprocess import Popen, PIPE
-import toml
+import tomlkit
 import time
 
 from castervoice.lib.actions import Key, Text
--- a/castervoice/lib/settings.py
+++ b/castervoice/lib/settings.py
@@ -153,7 +153,7 @@ def _validate_engine_path():
         return ''
     if os.path.isfile(_SETTINGS_PATH):
         with io.open(_SETTINGS_PATH, "rt", encoding="utf-8") as toml_file:
-            data = tomlkit.loads(toml_file.read())
+            data = tomlkit.loads(toml_file.read()).value
             engine_path = data["paths"]["ENGINE_PATH"]
             if os.path.isfile(engine_path):
                 return engine_path
@@ -429,7 +429,7 @@ def _init(path):
     result = {}
     try:
         with io.open(path, "rt", encoding="utf-8") as f:
-            result = tomlkit.loads(f.read())
+            result = tomlkit.loads(f.read()).value
     except ValueError as e:
         print("\n\n" + repr(e) + " while loading settings file: " + path + "\n\n")
         print(sys.exc_info())
@@ -445,7 +445,7 @@ def _init(path):
         new_path = old_clipboard[:-4] + "json"
         print("\n\n Migrating clipboard from {} to {}".format(old_clipboard, new_path))
         with io.open(old_clipboard, "rt", encoding="utf-8") as f:
-            clipboard = tomlkit.loads(f.read())
+            clipboard = tomlkit.loads(f.read()).value
         formatted_data = unicode(json.dumps(clipboard, ensure_ascii=False))
         with io.open(new_path, "wt", encoding="utf-8") as f:
             f.write(formatted_data)
--- a/castervoice/lib/settings.py
+++ b/castervoice/lib/settings.py
@@ -6,7 +6,7 @@ import collections
 import io
 import os
 import sys
-import toml
+import tomlkit
 import _winreg
 import version
 import errno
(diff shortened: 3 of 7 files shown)


LexiconCode:
One would expect `dep_missing()` Tot inform the user of tomlkit missing dependency with instructions when users upgrade to the latest master branch.

However since `dependencies.py` utilizes `castervoice.lib import settings` tomlkit tries to load before `dep_missing()`can execute. Disappointing that this isn't covered but `tomlkit` is the only third-party module to my knowledge that loads before `dep_missing()` runs.

LexiconCode:
@comodoro one thought would be to add some unit tests to writing/reading to toml files.

comodoro:
Well, you can maybe read requirements (isn't setup.py better as more lenient?) without settings, using `__file__`, but it would surely need reorganization of dependencies.py. Speaking of which, I can see some problems in the file, like line 21 (is always true), line 46 (sets default timeout, not just for one socket), line 104 (hardcoded dependency, different from setup.py). And it wants to update despite me having latest master. Updates to some 0.6.13 - when current master has version.__version__ 0.6.11. Is there some other branch on pypi?

LexiconCode:
> Well, you can maybe read requirements (isn't setup.py better as more lenient?) without settings, using `__file__`

That's a thought and should be possible.

> Speaking of which, I can see some problems in the file, like line 21 (is always true)

`    if sys.platform == "linux" or "linux2":` line 21 is always true?

>  line 46 (sets default timeout, not just for one socket

good catch `socket.setdefaulttimeout(timeout)` should be `socket.settimeout(value)`

> line 104 (hardcoded dependency, different from setup.py)

This is intentional `["dragonfly2", ">=", "0.13.0", None],` is the bare minimum to run Caster. Using anything lower will crash Caster on startup. This is used not only as an example but for those who choose not to utilize `setup.py`

LexiconCode:
> Updates to some 0.6.13 - when current master has version.**version** 0.6.11. Is there some other branch on pypi?

Yes the versions are out of sync. You have to bump the package version each to update pypi. There were few gotchas running the PIP install next to the Classic install. Imports will always come from the PIP install even when you're running the classic install. Except for testing purposes I recommend uninstalling PIP if you have the classic install.

The versions will be put back in sync once we have a method for the automating the changelog.

comodoro:
It seems that `x == 'a' or 'b'` is True if the condition in True else 'b'. I thought in the latter case it would be bool('b'), i.e. True. Anyway, you want either sys.platform.startwith("linux"), or ` sys.platform == "linux" or sys.platform == "linux2"`

I thought there was some Context thing later, was it 0.14.1? Anyway, maintaining yet another dependency setting makes things harder.

As for the pip package, I suppose you can just increase the version in the file for each release, or I suppose that's the way they usually do it. Otherwise it seemed to work, are there more problems?

LexiconCode:
@comodoro 
> It seems that `x == 'a' or 'b'` is True if the condition in True else 'b'. I thought in the latter case it would be bool('b'), i.e. True. Anyway, you want either sys.platform.startwith("linux"), or ` sys.platform == "linux" or sys.platform == "linux2"`

Will do

> I thought there was some Context thing later, was it 0.14.1? Anyway, maintaining yet another dependency setting makes things harder.

True enough. However how should we handle the case would've had dependency is changed to something version specific between releases?

The crux of the issue is for the classic install setup.py isn't used to manage dependencies beyond the initial install. The alternative is to simply state that there's been a version dependency change within chat however that's cumbersome and gets buried quickly. This not only handles notifying users by giving them instructions automatically. Although dragonfly minimum version is a rare case and should of been updated to 0.14.1. I might just comment out that line for an example. 

If you have other thoughts on how to handle this without the function altogether by all means let me know.

> As for the pip package, I suppose you can just increase the version in the file for each release, or I suppose that's the way they usually do it. Otherwise it seemed to work, are there more problems?

That's the plan I just want to be able to show the changes between each release. For now I could do a version bump without explanation and a change log. Otherwise everything else seems to check out.

LexiconCode:
@comodoro what you think?

```Python
def internet_check(host="1.1.1.1", port=53, timeout=3):
    """
    Checks for network connection via DNS resolution.
    :param host: CloudFire DNS
    :param port: 53/tcp
    :param timeout: An integer
    :return: True or False
    """
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    try:
        s.settimeout(timeout)
        s.connect((host, port))
        return True
    except Exception:
        return False
```

comodoro:
I would consider it cleaner with a more specific exception, not sure which, but it is probably fine.

No idea about the dependency issue, what I would probably do is limit dependency checking in non-pip install to what can be reasonably simply coded, in this context the line does not look so bad, else OTOH another file with this kind of more relaxed dependencies could be parsed by both setup.py and dependency checking

LexiconCode:
> I would consider it cleaner with a more specific exception, not sure which, but it is probably fine.

Done https://github.com/dictation-toolbox/Caster/commit/f3781bb5be881e11ee226b5a67e77656625908a6

LexiconCode:
@comodoro 
[Reimplemented](https://github.com/dictation-toolbox/Caster/compare/master...LexiconCode:settings_dependency_removal) dep_missing without caster settings dependency 

The last tricky bit. Thoughts on how to handle this?
https://github.com/dictation-toolbox/Caster/blob/f3781bb5be881e11ee226b5a67e77656625908a6/castervoice/lib/ctrl/dependencies.py#L143

comodoro:
That's tricky. If you want requirements checking at the very beginning and use this setting for update checking, I suppose you have to separate the two and use the requirements checking before loading settings. Sorry, I did not understand the last line.

LexiconCode:
> That's tricky. If you want requirements checking at the very beginning and use this setting for update checking, I suppose you have to separate the two and use the requirements checking before loading settings.

All right have get something worked out for that.

> Sorry, I did not understand the last line.
No problem, Your feedback has been very helpful!

>  Disappointing that this isn't covered but `tomlkit` is the only third-party module to my knowledge that loads before `dep_missing()` runs.

Unfortunately this assumption is wrong. 'settings', `utilities` and `control.nexus` load before `dep.initialize()`. hmm

LexiconCode:
@comodoro 
Let me know what you think of this DependencyMan loads before first and has been removed from the Nexus.

https://github.com/dictation-toolbox/Caster/compare/master...LexiconCode:settings_dependency_removal

PR #687, opened 2019-10-07
Fix copy line down command

dusty-phillips:
# trivial change

Looks like a copy paste error.


LexiconCode:
Good catch! 
The ccr rewrite contains a significant change to almost all grammars that would invalidate any current open pull requests. Once the ccr rewrite is merged this pull request can be resubmitted and merged.

I'll keep this open until that time comes and follow-up.

LexiconCode:
Do the complexity of resolving conflicts I went ahead and included your fix in a push to the main repository. Thanks again @dusty-phillips 
https://github.com/dictation-toolbox/Caster/commit/7eb7d8eeea3785aca78f89bda7fca86b8a30ca0f

PR #691, opened 2019-10-11
Add discussion of bug to text manipulation commands buglist

alexboche:
I am adding a discussion of a bug where it's misinterpreting character names as dictation.


kendonB:
Are you sure this bug is related to the `text_manipulation` grammar?

LexiconCode:
I suspect it could be mitigated altogether for people to be using `command mode` so that free dictation does not interfere.

PR #699, opened 2019-11-13
Clarify Aenea setup instructions

ericricky:
I added one step to the Aenea setup instructions to let people know that they will have to copy `aenea.json` to their Caster install directory. I also added the location of `settings.toml` to the next step.


LexiconCode:
Good catch! This will merge when the CCRrewrite. Thanks!

ericricky:
@LexiconCode No problem!

PR #703, opened 2019-12-04
Travis update test runner path and utilize dragonfly test engine

LexiconCode:
Test runner script path change


LexiconCode:
the path is corrected. However minimal aspects of the unit tests require an engine.

PR #706, opened 2019-12-06
Fix AppContext for Aenea in Mapping Rule Maker

LexiconCode:
## Description

Fixed AppContext for Aenea
See details at `types of changes` section

## Related Issue

https://github.com/mpourmpoulis/PythonVoiceCodingPlugin/pull/2

## Motivation and Context

Correct `AppContext` so grammars are proxied with Aenea enabled

## How Has This Been Tested

Needs testing from Aenea Caster Users

## Types of changes

- Added to mapping_rule_maker.py which should support any loaded rule by default.  Exception if AppContext is used within the grammar such as `ContextAction`

- Clarified `AppContext` import use in  Aenea documentation  with `ContextAction`.
- Removed unnecessary AppContext from applications
- Added Aenea aware  AppContext: nav.py, text_manipulation_support.py, bringme.py, text_manipulation_support.py, html_rule.py, nav.py

## Checklist


LexiconCode:
A report
```
File "C:\Python27\lib\site-packages\dragonfly\engines\backend_natlink\engine.py", line 285, in begin_callback
self.grammar.process_begin(executable, title, handle)
File "C:\Python27\lib\site-packages\dragonfly\grammar\grammar_base.py", line 444, in process_begin
or self._context.matches(executable, title, handle):
File "C:\NatLink\NatLink\MacroSystem\aenea\proxy_contexts.py", line 146, in matches
return self._reduce_matches(self._check_properties())
File "C:\NatLink\NatLink\MacroSystem\aenea\proxy_contexts.py", line 119, in _check_properties
self.arguments[key])
File "C:\NatLink\NatLink\MacroSystem\aenea\proxy_contexts.py", line 127, in _property_match
desired = desired.lower()
AttributeError: 'NoneType' object has no attribute 'lower'
```
Which is mitigated by commenting out `from castervoice.lib.context import AppContext` in `mapping_rule_maker.py`. This need to be reproduced on another `Aenea` set up according to the contributor.

mpourmpoulis:
ok i think i found the root of the problem 
in mapping rule maker there is this line: 

```
if details.executable is not None or details.title is not None:
            context = AppContext(executable=details.executable, title=details.title)
```
  
allows to pass executable = None! 

Then we get in the aenea proxy_context

```
class ProxyCustomAppContext(dragonfly.Context):
    def __init__(self, match='substring', logic='and', case_sensitive=False,
                 query=None, **kw):
        if query is None:
            query = {}
        query.update(kw)
        self.arguments = query
```


and then in 

```
def _check_properties(self):
```

in  the loop over the self.arguments items we enter the case

```
elif key in properties:
                matches[key] = self._property_match(key, properties[key],
                                                    self.arguments[key])
```

when key is "executable", and the last parameter will be none! 

This is the "desired"parameter that appears in the log! and inside that function call, calling lower in the None object causes a crash!

mpourmpoulis:
I have tried to modify These line into

```python
        if details.executable is not None or details.title is not None:
            context = AppContext(executable=details.executable if details.executable else "dummy", 
                title=details.title if details.title else "dummy")
```
and everything seems to be working Fine, I verified against both my plug-in and that The default sublime rule it is taking into account the VM context And no exceptions are thrown! but I guess a neater solution is needed

LexiconCode:
Well done! I'm grateful, As this is been really tricky to try to debug without working Aenea setup. It passes casters unit tests! I will follow up with some manual testing.

mpourmpoulis:
Glad to be of assistance:)

LexiconCode:
This current issue takes care of MappingRules and it looks like MergeRules should be okay but let's not leave that up to chance.

@mpourmpoulis 
Here's the equivalent for creating AppContext for ccr based rules like MergeRules.   
https://github.com/dictation-toolbox/Caster/blob/0a0d2a41e92ec12a33f0affaea1a1fdc4c0f202e/castervoice/lib/merge/ccrmerging2/ccrmerger2.py#L136
It imports `from castervoice.lib.context import AppContext` already

Here's your grammar made into a App_CCR_MergeRule. Let me know if this behaves as expected for Aenea.
[App_CCR_MergeRule.zip](https://github.com/dictation-toolbox/Caster/files/3935802/App_CCR_MergeRule.zip)

mpourmpoulis:
Okay it seems I was not Not much assistance:P

My makes Shift solution Did not really satisfy me fully It turns out for good reason,As it breaks a lot of stuff. Essentially it Was only   a Workaround  around the exception caused by the differences in the constructor of the pure dragonfly and aenea Application context. it did not address the issue Off where all of these none Values came from. Well, After some debugging I figure out  they come from:
```
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'legion grid rule', 'title': 'legiongrid', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/rules/apps/mouse_grids/gridlegion.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'douglas grid rule', 'title': 'douglasgrid', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/rules/apps/mouse_grids/griddouglas.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'rainbow grid rule', 'title': 'rainbowgrid', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/rules/apps/mouse_grids/gridrainbow.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'h m c rule', 'title': 'HMC v 1.1.4', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/asynch/hmc_rules/hmc_base_rule.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'h m c confirm rule', 'title': ' :: Confirm', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/asynch/hmc_rules/hmc_confirm_rule.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'h m c directory rule', 'title': ' :: Directory Selector', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/asynch/hmc_rules/hmc_directory_rule.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'h m c history rule', 'title': ' :: Recording Manager', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/asynch/hmc_rules/hmc_history_rule.py'}, '\n')
('details ease\n', {'executable': None, 'declared_ccrtype': None, 'name': 'h m c settings rule', 'title': 'Caster Settings Window v ', 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/asynch/hmc_rules/hmc_settings_rule.py'}, '\n')
('details ease\n', {'executable': 'chrome', 'declared_ccrtype': None, 'name': 'google chrome', 'title': None, 'grammar_name': None, 'transformer_exclusion': False, 'watch_exclusion': False, '_filepath': 'C:/Users/Admin/Documents/GitHub/Caster/castervoice/rules/apps/browser/chrome.py'}, '\n')
```
So basically rules that Work only with executable or only with the title!!!!!

 Unfortunately though my solution stops the exceptions, by passing a dummy Default value instead of none, these rules break both the native and Proxy mode. For example I tested the douglas rule and after the grid appeared, there was no response to further commands rule,as it now required an executable named "dummy"

To fix this and prevent exceptions when used with aenea The following seems to work:
```python
       context = None
        constructor_parameters={}
        if details.executable is not None or details.title is not None:
            if details.executable is not None:
                constructor_parameters["executable"] = details.executable
            if details.title is not None:
                constructor_parameters["title"] = details.title
            context = AppContext(**constructor_parameters)
 
```
As we prevent passing down keyword barometers that would break the aenea set up in the default barometers for the dragonfly version are already none

On the native version, I benchmarked it against the chrome grammar ( only executable ) , the Douglas grid ( only title ) and sublime ( both ) and everything worked fine

On the remote version, I benchmarked it against fire fox ( only executable ) and again sublime( both )
The grid Though it appears in my screen does not work because it's running on the Windows side and not on the Virtual machine side so its appcontext is never triggered

As for the CCR version of my plug-in, I get an error That it is missing it CCR type, so I haven't got it working yet, But I have a feeling that it may be faced with the same issue

My apologies for the previous buggy sketch  of a solution, it seems there is more to it than originally expected.

LexiconCode:
You beat me to the punch line there as I was working through this late last night. 

> As for the CCR version of my plug-in, I get an error That it is missing it CCR type, so I haven't got it working yet, But I have a feeling that it may be faced with the same issue

That was an error on my part. I've re-uploaded the rules in the above post.

mpourmpoulis:
Many thanks for the correction, I tested it and it is working as expected(With a minor technicality in My part ). It correctly takes into account the remote context and does not throw exceptions. However, if I comment out the title keyword argument , the behavior is a little bit weird,In my first trials for some reason latency was massive ( close to 10 seconds )  and I thought  it did not work ,  but fortunately in later trials I established that commands get executed.  Nonetheless the exceptions come back and it seems to fail to differentiate between whether sublime or fire fox are in the foreground :(

LexiconCode:
> 
> 
> Nonetheless the exceptions come back and it seems to fail to differentiate between whether sublime or fire fox are in the foreground :(

Can you clarify what seems to fail? App_CCR_MergeRule is context specific to Sublime Text.

mpourmpoulis:
Precisely, here is an example  example where I have commended out Title Keyword

without
 
( in the last example Please note that sublime coming to the foreground AFTER the Command is normal, No worries)

LexiconCode:
Ultimately we have two separate issues here. correct?
1. Proxying AppContext in a way that doesn't break Aenea.
2. Rules work only with `executable` or only with the `title` with Aenea.

What is your experience with older grammars from Caster regarding the 2nd issue?

mpourmpoulis:
Yes I agree.

In order to be sure to have not missed anything I decided to retest the aenea setup using Caster  0.5.11 with the sublime rule, Which has only executable keyword and I can confirm its context is working perfectly with  aenea

```python
context = AppContext(executable="sublime_text")
```
I also tested that it works only with Title keyword instead

```python
context = AppContext(title="Sublime Text")
```

And again there was no problem. No exceptions were thrown, and it would correctly differentiate whether sublime or another application was in the foreground In the virtual machine! However when I tried leaving the executable keyword as it is and setting the title to none:

```python
context = AppContext(executable="sublime_text",title=None)
```

The same problem appeared as in the 1.1.3 version. The NatLink window was polluted with exceptions Originating from the application context and as a result The rule acted as if it was in global  scope  and not bound to sublime!

So essentially, while the whole dragonfly/aenea discrepancy could theoretically goes back quite a while, be$use I think most people would simply skip the keyword they didn't care about altogether, when they were writing custom scripts , not much of an issue came up for previous versions of Caster.( or at least 0.5). Or I don't know ,perhaps some did but got lucky or were not to really not much annoyed to file an issue.

I think there are two ways that the seizure can be dealt with right now:

1) on the Caster side Both the mapping rule maker and the part with CCR rules You send me a link to need to be patched but in general any script containing
```python
context = AppContext(executable=details.executable, title=details.title)
```
 needs testing and might need  modification, as By default the rule Details class sets title=None

2) may be Raise an issue on aenea and See if they can make  a patch on their side? Something like checking for non-values.

Not sure which is the most Feasible or the best.

LexiconCode:
> I think there are two ways that the seizure can be dealt with right now:
> 
>     1. on the Caster side Both the mapping rule maker and the part with CCR rules You send me a link to need to be patched but in general any script containing
> 
> 
> ```python
> context = AppContext(executable=details.executable, title=details.title)
> ```
> 
> needs testing and might need modification, as By default the rule Details class sets title=None
> 
>     1. may be Raise an issue on aenea and See if they can make  a patch on their side? Something like checking for non-values.
> 
> 
> Not sure which is the most Feasible or the best.

I'm not exactly sure either. Comparing degrees of effort If the check in Aenea is simple and straightforward that sounds tempting. However since the Caster rewrite is changed part of me feels it's our responsibility to fix Caster side. @Danesprite do you have any thoughts?

Danesprite:
It should be fairly simple to fix the `AttributeError: 'NoneType' object has no attribute 'lower'` error in Aenea.

LexiconCode:
Thanks for that advice! That's the direction we'll pursue then.

LexiconCode:
>     1. may be Raise an issue on aenea and See if they can make  a patch on their side? Something like checking for non-values.
> 

`_check_properties` looks like an appropriate place to check for non-values.

mpourmpoulis:
I agree,changing just a single line into

```python
if value == VALUE_DONT_CARE or value is None:
```

Seems to be enough to solve all problems at once. I tested  against both mapping and CCR rules and everything seems to be working flawlessly. No exceptions are thrown, the AppContext Works as expected and no issues with rules with only one of the two keywords.

The tests were performed with the latest version of your pull request

LexiconCode:
@mpourmpoulis thank you for all the work you put into this! I've opened up a pull request on the aenea repository. https://github.com/dictation-toolbox/aenea/pull/207 

I have also synchronize this pull request with master however that should have no impact on the issue.

mpourmpoulis:
Glad to be of assistance and I hope this time for good:)

 Many thanks to you as well , especially for your swift response and eagerness to help out when I originaly commented :D

LexiconCode:
The PR waiting upstream has been merged.

PR #709, opened 2019-12-09
move _set_rdescript to grammar manager

Cliveo:
# Title

make descriptions visible for all rules

## Description

moving the description logic

## Related Issue

https://github.com/synkarius/Caster/issues/38
 
## Motivation and Context

better visibility  in the console

## How Has This Been Tested

tested mapping and merge rules

## Types of changes

## Checklist

- [ ] I have read the CONTRIBUTING document.

### DIFF
--- a/castervoice/lib/ctrl/mgr/rule_formatter.py
+++ b/castervoice/lib/ctrl/mgr/rule_formatter.py
@@ -1,0 +1,17 @@
+import re
+
+def set_rdescript(mapping, rcn):
+    for spec, action in mapping.items():
+        # pylint: disable=no-member
+        if hasattr(action, "rdescript") and action.rdescript is None:
+            mapping[spec].rdescript = _create_rdescript(spec, rcn)
+
+def _create_rdescript(spec, rcn):
+    rule_name = rcn
+    for unnecessary in ["Non", "Rule", "Ccr", "CCR"]:
+        rule_name = rule_name.replace(unnecessary, "")
+    extras = ""
+    named_extras = re.findall(r"<(.*?)>", spec)
+    if named_extras:
+        extras = ", %(" + ")s, %(".join(named_extras) + ")s"
+    return "%s: %s%s" % (rule_name, spec, extras)
--- a/castervoice/lib/ctrl/mgr/rule_formatter.py
+++ b/castervoice/lib/ctrl/mgr/rule_formatter.py
@@ -1,10 +1,14 @@
 import re
 
-def set_rdescript(mapping, rcn):
+def _set_rdescripts(mapping, rcn):
     for spec, action in mapping.items():
         # pylint: disable=no-member
-        if hasattr(action, "rdescript") and action.rdescript is None:
-            mapping[spec].rdescript = _create_rdescript(spec, rcn)
+        _set_rdescript(action, spec, rcn)
+
+
+def _set_rdescript(action, spec, rcn):
+    if hasattr(action, "rdescript") and action.rdescript is None:
+        action.rdescript = _create_rdescript(spec, rcn)
 
 def _create_rdescript(spec, rcn):
     rule_name = rcn
--- a/castervoice/lib/ctrl/mgr/rule_formatter.py
+++ b/castervoice/lib/ctrl/mgr/rule_formatter.py
@@ -3,10 +3,10 @@ import re
 def _set_rdescripts(mapping, rcn):
     for spec, action in mapping.items():
         # pylint: disable=no-member
-        _set_rdescript(action, spec, rcn)
+        _set_the_rdescript(action, spec, rcn)
 
 
-def _set_rdescript(action, spec, rcn):
+def _set_the_rdescript(action, spec, rcn):
     if hasattr(action, "rdescript") and action.rdescript is None:
         action.rdescript = _create_rdescript(spec, rcn)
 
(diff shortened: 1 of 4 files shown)


LexiconCode:
Awesome! I look forward to testing this. We'll have to figure out how to included in the unit tests.

Cliveo:
@LexiconCode only just start using a bring rule and noticed that didn't work for that is this what you mean?

LexiconCode:
> 
> 
> @LexiconCode only just start using a bring rule and noticed that didn't work for that is this what you mean? i could fix that by adding the set back into the constructor of merge rule, but was hoping for a more elegant way to get all rules, i can't seem to find a single point where all rules converge

I've updated my previous post to clarify. That is a point BringMe rule is a SelfModifyingRule Which is A CCRType. Therefore utilizes the in constructor of mergerule.

Yes I was looking through the source code for such a centralized point for all grammar types. I think we might have to have it in a number of places for it to cover what it needs to. Not elegant but I'm not seeing any other option either.

If we end up putting it in multiple places it would be best for it to have its own file and imported.

Cliveo:
@LexiconCode looks like it now works for mapping, merge and self modifying, are there any others that i need to test for. I am new to python so not sure what the naming conventions are, any advice on how to better format the code would be much appreciated
image

LexiconCode:
The standard for naming conventions for Python is [PEP 8](https://www.python.org/dev/peps/pep-0008/). Currently the continuous integration only checks for syntax errors and executes unit tests. As time allows I plan to utilize more of `PEP 8` standards. To do this I'm evaluating [Black](https://black.readthedocs.io/en/stable/) as a way to reduce the burden on contributors and maintainers to enforce `PEP 8`. Ultimately I would like formatting to become more transparent and we can focus on the content instead.

So there is one other rule type. It's called NodeRule. A bit Dated [documentation](https://caster.readthedocs.io/en/latest/readthedocs/NodeRule/). Implemented in v1.0.0.
The documentation explains the use case. It's based off of a SelfModifyingRule rule.

Drop the following rule `DemoTreeRule` into `.caster\rules`.
You will need to add to `settings.toml` Under `[paths]`  the following on its own line `SM_DEMO_NODES_TREE_PATH = "C:/Users/<Insert username here>/.caster/data/sm_demo_tree_node.toml"` 
Restart Dragon.

 To enable the rule say `Enable Demo Tree Rule`

If you forget to add the path above two settings the following would print out.
`TreeRuleConfigurationError: Path 'SM_DEMO_NODES_TREE_PATH' was not found in the 'paths' section of settings.toml. Did you add it?`

```Python
from castervoice.lib.actions import Text
from castervoice.lib.merge.selfmod.tree_rule.tree_node import TreeNode
from castervoice.lib.merge.selfmod.tree_rule.tree_rule import TreeRule

H = TreeNode

demo_nodes = H("I say zero", R(Text("LevelZero")), [
            H("I say one", Text("Level One A"), [
                    H("two A one", R(Text("Level Two A1")), [
                            H("three A one", R(Text("Level Three A1"))),
                            H("three A two", R(Text("Level Three A2"))),
                            H("three A three", R(Text("Level Three A3")))
                        ]),
                    H("two A two", R(Text("Level Two A2")))
                ]),
            H("level one be", R(Text("Level One B"))),
            H("level one see", R(Text("Level One C")))
    ])

class DemoTreeRule(TreeRule):
    def __init__(self):
        super(DemoTreeRule, self).__init__("demo nodes", demo_nodes)

def get_rule():
     return [DemoTreeRule, RuleDetails(ccrtype=CCRType.SELFMOD)]
```
To clarify the flow of commands throughout different node levels I've included a flowchart.


image

For simplicity sake think of each Node level as its own grammar. To begin transitioning the notes say `I say one` It will perform that action and advance to the next level (Node). However if I say `level one be` the NodeTree will reset back all the way to the beginning `I say zero`.

Currently it also prints `none` when the commands are executed. As for where the code needs to change to `set_rdescript` that I'm still looking into.

LexiconCode:
As a side note I would not be against merging without implementing it for `NodeRule`.

Cliveo:
@LexiconCode thank you for all of the detail, sorry it has taken me so long it is a very busy time of year. i have done an update to include tree rules and it seems to be working, it doesn't work for level one as it is not a registered action. i struggled to fit it in the current structure as these are action series's in the tree rule mappings not just registered actions, so i split the set_rdescript into two methods

castervoice/lib/ctrl/mgr/rule_formatter.py
```diff
@@ -1,10 +1,14 @@
...
-def set_rdescript(mapping, rcn):
+def _set_rdescripts(mapping, rcn):
     for spec, action in mapping.items():
         # pylint: disable=no-member
-        if hasattr(action, "rdescript") and action.rdescript is None:
-            mapping[spec].rdescript = _create_rdescript(spec, rcn)
+        _set_rdescript(action, spec, rcn)
+
+
+def _set_rdescript(action, spec, rcn):
```
LexiconCode: A small suggestion `_set_rdescript` might be better renamed to `_set_the_rdescript` That helps differentiate function names.

LexiconCode:
No worries, I appreciate your time and contribution especially this time of year. Thanks to your efforts to the caster user experience will be much improved!

castervoice/lib/ctrl/mgr/rule_formatter.py
Cliveo: good call i've committed this

PR #730, opened 2020-01-15
Bug fix: multi-clipboard wasn't initialized.

Versatilus:
# Trivial change

Fixed a bug where the multi-clipboard was being erased whenever it was **pasted from** before being **copied into**. This was due to failing to initialize the multi-clipboard on module load.

Initialization was segregated into a function during a refactor but that function was never called.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ caster/bin/data
 
 # Ignore dist files
 dist
+build/
 castervoice.egg-info
 /.stignore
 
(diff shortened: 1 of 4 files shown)


LexiconCode:
Nice catch so this needs to be anywhere that writes to a file utilizing settings?

Versatilus:
I'm not sure what the logic is behind these changes. I'm still trying to eliminate that trivial test error.

LexiconCode:
Good to go!

PR #736, opened 2020-02-01
Replace VirtualDesktopAccessor with pyvda

mrob95:
I have been meaning to do this for some time. I have reimplemented the functionality offered by the VirtualDesktopAccessor DLLs in Python using COM. It will definitely be good to not have to compile and ship multiple versions of a binary in the future, just a Python package which should work on all versions and architectures. I have also mostly fixed the bug where moving to a different workspace wouldn't bring the new active window into focus.


LexiconCode:
Thanks for this! 

I was going to re-factor Virtual Desktop Accessor grammars but I saw this in your repository a while ago and figure it would float upstream upstream when ready :)

It's an idea that pyvda could be easily extended to other platforms with the help of some contributors.

LexiconCode:
Since it's not Caster original work (In the sense that it's in its own repository with license)  do you want to include it under the LICENSE.md?

mrob95:
It's up to you but personally I don't think it's necessary since it is just an imported library, caster is not actually distributing any of the code.

PR #739, opened 2020-02-02
Pass-through text formatting commands

dusty-phillips:
When I am writing natural language,
for example in comments and READMEs,
I often wish the dictation would allow Dragon to
choose the capitalization method for me.
Sometimes I want to choose the initial letter, however.

As an example consider the phrase,
"My name is Dusty and I like to dictate"

If I dictate this without a caster command,
it comes up the way I want it to,
but I often use either 'Sing' or 'laws'
because I want to control the capitalization of the 1st letter,
or because I want to insert it with other commands in CCR mode.

When I do that to my example sentence,
it lowercases my name and the 1st person 'I.'
The 3 text formatting commands I've added
operate on only the 1st character, but leave the remaining
words is capitalized by Dragon.

### DIFF
--- a/castervoice/lib/textformat.py
+++ b/castervoice/lib/textformat.py
@@ -14,6 +14,9 @@ class TextFormat():
     3 Gerrish - camelCase
     4 sing - Sentencecase
     5 laws (default) - alllower
+    6 say - whatever speech engine provides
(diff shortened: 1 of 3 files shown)


LexiconCode:
Thanks for this and especially updating the tests as well! I don't mind adding the documentation. I will test this out a little later today.

PR #740, opened 2020-02-02
add command for capitalization

dusty-phillips:
One of the most common corrections i have to make
is to change the capitalization of a word and recently dictated text.

This commit introduces the "capital"  text manipulation command,
which can be used to make a word or character
upper or lowercase.

## How Has This Been Tested

I've tested this was every combination of characters versus dictation, direction, and uppercase versus lowercase versus default that i could think of

## Types of changes

## Checklist

    - **I couldn't find any tests for test manipulation**

### DIFF
--- a/castervoice/rules/core/text_manipulation_rules/text_manipulation.py
+++ b/castervoice/rules/core/text_manipulation_rules/text_manipulation.py
@@ -116,6 +116,17 @@ class TextManipulation(MergeRule):
             R(Function(text_manipulation_support.select_until_phrase,
                        dict(character="phrase"), dictation_versus_character="character"),
             rdescript="Text Manipulation: select until chosen character"),
+
+        # capitalized 1st word of text or character
+        "capital <direction> [<number_of_lines_to_search>] [<occurrence_number>] [<letter_size>] <dictation>":
+            R(Function(text_manipulation_support.copypaste_change_phrase_capitalization,
+                       dict(dictation="phrase"), dictation_versus_character="dictation"),
+                 rdescript="Text Manipulation: change capitalization phrase"),
+        "capital <direction> [<number_of_lines_to_search>] [<occurrence_number>] [<letter_size>] <character>":
+            R(Function(text_manipulation_support.copypaste_change_phrase_capitalization,
+                       dict(character="phrase"), dictation_versus_character="character"),
+            rdescript="Text Manipulation: change capitalization character"),
+
     }
     new_text_punc_dict = text_punc_dict()
     new_text_punc_dict.update(alphabet_support.caster_alphabet())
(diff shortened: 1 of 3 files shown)


LexiconCode:
Could we leverage the current system for formatting commands? That way we could to reuse commands and formatting functionality. For example `sing ross lower multiple` The other formatting's could be helpful as well. However I respect your time and it may not be an option at this time. Therefore we can merge as is but your thoughts would be helpful.

dusty-phillips:
Something like `replace ross multiple with <formatting command>` where <formating command> could be `Sing multiple` in this case would be great. It's a step toward a powerful modal component system like the keyboard modes used in vim. That would allow voice commands to be built up on demand instead of having to write custom ones.

But that's a pretty huge undertaking and I don't know where to start. It's feeling like caster might eventually need strong application integration (I'm thinking having its own editor like dragon pad, except not awful, or maybe a plugin to vscode or another editor, or something that is windows accessibility aware so it can operate on generic text boxes). The text manipulation rule is a brilliant hack that I can't believe works across applications, but I don't know how far it can be pushed. I'm up for doing some refactoring on the text manipulation rule at some point, but I need to get more efficient at voice coding, first. Right now I'm barely able to keep up with my work (which is an amazing improvement over where I was before I started dictating, so I'm happy!).

Even if I did have modal commands available, I'd probably still want a `captial` command because `replace <text> with <command>` is a mouthful to say and for the recognizer to identify correctly. So I'd suggest merging as is and looking to the future to enhance the system.

LexiconCode:
> The text manipulation rule is a brilliant hack that I can't believe works across applications, but I don't know how far it can be pushed. I'm up for doing some refactoring on the text manipulation rule at some point, but I need to get more efficient at voice coding, first.

That's exactly how it described the text manipulation commands. Fortunately there's a few things in the works. One of which should be an overhaul of the text manipulation/The jump in and jump out commands. 

The text manipulation commands will work line by line and be tokenized for recognition boost. The other improvement is the regex to handle camelcase, underscores and so on. Even to the point of defining variables by style. So you can say next variable.

Even then there are limitations especially how applications treat input and clipboard function.

Last but not least in this is where I hope the future is dragonfly accessibility API.  There's a grammar specifically to use that API [here](https://github.com/LexiconCode/Caster/blob/Accessibility_API/castervoice/rules/core/text_manipulation_rules/accessibilityapi.py) Windows support is limited but it works with Firefox, Chrome and Gitter. The reason it's not integrated directly into Caster yet is because in upstream dependency which registers the dll is merged in repository but not pushed in a release. It's a great replacement for Dragon's built-in Select-and-Say for browsers. 

The overhaul of text manipulation commands will use the API when available and fall back to the clipboard-based method.

dusty-phillips:
That sounds really amazing. I am looking forward to the future!

PR #767, opened 2020-03-12
Trivial Change: Add missing back tab command in chrome

jdcai:


LexiconCode:
I would take out the extra optional's for the back, and rename it to `(back|backward)` That way both commands are paired equally with sets of commands.
```       
"(next|forward) tab [<n>]|tab (right|sauce) [<n>]":
      R(Key("c-tab")) * Repeat(extra="n"),
"(back|backward) tab [<n>]|tab (left|lease) [<n>]":
      R(Key("cs-tab")) * Repeat(extra="n"),    
```

jdcai:
I copied it from firefox.py so i'll change it there too. Though I think it should be `(back|previous)` since most commands use previous.

LexiconCode:
Sounds good then that would have been something that I accidentally added from converting from the pseudo grammar API.

PR #768, opened 2020-03-12
VScode: "go back" default to 1

jdcai:


LexiconCode:
Good catch, thank you!

PR #771, opened 2020-03-13
Add Webex Teams Rule

MarkRx:
# Title

Add a rule for Webex Teams

## Description

Integration with Webex Teams. I used the keyboard shortcuts documented [here](https://help.webex.com/en-us/7wr87q/Keyboard-Navigation-and-Shortcuts-for-Cisco-Webex-Teams). I didn't implement meetings/calls since it seems strange to me use commands while on a call - they could be added later. I didn't implement markdown - the markdown rule should be used instead.

A few notes:
- Wasn't sure what to call the search/find command. The UI calls it "search" but I could see a case to use "find" as well. I ended up using "search | find" like vscode
- There is some text formatting for bold/italicize/underline which uses "standard shortcuts" which could be put somewhere shared in the future. I saw some modules implement it themselves so I did the same
- Commands were kept simple / "unchained" as it is easier to add functionality than remove functionality. The only one I chained was "contact person (name)"
- Chances are speech engine needs to be taught the word "webex"
- One command is windows specific since it uses the Windows key

## Motivation and Context

Adds integration with Webex Teams application

## How Has This Been Tested

I ran through every command while using Webex Teams

## Types of changes

## Checklist

Not sure how this would be unit tested.

- [x] New documentation is clear and complete.

### DIFF
--- a/castervoice/rules/apps/chat/webexteams.py
+++ b/castervoice/rules/apps/chat/webexteams.py
@@ -1,0 +1,114 @@
+from castervoice.lib.actions import Key, Text
+from castervoice.lib.ctrl.mgr.rule_details import RuleDetails
+from castervoice.lib.merge.additions import IntegerRefST
+from castervoice.lib.merge.state.short import R
+from dragonfly import (Dictation, MappingRule, Repeat)
+
+class WebexTeamsRule(MappingRule):
+    # See https://help.webex.com/en-us/7wr87q/Keyboard-Navigation-and-Shortcuts-for-Cisco-Webex-Teams
+    mapping = {
+        # Navigation - Primary
+        "show spaces":
+            R(Key("c-1")),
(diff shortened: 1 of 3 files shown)


LexiconCode:
>     * Chances are speech engine needs to be taught the word "webex"

Webex Teams could be changed to 'Web X Teams' for the pronunciation?

>     * One command is windows specific since it uses the Windows key
I believe dragonfly translates `windows key` to other OS equivalents as needed.

MarkRx:
Can the rule be given more than one pronunciation? After thinking about it I realized that if someone uses Webex a lot they would probably want to teach their dictation engine the word for normal dictation. If that happens will calling it "web X" still work?

LexiconCode:
I'm not hundred percent sure actually `(Webex Teams | Web X Teams)`

Issue #711, opened 2019-12-10
Override source code `support files` for grammars in user directory.


LexiconCode:
Currently Caster tracks grammars in the `castorvoice\rules`. When grammar is copied over to the user directory `.caster/rules` it overrides imports so that it's loaded from the user directory `.caster/rules` instead of source code `castorvoice\rules`. 

For instance if you copy the grammar file `eclipse.py` into the user directory, It will load from the user directory not the source code. Folder structure doesn't matter `eclipse.py` could be placed anywhere.

However it does not do the same for support files which contain functions. The template follows this standard (AppName_support/CcrName_support). For example `eclipse_support.py` Even though it's included with the grammar file in the same directory it still loads from source not the user directory.

This will cause a lot of confusion. What expect to be able to edit both the grammar and support files and have them loaded from when copied to the user directory.

Why not use conditional imports?
```Python
try: # Try first loading from user directory
    from eclipse_support import ec_con
except ImportError: # load from source code.
    from castervoice.rules.apps.editor.eclipse_rules.eclipse_support import ec_con
```
That does place limitations though. The user cannot rename `eclipse_support.py` or have it in a different directory without having to change imports.

This limitation isn't too bad as long as these are copies `eclipse_rules` directory. As long as the files are together this shouldn't be an issue. This is the case for all grammars in `apps` and `ccr` folder as they do not import each other. 

Where this becomes a significant issue is the `core` directory. Application and CCR grammars do import various `core` support_files. 

For instance "eclipse2.py" grammar imports `alphabet_support.py`
`from castervoice.rules.core.alphabet_rules import alphabet_support`

Conditional imports are too fragile as we don't know where `alphabet_support.py` could be placed by the user.

LexiconCode:
Temporarily addressed through conditional imports https://github.com/dictation-toolbox/Caster/commit/205b39a3b143081e2ff5bc11955b6a30219ffcf8 until a more elegant solution is devised.

LexiconCode:
Note: this also affects from `castervoice.rules.ccr.standard import SymbolSpecs` in `ccr`

kendonB:
There are many of these support files that either the user would want to customize or are a dependency of a file that the user would want to customize. An ideal solution would allow the user to move either a rule file or a support file to the `user/rules` folder, or any subfolder thereof, and the user file takes precedence over the caster file.

Implementing this would allow the user to either move a `_rule` folder and make edits or move an individual file and make edits - both would "just work".

[...]

PR #775, opened 2020-03-20
Sublime snippets first part

mpourmpoulis:
# Title

Sublime snippets first part

## Description

This is part one of an attempt to integrate https://github.com/mpourmpoulis/CasterSublimeSnippetInterfaceExample into Caster

@LexiconCode I know this is a busy period For you so fear not as this is not ready to merge! There are quite a feat few things that need to be taken care off, documentation to be written and improvements to be made and will gradually get round to it over the next week. This is just a preliminary a PR For testing purposes

In a nutshell these pull requests contains

- Files under the lib folder that handle the various tasks needed for implementing the desired functionality,such as

    - Interfacing with sublime

    - generate the the actual snippet text

    - maintaining a snippet state

-  much of the above functionality  is exposed via the dragonfly custom actions `Snippet,SnippetVariant,SnippetTransform,DisplaySnippetVariants` Which you can import from `castervoice.lib.sublime_snippets`

- a global control grammar,with which currently does not have much( mainly the commonly used `variant <n>`, so you do have not to define it everywhere) in which we enable by `sublime snippet control` ( to be improved)

But there are some improvements as well Like  Snippet lambda can receive more arguments

And something to be the more juicy,which I want to eventually fully integrate in the control grammar But will need work

## Motivation and Context

Say goodbye to .sublime-snippet files

## How Has This Been Tested

Tested locally on Windows, need more people to test

## Types of changes

- [x ] New feature (non-breaking change which adds functionality)

## Checklist

- [ x] I have read the CONTRIBUTING document.
- [ x] I have checked that my code does not duplicate functionality elsewhere in Caster.
- [x ] My change requires a change to the documentation.


mpourmpoulis:
has a lot of technicalities but still I pushed a proof of concept of activating only meaningful variance using a self modifying rule and a decorator

example11

LexiconCode:
Nice Work, if I'm feeling better will test this tomorrow. 

I have a few thoughts regarding the structure of the the modules you've included. Historically many things restored in `lib`. Much of which is now been moved into support files for their respective grammars. A recommendation would be to create a sublime folder in editor and keep all of its supporting functionality there.`

Broadly speaking application grammars and supporting grammars are paired with their supporting functions in a `<grammarname>_support.py`. Once an application graduates to having a support file or multiple grammars it gets its own folder. VScode is an example. If there's enough different types of functions they that could have multiple support files.

Function_like_utilities.py` might be re-factored into `lib` if more than one grammar utilizes it. For now it would be better stored in the sublime folder.

Functionally what this does is separate the grammars in the core code of Caster. Its a WIP as I'm still working on re-factoring's a few things out of the core code back into grammars and supporting files.

In order for the override system for the user directory to work you will need to place conditional imports in the code. These imports try to import locally first if not found them from source code. 

```
try: # Try first loading from caster user directory
    from eclipse_support import ec_con
except ImportError: 
    from castervoice.rules.apps.editor.eclipse_rules.eclipse_support import ec_con
```

Not ideal my opinion but will work until we get a better system in place for support files.

mpourmpoulis:
Thank you very much for your feedback! Polluting the `lib` folder is something I did fear and I agree with you that it would be best if those files were moved under `sublime_rules` to keep things compatible with the rest of the project. 
I will try to restructure it accordingly! Nonetheless, I do have a couple of concerns, though this might be due to me misunderstanding something or they may turn out to be unimportant after all.

Firstly, the main usage of the supporting functions is most likely going to be in user made grammars. Even the control grammar is here to support such custom grammars. As a consequence, conditional imports are going to be needed note only in the code that we ship by default but kinda everywhere in order to avoid inconsistencies or things breaking.

Secondly,even though the current functionality can be wrapped up inside a single file without losing too much readability, I'm not sure sure that this will be the case forever, for instance if we port and more advanced version of . More importantly, if the community finds these features beneficial, it might also make sense to provide at some point utilities that could make writing more complex snippets even easier and faster, such as voice commands (`placeholder 1` to type `${1:}` in one go), commonly used regular expressions ( so we can write `"$1 = "+capitalize("$1")+"()"`), and what ever is deemed necessary. so there's a chance we may end up with multiple files the users might want to override and thus a need for multiple conditional imports.

Either way I think I might be needlessly overthinking this potential issue and  I will start by placing everything inside `sublime_rules` folder.  Any further thoughts whether on this or  the  pull request in general are welcome and I hope you feel better soon:)

mpourmpoulis:
It turns out that my feeling of conditional imports could be error-prone If used incorrectly was correct after all and unfortunately I found about it the hard way:)

All those conditionally imports never enter there except clause because all the vscode_rules,sublime_rules, are in the sys.path!!! so they're not necessary as they're not doing anything:D

But with my current implementation, If a user grammar Forgot the conditional import and imported directly with the long`castervoice.rules.apps.editor.sublime_rules.sublime_snippets` then the controlled grammar gets broken!( not really sure but it seems as if these two grammars get two different copies of the same module oor something need to look into it Over the coming days)

LexiconCode:
Oh interesting, I knew that core code was added to path. So the reason the conditional imports exist is all is for the user directory. I would've not predicted it would have issues in the source code receiving two modules. Because the imports come from the same place.

As you figure it out could you describe the methodology are using to test this and the issues you're seeing. I'd like to be able to get to a place where I can help you troubleshoot this.

Are you using a debugger?

mpourmpoulis:
Thank you very much for your eagerness!

Well I was debugging in the little bit more primitive way Using
[custom_cpp_more_experimental.zip](https://github.com/dictation-toolbox/Caster/files/4380882/custom_cpp_more_experimental.zip)

And I had the common variable `snippet_state` printed both when a snippet is inserted and when that control grammar is about to refresh only to find out different values were printed! These had me thinking that This is probably an import issue And after adding the following code In both grammars 

```
try : 
    # raise ImportError()
    from sublime_snippets import Snippet,SnippetVariant,DisplaySnippetVariants,snippet_state,send_sublime,SublimeCommand,mark_as_snippet_grammar
    print(" success with dry")
    print(" success with dry")
    print(" success with dry")std::cerr<<  << " " <<  << " " <<  << std::endl;
except ImportError:
    print(" and import there is expected")
    print(" and import there is expected")
    print(" and import there is expected")
    from castervoice.rules.apps.editor.sublime_rules.sublime_snippets import Snippet,SnippetVariant,DisplaySnippetVariants,snippet_state,send_sublime,SublimeCommand,mark_as_snippet_grammar
```

From the NatLink window messages it became clear that only the try was executed And by toggling The comments in `    # raise ImportError()` in one of the grammars And rebooting you can switch from things working two things breaking! (So long as Both of them enter either try or except It is fine) You can verify this by trying `error line 3` then `variant output`

To make sure I understood path correctly I also added
```
import sys
print(sys.path)
from eclipse_support import ec_con
```
(my user directory Does not contain anything for eclipse)

So I more or less came to the conclusion you summarized:

> From my understanding I'll try to restate your issue. The original file is already added to the sys path. When the conditional import is called it's except clause is never reached because the relative import is always true. The end result may be there's two instances of the module being referenced which may not be a identical.

And it turns out that having subfolders of the package inside the path is a known issue But can cause trouble


> The only time the relative import is not true is that the user does not copy the entire folder breaking the relative imports.

I'm not sure I understand this one, could you provide an example of such a directory structure?

LexiconCode:
I'll respond in multiple posts but I'll try to clarify an answer to your last question. Forgive me if I'm a bit verbose on my explanation and may repeat knowledge previously discussed. In part it helps me verbalize and verify my understanding through our discussions.

> I'm not sure I understand this one, could you provide an example of such a directory structure?

image

The first window on top represents the core Caster code. The second represents the user directory. Both of them contain identical `eclipse` grammars and support files.

When Caster loads It compares the `grammars` (not support files or other modules) from the user directory to source code. The primary mechanism for evaluating paths for grammars is scanning files for `def get_rule():` line by line then grabbing the class name. Once it does that it overrides the source code import with the user directory impport. ( which I don't fully understand yet) but can show you where in the code. It might be helpful to do a remote session on my PC to chat over voice on this.

All right so now that we have the basics there let's talk about modules A.k.a. `Support files`.  Imagine if the user makes an edit to `eclipse_support.py.` without a relative import changes would not be loaded by the grammars because of utilizing source instead. That's why I tried still load from the user directory first if it exists (If in the user directory the path would be different). A user expects edit the support files and have those changes reflected in the grammars. That's why I recommend copying directories like `eclipse_rules` as it preserves the file structure for relative imports.

My order of priorities.

1. Grammar should always work.

2. Satisfy user expectations with editing modules

Of course this breaks down when you look at the `core` directory in the user directory as they are referenced by other grammars. See Issue [711](https://github.com/dictation-toolbox/Caster/issues/711)

Now imagine for a second `eclipse_support.py` was not included in the user directory above. I understood `grammars` paths would be compared in the user directory to the source code. However I didn't consider how this would work with `modules` being added to SYS path at the same time. So that's why the try except block exists to make sure support files were always referenced based on my original understanding. However now that I know the source code modules not just grammars are in the SYS path the `try import` Is still needed the `Except` statement is not necessary.

However `try import` is potentially the source of the issues as you've pointed out. Does that make sense?

Once we have this issue nailed down let's update [711](https://github.com/dictation-toolbox/Caster/issues/711) with our findings.

mpourmpoulis:
the previous set of commits should be solving the problem with sublime path we talked about [here](https://github.com/mpourmpoulis/PythonVoiceCodingPlugin/issues/15)  and by following the example of `simple_alias` the control grammar is now not CCR and bound to the sublime context!

I am leaving a small checklist here for me to better keep track of what still needs to be done!

- [] make the control grammar work when the snippet was inserted by a CCR grammar( in those cases the dragonfly callback receives an RepeatRule!)

- [] make the whole system more lenient towards declaring a default value for a placeholder twice. Sublime does not like it when `${1:default}` appears twice and as a consequence lambda generating snippets should have extra code in order to avoid such situations. in order to avoid users having to write boilerplate code, a solution would be to programmatically remove duplicate default values before sending them sublime.

- [] solve problems with reregistering the same call back after the control grammar reloads. Fortunately, even if the call back is a registered multiple times, due to condition checking in the `_refresh` method the grammar is only going to refresh once but this is still an issue


- [] handle ties,when multiple extras share a common spoken item, and display all the alternatives of the quick panel

- [] create commands of the form `display <choice_name> variant` that will prompt the user with all the alternatives for a particular extra,the decorator should probably be adopted to support remapping when dealing with unspeakable words

- [] add type checks and docstrings everywhere

mpourmpoulis:
okay but the merge conflict first, at the moment not really sure how to fix it, as both Github Desktop do not show conflicts  and manual  `git merge --no-commit`  and `git status` yield 

```
All conflicts fixed but you are still merging.
  (use "git commit" to conclude merge)

Changes to be committed:
        new file:   castervoice/rules/apps/editor/sublime_rules/Function_like_utilities.py
        new file:   castervoice/rules/apps/editor/sublime_rules/__init__.py
        renamed:    castervoice/rules/apps/editor/sublime.py -> castervoice/rules/apps/editor/sublime_rules/sublime.py
        new file:   castervoice/rules/apps/editor/sublime_rules/sublime_snippet_additional_control.py
        new file:   castervoice/rules/apps/editor/sublime_rules/sublime_snippet_control.py
        new file:   castervoice/rules/apps/editor/sublime_rules/sublime_snippets.py
        modified:   tests/rules/apps/test_rule_modules.py
```

Which shows that the local git Can handle the renaming on its own... Either way I will keep looking into these...

Now for the interesting part, the variance system received an upgrade  and for the most part is at long last finished. In particular, if a user say something like 

```python
variant output
```

If there is a single extra consumed by the snippet, the corresponding variant should be inserted right away, but if there are two or more extras, then the user should be prompted to choose which one from the quick  panel.
Initially, my approach was to precomputed all the conflicts, but doing these for any pair of dragonfly elements of any type proved highly nontrivial! (for instance imagine choice element containing " hello world" overlapping with a Repetition of up to 2+ times of the words ["hello","world"]). So in the end, a hacky solution was employed using a local text engine to check if the spoken words could be matched by more than one extras. so far in all my tests they seems to be doing the trick, and any performance overheads seem nonexistent :D
Furthermore, if there are no variance for a snippet, via use of the function context, the control rule goes inactive! Nonetheless, in order to preserve sanity,`display <extra_name> variants` we only be available for integers and shallow choices!

Furthermore, after getting around to write code to handle the double default issue , namely `${1:default value}` causing problems if it appeared twice or more,I realized it would no longer reproduce this and after asking the sublime [forum](https://forum.sublimetext.com/t/question-for-duplicate-placeholder-default-values-for-the-same-snippet-field/52848) it appears that unless these was a bug appearing only in a very particular build, it was most likely a personal error in my experimentation, so my code not really needed. I have pushed it just in case we may need to borrow bits from it in the future for other purposes but the corresponding function is not getting called!


Now over the coming days, I plan to restructure the directory structure, settle on the final API, add some utilities, resolve issues with the variance system when the snippet grammar is self modifying and  write documentation. Ideally, I would like to find a solution for when the snippet grammar  is CCR, but so far this has appeared to be a pretty tough nut to crack and I would prefer to avoid further delays... 

Afterwards, my intention is to create a snippet_grammars repository, to start stockpiling snippets the users can clone directly in their user directory and and hopefully if some of them are mature enough, they could one day make it to Caster proper.

LexiconCode:
> my approach was to precomputed all the conflicts, but doing these for any pair of dragonfly elements of any type proved highly nontrivial! (for instance imagine choice element containing " hello world" overlapping with a Repetition of up to 2+ times of the words ["hello","world"]). So in the end, a hacky solution was employed using a local text engine to check if the spoken words could be matched by more than one extras. so far in all my tests they seems to be doing the trick, and any performance overheads seem nonexistent :D

That's a very creative approach with text engine. Can you describe when the snippet grammar should change to CCR?

mpourmpoulis:
essentially the issue boils down to the fact that due to the way the CCR grammars merge together into RepeatRules , The only information about the original rules themselves is encoded in their registered action due to the `_set_rdescription` Called by the merge rule [constructor](https://github.com/dictation-toolbox/Caster/blob/30cb4371fc30b2edb65541420bf265e01c17b22d/castervoice/lib/merge/mergerule.py#L26),

As a consequence,when the recognition observer calls back receives those repeat rules, it makes life harder To isolate which particular CCR was responsible for the snippet inserted, something that was needed with the old Implementation Even for the very miniumum of retrieving the extras That we will upload as variants, but I'm working to change locally So they get retrieved directly from the merged rule. Still, there are more subtle issues  that need to be addressed as for example  it brings into question The idea of Provide variants Only for the grammars the user has intentionally  decorated

mpourmpoulis:
okay the previous series of commits should fix any issues with CCR's and self modifying rules and I have decided to remove the need for decorating only specific grammars, which to be honest,might not have been such a great idea in the first place, for many users it would simply raise complexity of usage without offering that much. Also I took the liberty Of making these small [change](a890051be4ed1dd33df105be90db05d9dc7a3518) so when you get a chance to look at this let me know if that is an issue

From now on, things to be done

- [ ] Restructure the big file into multiple ones and create `sublime_support` file from which users will be importing

- [ ] Cleanup and finalize, or at leastformat it in a way that can be extended easily in the future, the whole thing with the additional snippet parameters as they can be used for some spiced up features-in-the-future

- [ ] Write documentation. My thought was to create a dedicated Snippets.md file as there will probably be Quite a few things to document. Would that be okay? Furthermore, I was in easily considering using gifs but from what I see  the rest of the documentation is just plain text and Use links for video/demonstrations hosted elsewhere.

- [ ] create a series of snippets for dev rules(similar to do the existing commands like `dev Mouse` but slightly more complicated). I already have some I use personally but they need a little bit polishing

- [ ] Create some initial utilities For writing snippets faster
    - [ ] Probably a handful of voice commands, For instance a command to print `${1:}`
    - [ ] some can be used programmatically, a function `placeholder` when called like `placeholder(1,"data")` return  `"${1:data}"`

- [ ] utilities for easy loading Snippets stored in a dictionary to a Rule might also be good

LexiconCode:
> Also I took the liberty Of making these small [change](https://github.com/dictation-toolbox/Caster/pull/775/commits/a890051be4ed1dd33df105be90db05d9dc7a3518) so when you get a chance to look at this let me know if that is an issue

I think this is a good move, so the question then becomes what happened to the grammar somehow becomes invalidated when refreshing? we could hold its previous state and a singleton and update that singleton when the rule is successfully validated.

PR #777, opened 2020-03-31
Trivial change: Expand function key range to include F10-F12

TTeig:
# Trivial change: Expand function key range to include F10-F12


LexiconCode:
Thanks

PR #779, opened 2020-04-02
[Trivial change] Fix broken sample.py command

TTeig:
# [Trivial change] Fix broken sample.py command
The output of `"i choose <choice>" : Text("<choice>")` was just `<choice>`


LexiconCode:
Thanks for this! Some things like this slip through when I was updating the documentation and examples.

PR #787, opened 2020-04-09
Trivial Change: fix alias link

TTeig:
# Trivial Change: fix alias link
A link to the alias page in the docs pointed to Aenea instead.


LexiconCode:
Thanks looks like a a copy and paste error on my part.

Issue #701, opened 2019-11-26
Speed up Legion using sampling and/or subsets of the screen


kendonB:
**Is your feature request related to a problem? Please describe.**
Legion exhibits noticeable latency due to the computational requirements of the algorithm. This is especially noticeable on a 4K display (@LexiconCode said this once).

**Describe the solution you'd like**
It would be good to down sample the image to only what is required to detect the text on the page before processing.

Additionally, we could limit the bounds of the image using a number from 1 to 9 denoting a region of the screen.

LexiconCode:
That's an interesting thought and shouldn't be too difficult to do at least in the portion of downscaling a screenshot. However scaling it to the current screen afterwards is a bit of an unknown to me.

ethangoldstein:
I have some experience dealing with image processing. Computations on images are very processor taxing, especially once the image size becomes 4k. Capturing a screenshot and then down sampling it would require even more processing as it would have to be low pass filtered and then decimated. The GPU is designed for these types of computations, not the CPU. It may be worth sending the computations to the graphics card. Another potential solution is the write the computationally expensive part in C. I have had success with that in the past using Cython.

kendonB:
> Another potential solution is the write the computationally expensive part in C.

I would hope there's a package that has already done this

LexiconCode:
> > Another potential solution is the write the computationally expensive part in C.
> 
> I would hope there's a package that has already done this

For image manipulation Caster already utilizes [Pillow](https://python-pillow.org/) library. Pillow which leverages the power of the C directly (rather Cython) than wrapped up in Python. It does not leverage GPU instances like (CUDA) but runs on CPU.

Premature optimizations undesirable. However profiling or even just simply using timeit to see which part of the process is taking longer. 

`castervoice\asynch\mouse\legion.py`  is where "tirg-32.dll" ts loaded for Legion and gdi.

`castervoice\lib\gdi.py` mostly deals with grabbing the image on the screen for processing.

kendonB:
This one was partially addressed in #788. https://github.com/kendonB/Caster/tree/downscale_legion2 implements subsets of the screen but finds no major speed up. It seems like the big bottleneck is actually spinning up the process. Closing for now.

PR #788, opened 2020-04-10
downscale legion

kendonB:
# Downscale legion by half before processing

## Description
One reason that legion can be slow as the image processing stage, especially on a 4K screen. This pull request changes the default resolution to be half of the screen's native resolution. On my 1080p screen I find a speed up of around 0.1 seconds in the processing stage. Printing to screen is also a bit faster because there are fewer rectangles that get picked up.

The PR introduces another extra "rough" which can take values "rough" and "detailed". The default is "rough". "detailed" uses the full image for processing.

## Related Issue

#701 addresses part of this issue (sampling) but not subsetting by region.

## Motivation and Context

Legion can be slow. This helps to speed it up.

## How Has This Been Tested

1. Trying out "legion", "legion rough", "legion detailed". 
2. Trying out "legion 2", "legion 2 rough", "legion 2 detailed"

## Types of changes

## Checklist

- [x] Remove the profiling code. I will leave it there initially so that you can see how it works on your machine.


LexiconCode:
Looking good, I've included screenshots.

Have you tested the effects of changing window scaling?
image


Legion detailed
MouseAlternatives: legion [<monitor>] [<rough>], %(monitor)s, %(rough)s
0.599999904633
image

Legion
MouseAlternatives: legion [<monitor>] [<rough>], %(monitor)s, %(rough)s
0.25200009346
image

Legion rough
MouseAlternatives: legion [<monitor>] [<rough>], %(monitor)s, %(rough)s
0.238000154495
image

kendonB:
@LexiconCode do you think that it's worth it to have a user setting for how rough to go by default?
I'm finding that going to half of 1080p works pretty well for most purposes. though the time savings are quite small. will also test window scaling right now

kendonB:
yes it works fine with window scaling

LexiconCode:
@kendonB I think that's a sensible idea to have a default setting for down scaling.

kendonB:
added a user setting @LexiconCode

kendonB:
actually i'm finding a bug associated with the re- converting back into integers. it's intermittent

kendonB:
Hold often on merging this until I've tested this for a day. after a restart, I haven't been able to reproduce the error.

kendonB:
I managed to get the error back. any idea how to defend against this?

```python
ERROR:action:Exception from function mouse_alternates:
  File "C:\Python27\lib\site-packages\dragonfly\actions\action_function.py", line 145, in _execute
    self._function(**arguments)
  File "castervoice\lib\navigation.py", line 42, in mouse_alternates
    ls.scan(bbox, rough)
  File "castervoice\asynch\mouse\legion.py", line 212, in scan
    result= [int(float(i)/factor) for i in result]
ValueError: could not convert string to float:
ERROR:action.exec:Execution failed: mouse_alternates(): could not convert string to float:
ERROR:action.exec:Execution failed: mouse_alternates()+focus_mousegrid()
```

LexiconCode:
Oh strange, I will look into that tomorrow.

LexiconCode:
Sorry it's taken me so long to get back to you on this. I have yet to replicate this so definitely let me know if you find a reliable way to replicate the error. However I do have some thoughts on what's going on.

`result= [int(float(i)/factor) for i in result]`

It's converting to an `int` as the last step.

For debugging I would print out `i` and `print type(i)`. I would examine the output to check if it is a valid number and does it contain any extraneous letters or special characters?

LexiconCode:
@kendonB 
Printing out `result` variable, here are the results.

Works as expected.
```
LegionGrid: (terminate | escape | exit | cancel)
MouseAlternatives: legion [<monitor>] [<rough>], %(monitor)s, %(rough)s
 ['686', '144', '696', '1129', '686', '1163', '696', '1043', '705', '1272', '717', '7', '706', '311', '715']
0.289000034332
```

Your error is produced with the following below.
```
LegionGrid: (terminate | escape | exit | cancel)
MouseAlternatives: legion [<monitor>] [<rough>], %(monitor)s, %(rough)s
['696', '1129', '686', '1163', '696', '1043', '705', '1272', '717', '7', '706', '311', '715', '']
```

For some reason there's extraneous `''` at the end of the list. That's what causes could not convert string to float.

There are two ways I think this could be handled:
1. Stripping out `''` from the list. 
2. Diagnose why `''` is being added to the list.

kendonB:
@LexiconCode I made that change and will try it out over the next couple of days. Not the most elegant thing but it won't cause performance issues so I'm not upset about it.

kendonB:
Okay it looks like that last fix you suggested works. I am finding that it is a marginal improvement in speed on a 1080p screen and using a down scaling factor of 0.5 can make the loss in accuracy enough to be annoying. 

Looks like the speed improvement is a lot better on a 4K screen. 

Is there a way to detect resolution at caster startup? We could then make the default down scaling factor dependent on the screen detail.

LexiconCode:
> Is there a way to detect resolution at caster startup? We could then make the default down scaling factor dependent on the screen detail.

I don't think caster startup would be the best approach. The primary reason for that thinking is because we don't know which will monitor would be used. However it's something that could be detected with the Legion command itself and then downscaled.

kendonB:
@LexiconCode I have added an automatic setting which makes the accuracy of legion quite comparable between rough and detailed for a 1080p screen. this way this change will not be very disruptive by default. Should be reasonable gains for people on 4K screens.

I think this is ready for a final review now.

kendonB:
https://github.com/kendonB/Caster/tree/downscale_legion2

@LexiconCode I implemented legion with a 3x3 grid in the above branch. I find that it doesn't substantially improve performance on my screen/system. Take a look and see if it makes a big difference on your 4k screen? Note that the default if you are choosing the grid method is to make the image "detailed". Invoke using "legion grid four" etc.

LexiconCode:
I did get a chance to experiment with the 3x3 grid. I didn't notice a significant speed up either. However I would surmise there is a significant delay with all grids and due to having to spin up a separate process. Once the GUI is unified there'll be one thread managing The grid system and the GUI. That should eliminate the associated delay. Therefore I would keep that branch until we can test with that in the future.

kendonB:
So I think this is worth keeping the downscaling as it seems to give about a 0.25 second speedup. Not a huge gain but no point in leaving it on the table. I actually think if spinning up a process is the remainder of the time cost then leaving it at that should be just fine I think. 

Stopwatch timing for me puts it at about 1 full second to get it to print to screen now. If the printing bit was free then we'd have just 0.2 leftover by the python function. 0.2 would be not too bad at all.

kendonB:
I think this is good to go now @LexiconCode.

dusty-phillips:
I'm really glad to have this!

PR #792, opened 2020-04-21
Clarified Caster vs Natlink User directory docs

RyanQuey:
Avoid confusion between use of term "User Directory" here and the Natlink User Directory

# Title: Trivial Change

### DIFF
--- a/docs/readthedocs/User_Dir/Caster_User_Dir.md
+++ b/docs/readthedocs/User_Dir/Caster_User_Dir.md
@@ -6,6 +6,8 @@ This provides a brief overview of the caster user directory. The main function o
 
 Windows  OS `C:\Users\%USERNAME%\AppData\Local\caster`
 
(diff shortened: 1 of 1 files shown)


LexiconCode:
Thank you that's a good point to clarify!

docs/readthedocs/User_Dir/Caster_User_Dir.md
```diff
@@ -6,6 +6,8 @@ This provides a brief overview of the caster user directory. The main function o
 
 Windows  OS `C:\Users\%USERNAME%\AppData\Local\caster`
 
+Please note that this directory is set by default by Caster, and is distinct to the Natlink UserDirectory, which needs to be set to the main Caster directory if using the [Alternative Natlink Configuration](https://caster.readthedocs.io/en/latest/readthedocs/Installation/Dragon_NaturallySpeaking/#-alternative-natlink-configuration).
```
LexiconCode, line 9: At this point setup should be complete and engine agnostic.

**DNS/DPI** The Caster User directory is managed by Caster, and should not be confused with Natlink's UserDirectory, which is set to the main Caster directory if using the [Alternative Natlink Configuration](https://caster.readthedocs.io/en/latest/readthedocs/Installation/Dragon_NaturallySpeaking/#-alternative-natlink-configuration).

RyanQuey: Love it, let's do that then, thanks!

PR #803, opened 2020-04-30
Trivial change: Four new Rust syntax commands

dusty-phillips:
# Trivial Change: Four new Rust commands

## Description

`enum` , `async`, and `await` are often misrecognized by dragon,
so having extra commands for them reduces frustrationand can.
The clone command is just a quick shortcut.


dusty-phillips:
Oops, it included a line from my personal customizations that shouldn't have gotten in there. Rebuilding now...

PR #805, opened 2020-04-30
Add command to open line above

dusty-phillips:
I'm not sure if this is generally useful, so feel free to close.

This command adds a character at the end of the current line
and then insert a new line above.

It is symmetric to the `tell` command, which does the same thing,
except
goes below the current line.

both commands are analogous to the o and O commands in vim.

I chose `hark` as it is kind of the inverse of `tell`,
but dragon often hears it as `heart`, so I used both.

### DIFF
--- a/castervoice/lib/navigation.py
+++ b/castervoice/lib/navigation.py
@@ -222,13 +222,24 @@ def curse(direction, direction2, nnavi500, dokick):
         elif int(dokick) == 2:
             right_click()
 
+def previous_line(semi):
+    semi = str(semi)
+    Key("escape").execute()
+    time.sleep(0.05)
+    Key("end").execute()
+    time.sleep(0.05)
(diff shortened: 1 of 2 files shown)


kendonB:
thanks for this @dusty-phillips. Can you tell us a little bit about your use cases for this one?

dusty-phillips:
I use it in code editing when I am in the middle of a thought and realize that (for example) I haven't defined the variable that I'm about to use. I want to finish typing that line, add the `;` line terminator to that line, and then open a line above it to instantiate that variable. It's just a verbal shortcut for the key sequence "end semicolon home enter up."

Feel free to close if you don't think it's globally useful; I am happy to just maintain it locally.

LexiconCode:
I don't mind including it to globally as later this summer rewrite most of the commands for Caster based on the human machine grammar.

PR #806, opened 2020-05-01
content_request_generator handles special characters UTF-8 py3

kendonB:
# Trivial change

This one fixes a bug I was seeing when using python 3 with a user file containing special characters.

### DIFF
--- a/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
+++ b/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
@@ -32,7 +32,7 @@ class ContentRequestGenerator(object):
     def _get_file_lines(self, file_path):
         """File i/o broken out for testability"""
         content = None
-        with open(file_path) as f:
+        with open(file_path, encoding="UTF-8") as f:
             content = f.readlines()
         return content
 
--- a/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
+++ b/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
@@ -33,7 +33,7 @@ class ContentRequestGenerator(object):
     def _get_file_lines(self, file_path):
         """File i/o broken out for testability"""
         content = None
-        with open(file_path) as f:
+        with open(file_path, encoding="utf8") as f:
             content = f.readlines()
         return content
 
--- a/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
+++ b/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
@@ -1,6 +1,8 @@
 import re
 import os
-from io import open
+import six
+if six.PY2:
+    from io import open
 
 from castervoice.lib.ctrl.mgr.loading.load.content_request import ContentRequest
 from castervoice.lib.ctrl.mgr.loading.load.content_type import ContentType
--- a/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
+++ b/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
@@ -32,7 +32,7 @@ class ContentRequestGenerator(object):
     def _get_file_lines(self, file_path):
         """File i/o broken out for testability"""
         content = None
-        with open(file_path, encoding="UTF-8") as f:
+        with open(file_path) as f:
             content = f.readlines()
         return content
 
--- a/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
+++ b/castervoice/lib/ctrl/mgr/loading/load/content_request_generator.py
@@ -1,5 +1,6 @@
 import re
 import os
+from io import open
 
 from castervoice.lib.ctrl.mgr.loading.load.content_request import ContentRequest
 from castervoice.lib.ctrl.mgr.loading.load.content_type import ContentType


kendonB:
Sorry, I should have tested this with Python 2.7. Just did and it fails, unfortunately. Asking on the gitter what to do.

LexiconCode:
Troubleshooting encoding
https://www.azavea.com/blog/2014/03/24/solving-unicode-problems-in-python-2-7/

A similar question on stackoverflow
https://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2

I would recommend looking to see if `six` Which is a dedicated library for maintaining python 3/2 compatibility.
https://six.readthedocs.io/

kendonB:
the solution now is pretty clean. From the SO post. Thanks!

kendonB:
should be good now.

This one seems like we should be able to make a test to cover it. I will have a try

kendonB:
So the relevant test is here: https://github.com/dictation-toolbox/Caster/blob/04c6299e69f47c8608355808c48d59f5e77d0fba/tests/lib/ctrl/mgr/loading/load/test_contentRequestGenerator.py#L20L28

I tried just adding some special characters there but tests would still pass on master.

Do you think creating a test for this would have to rely on the testing object being a real file?

LexiconCode:
> 
> 
> So the relevant test is here: [/tests/lib/ctrl/mgr/loading/load/test_contentRequestGenerator.py@`04c6299`#L20L28](https://github.com/dictation-toolbox/Caster/blob/04c6299e69f47c8608355808c48d59f5e77d0fba/tests/lib/ctrl/mgr/loading/load/test_contentRequestGenerator.py#L20L28)
> 
> I tried just adding some special characters there but tests would still pass on master.
> 
> Do you think creating a test for this would have to rely on the testing object being a real file?

Typically with tests we don't read and write files. What you could do is figure out the The default encoding for `with open` and we can encode the test case string. Then use that String for testing purposes.

kendonB:
@LexiconCode so I tried doing something like what you're describing by doing this"

```
    def test_get_rule(self):
        self.crg._walk.side_effect = [[("/some/path", [], ["__init__.py", "something.pyc"]),
                                       ("/relevant/path", [], ["some_rule.py"])]]
        self.crg._get_file_lines.side_effect = [["from stuff import *",
                                                 "class Abc(MappingRule):",
                                                 "  mapping={u\"test āction\":NullAction()}",
                                                 "def get_rule():",
                                                 "  return Abc, RuleDetails(name=\"test\")"]]
        self._do_assertions(ContentType.GET_RULE, "some_rule", "Abc")
```

LexiconCode:
https://docs.python.org/3.7/library/functions.html#open
Looking into this a little further open() instead uses your systems environment to choose an appropriate encoding. I think that can be determined with `locale.getpreferredencoding()`. given that you are using a Windows system I suspect  `Windows-1252` character set which is overridden by setting encoding to utf-8 overrides this.

So I think you need to encode and decode with the appropriate character set.

bytes which you can decode to a string:
b"class Abc(MappingRule)".decode('windows-1252')

text string, which you can encode to a byte string
"class Abc(MappingRule)".encode('windows-1252')

kendonB:
@LexiconCode again, I'm not sure I follow. 

First, I think it's fine to assume that all text files that caster reads are in UTF-8. Looking at Notepad.exe, the default is UTF-8 now. Of course, this is true of other editors as well. 

The test should then just be testing that the input is correctly assumed to be UTF-8. The old behavior, which should fail, incorrectly assumed the input to be something else.

LexiconCode:
> 
> caster reads are in UTF-8. Looking at Notepad.exe, the default is UTF-8 now. Of course, this is true of other editors as well.

UTF-8 is the standard for most editors however it is not for operating systems especially Windows. 

Try executing snippet below, it should returns CP1252 on Windows systems (Windows-1252).
```
import locale
print (locale.getpreferredencoding())
```

Therefore `open()` all files read by caster utilize the default system/os encoding unless overridden.

```
self.crg._get_file_lines.side_effect = [["from stuff import *",
                                                 "class Abc(MappingRule):",
                                                 "  mapping={u\"test āction\":NullAction()}",
                                                 "def get_rule():",
                                                 "  return Abc, RuleDetails(name=\"test\")"]]
```

The reason The reason it passes with your tests the default encoding for strings are the following `ascii`(Py2) and `utf-8`(Py3). Both of which can handle accented characters although utf-8 is a better standard. 

Therefore If you really wanted to test for operating system your strings should be encoded with whatever `locale.getpreferredencoding()` provides. This is what `open()` opens/reads files as behind-the-scenes. Tests would allow assertions per operating system assertions whether it will pass or fail to support special/accented characters.

By using `encoding="utf8"` in `open(file_path, encoding="utf8")` it overrides the encoding provided by the OS obtained with `locale.getpreferredencoding()`. Therefore if we specify encoding no matter which operating system we should be okay for special characters.

Therefore thinking it through based on that I don't think we need tests at least for `open()` as `encoding="utf8"` sets the encoding which should solve any issues at that point in the code.

One suggestion is to use `six`. This will make it much easier to track down Python 2 compatibility code to be stripped out when Natlink is released with Python 3.
```
if six.PY2:
	from io import open
```
`from io import open` only needs to be imported is py2

kendonB:
@LexiconCode Made that final change and tests pass - merge if you think it works

LexiconCode:
> Mine works without making the config change when using `Text()` via `castervoice.lib.textformat.master_format_text`. Did you try this with `Text` and not just `Key`?

I don't recall it was a while back.  Since is not directly related I'll merge...

kendonB:
This is my maori grammar:

```
# -*- coding: UTF-8 -*-
'''
Created on May 20, 2019
@author: Kendon Bell
'''
from dragonfly import Function, Choice

from castervoice.lib import control
from castervoice.lib.actions import Key, Text
from castervoice.lib import textformat

maori_phrases = { 
    "(kyah order | kia ora)":                                        u"kia ora",
    "ngaa mihhy":                                                    u"ngā mihi",
    "(a wa teh de | Awatere)":                                       u"Awatere",
    "kaitiakitanga":                                                 u"kaitiakitanga",
    "(manaaki whenua | manaaki fenoo a)":                            u"Manaaki Whenua",
    "(tamaki makaurau | tah mucky Makoto | Auckland)":               u"Tāmaki Makaurau",
    "maori | multi-":                                                u"Māori",
    "Whakatipu Rawa | fakatipu dawa":                                u"Whakatipu Rawa", 
    "LTO door | aa o teah doah | Aotearoa":                          u"Aotearoa", 
    "tear nah koto car tour | tena koutou katoa":                    u"tēnā koutou katoa", 
    "rongo":                                                         u"Rongo", 
    "motu | mor too":                                                u"motu", 
}

macron_letters = {
    "mack arch":                                                          u"ā", 
    "mack eco-":                                                          u"ē", 
    "mack India":                                                         u"ī", 
    "mack oscar":                                                         u"ō",
    "mack Uniform":                                                       u"ū"
}

def maori_words(capitalization, spacing, phrase):
    if capitalization == 0:
        capitalization = 6
    textformat.master_format_text(capitalization, spacing, phrase)

def letter_macron(macron_letter):
    textformat.master_format_text(0, 0, macron_letter)

class Maori(MergeRule):
    pronunciation = "maori" 

    mapping = {
        "kupu [(<capitalization> <spacing> | <capitalization> | <spacing>)] <phrase>":
            R(Function(maori_words, extra={"capitalization", "spacing", "phrase"}) + Text(" "),
              rdescript="Maori: Insert phrase"),
        "<macron_letter>":
            R(Function(letter_macron, extra={"macron_letter"}),
              rdescript="Maori: Insert macron letter"),
    }

    extras = [
        Choice("capitalization", {
            "yell": 1,
            "tie": 2,
            "Gerrish": 3,
            "dock": 4,
            "laws": 5
        }),
        Choice(
            "spacing", {
                "flat": 1,
                "flat": 1,
                "spine": 2,
                "score": 3,
                "pebble": 4,
                "incline": 5,
                "dissent": 6,
                "descent": 6
            }),
        Choice("phrase", maori_phrases),
        Choice("macron_letter", macron_letters),
    ]
    defaults = {
        "capitalization": 0,
        "spacing": 0,
    }
    
def get_rule():
    details = RuleDetails(ccrtype=CCRType.GLOBAL)
    return Maori, details
```

LexiconCode:
So the reason that particular grammar works is because of the shebang `# -*- coding: UTF-8 -*-`

I'm trying to figure out a way to do that without having to specify in the file itself the encoding. Otherwise all files will require that at least with Python 2.7. Once Python 3 is adopted this should no longer be an issue as `UTF-8` is the default string encoding.

PR #813, opened 2020-05-11
Two minor tweaks for vscode grammar (should be trivial)

dusty-phillips:
1. Use the explicit indent and unindent VSCode commands
    instead of simulating with home + tab.

    The built-in commands work correctly with multiple cursors.

2. VSCode command to "skip next" the current cursor

    This behaves like "next cursor" except that it removes the
    current word from the selection.

    This allows you to use next cursor interactively
    such that you don't have to select all the words in a row.


dusty-phillips:
I updated the documentation to include the new command. I end up reformatting the table a bit to make the ordering better.

Issue #794, opened 2020-04-26
Several Vscode bindings are wrong


TTeig:
I've come across several commands in the built in grammar that don't work. I installed Code this year and did not change any bindings from the default. The main one I can remember is "join line".

(I'm happy to fix that particular one and any others I can find next time I have the time to go through and test them, but I'll leave this issue report in the meantime.)

I can run the join lines action from the command palette and it works fine, but it also shows up as not being bound to any keys. This is the binding in vscode2.py: `"join line": R(Key("csa-j")),` and naturally pressing those keys doesn't do anything either. I replaced it with this to fix it: `"join lines": R(Key("cs-p") + Text("join lines") + Key("enter")),`

**System:**
 Windows / Visual Studio Code v1.44.2

LexiconCode:
Yes feel free to open up a pull request and fix any others that you might find. Occasionally defaults to change over time.

PR #816, opened 2020-05-16
Fixing and adding some vscode commands

TTeig:
Some of the commands I thought were broken, just had very specific use cases or required extensions, but I made a couple of fixes.

Here's the original issue: #794


LexiconCode:
There are quite a few commits here that aren't relevant to the PR.

PR #817, opened 2020-05-16
"Remedy" Dialogue (request for comment)

dusty-phillips:
## Short version

Say *remedy in* and get the following dialogue:

image

Then say *Choose 5* and the word "in" will be replaced with the word "Inn," in title case.

## Long version
This is a first pass at a corrections dialogue. My goal was to pop up a window when the user says "remedy selection" and display possible corrections for the selected word.

There are two commands, `remedy selection` shows the corrections window for whatever word is actively selected. `remedy <text manipulation args> <dictation>`  uses the text_manipulation select functions to select the active word before showing the window. It has an added feature where if it can't find the selection that it appears, it searches for any alternative words that are index of similar sounding words. This allows you to say, for example `remedy to` and it will initiate correction for any of `to`, `too`, or `two` that it finds. This is a big advantage compared to the `replace` commands, which demand that the text you want to replace be exactly what dragon recognizes.

Currently, the suggestions include differing capitalizations, and words that sound the same using a list that I crafted from various lists of homophones on the web. that list lives in a new project [here](https://github.com/dusty-phillips/similar-sounding-words). I was expecting to try to do something that used the sentence and context with some natural language processing, but this basic index is way faster, and actually seems to be more useful.

This PR is built in QT and depends on Python 3. I'm not really interested in maintaining a Python 2 port.

I had to develop a different mechanism for communicating with the open window then the XML RPC tooling that is used in the mouse grid dialogues. It needs to be really snappy in order to be useful, and the cost of starting a full new Python process, initializing a TCP connection, and constructing a new window was taking simply too much time.

Instead, I depended on the multiprocessing library, which is rocksolid and lightning fast since Python 3.4. I open the process when Caster begins so that it is ready to go when I ask for a correction. I also have the cutie app and window initialized so all that needs to be done when the user requests a connection is put the requested choices into the dialogue and show it. The time between Caster recognizing the command in the window being actively displayed is more or less instantaneous. I actually suggest that a similar model could be used to speed up the most great commands.

The project is essentially split into four files. there are two rule files: A global one `text_manipulation_rules/corrections.py` that waits for the user to request a remedy, and a window-specific one `text_manipulation_rules/corrections_window.py` that is used for the user to initiate a choice. then there is a `asynch/corrections/manager.py` did is responsible for sending instructions to the subprocess that contains the window, and `async/corrections/process.py` which runs in the sub process to receive those commands.

## Open Questions
 - [ ] Currently, I am initializing the multi-processing spawner in the corrections manager module. This only needs to be done once, and if this model is used for other Windows, it should probably be done at a different Caster initialization step.
 - [ ] Is this something that you would want to include in Caster, or should I consider releasing it as a separate 'Caster plug-in' project? if you would want Python 2 support, I would probably choose to do that instead of maintaining it.
 - [ ] Are there better or additional ways to find related words? @Danesprite was looking into piping dragons related words through dragonfly, but we determined that Dragon's API wasn't returning very comprehensive results. I tried to do something with AutoCorrect libraries, but found that they weren't very useful, and I've been wondering about doing something with Markov chains.
 - [ ] Behaviour when there are multiple instances of similar sounding words in the the selection may be unexpected, but it's not clear what it should be. For example in the sentence, "She dragged the ore through the oar, or it would have been lost," If you say "Remedy leap second or", it's not clear which of "or," "oar", and "ore" you want to chang.
 - [ ] I have tested this pretty thoroughly with every manner of race condition an error condition I could think of, and I've been using it for daily use for a couple of days. But I don't think it's ready for prime time without some additional testing from other environments.
 - [ ] Speaking of other environments, I haven't tried it with Kaldi, WSR, or operating systems other than Windows 10.
 - [ ] If you want to accept this into Caster, I will write some documentation.

For me, this has already been a game changer, and made drafting, for example, this post much less frustrating than it would otherwise have been! ☺️

let me know if you have any ideas or suggestions. If you don't think it's a good fit for Caster (I know it's more useful for general dictation than for coding anyway), I am happy to maintain it as a separate package.

### DIFF
--- a/castervoice/asynch/corrections/manager.py
+++ b/castervoice/asynch/corrections/manager.py
@@ -1,0 +1,103 @@
+"""
+Manager for the corrections window. This module runs in the 
+Caster process. Its primary purpose is to initialize, manage, and communicate
+with the process that is actively displaying the window.
+
+The new process is initialized using multiprocessing.
+It is initialized as caster starts so there is no startup cost
+at the time the user asks to display the window.
+The window can be reused so the process never has to stop.
+"""
+import threading
+from castervoice.lib import control
+from castervoice.lib.context import paste_string_without_altering_clipboard
+from castervoice.lib import settings
+from dragonfly import Pause, Window
+from multiprocessing import get_context, Process, Pipe
+from .process import window_manager
+
+GRAMMAR_ENABLED = False
+
+def choice_grammar_enabled():
+    return GRAMMAR_ENABLED
+
+def set_enabled(enabled: bool):
+    global GRAMMAR_ENABLED
+    GRAMMAR_ENABLED = True
+
+# If the user hasn't made a selection in 15 seconds,
+# instruct the subprocess to hide the remedy window.
+WAIT_FOR_CHOICE_TIMEOUT = 15
+
+# Instruct the multi- processing library to use pythonw
+# for spawning tasks. Otherwise it tries to boot new copy of DNS.
+#
+# TODO: This shouldn't run at the script level
+
+# Caster probably has some initialization hooks to use instead
+# The setting needs only be set once, even if we use this
+# structure for other async commands (eg the mouse grids)
+pythonw = settings.settings(["paths", "PYTHONW"])
+get_context("spawn").set_executable(pythonw)
+
+# Start a new process initialized from the window_manager function
+# in the process module. All communication happens over a subprocess pipe.
+(PIPE, SUB_PIPE) = Pipe()
+WINDOW_PROCESS = Process(target=window_manager, args=(SUB_PIPE,))
+WINDOW_PROCESS.start()
+
+
+def send_choice(choice):
+    """
+    Instruct the corrections window process that a user choice has been selected.
+    """
+    PIPE.send(("SELECT CHOICE", choice))
+
+
+def cancel():
+    """
+    Instruct the corrections window process that the user has said "cancel"
+    So it can hide the window.
+    """
+    PIPE.send(("CANCEL", ""))
+
+
+def query_user_for_choice(selected_text):
+    """
+    Start a thread to show the corrections window and wait for the user to
+    select a choice.
+    """
+    threading.Thread(target=wait_for_choice, args=(selected_text,)).start()
+
+
+
+def wait_for_choice(selected_text):
+    """
+    Function running in a thread to tell the subprocess module to
+    show the window with the appropriate choices.
+    Then wait for the window to inform us which choice the user selected,
+    and replace the current selection with whatever they chose.
+    Thread ends if the user doesn't make a choice within a set number of seconds.
+    """
+    set_enabled(True)
+
+    while PIPE.poll():
+        # It is possible there are cancel messages in the pipe
+        # from the previous run. Remove them to get into a known state
+        # before showing the window again.
+        (message_type, message) = PIPE.recv()
+        if message_type != "CANCEL":
+            print("unexpected value in pipe")
+    PIPE.send(("SHOW WINDOW", selected_text))
+
+    if PIPE.poll(WAIT_FOR_CHOICE_TIMEOUT):
+        (message_type, message) = PIPE.recv()
+        print(f"Got a message {message_type}, {message}")
+        if message_type == "CHOICE":
+            paste_string_without_altering_clipboard(message)
+        elif message_type == "CANCEL":
+            print("Corrections dialogue closed by user")
+    else:
+        print("Timed out waiting for choice")
+        PIPE.send(("CANCEL", ""))
+    set_enabled(False)
--- a/castervoice/asynch/corrections/process.py
+++ b/castervoice/asynch/corrections/process.py
@@ -41,7 +41,7 @@ class App(QtCore.QObject):
     for instructions from the multiprocessing pipe.
     '''
     def __init__(self, pipe, parent=None):
-        super(self.__class__, self).__init__(parent)
+        super().__init__(parent)
 
         self.window = Window()
         self.pipe_loop = PipeLoop(pipe)
@@ -73,7 +73,7 @@ class PipeLoop(QtCore.QThread):
     hideWindow = QtCore.Signal()
 
     def __init__(self, pipe, parent=None):
-        super(self.__class__, self).__init__(parent)
+        super().__init__(parent)
         self.pipe = pipe
 
     def run(self):
--- a/castervoice/asynch/corrections/process.py
+++ b/castervoice/asynch/corrections/process.py
@@ -1,0 +1,214 @@
+'''
+This module lives in a subprocess and communicates with the main caster process
+over a multiprocessing pipe.
+
+It's primary purpose is to show and hide the window with a list of corrections
+and ensure that it contains the correct list of corrections depending on
+instructions sent from the manager module.
+
+the list of similar sounding words comes from an index that I created
+and curate in the similar_sounding_words python package.
+'''
+from PySide2 import QtWidgets, QtCore
+import similar_sounding_words
+from functools import partial
+
+try:
+    # Tip: because this module runs in the subprocess its logging output gets eaten
+    # `pip install q`
+    # then `tail -f /c/Users/<you>/AppData/Local/Temp/q` in `git-bash`
+    # See https://github.com/zestyping/q for more useful stuff
+    import q
+
+    print = q
+except ImportError:
+    pass
+
+def window_manager(pipe):
+    '''
+    Entry-point for the subprocess. Initializes a Qt application, but does not
+    show the window. This function is called at program startup.
+    '''
+    qtapp = QtWidgets.QApplication()
+    app = App(pipe, qtapp)
+    qtapp.exec_()
+
+
+
+class App(QtCore.QObject):
+    '''
+    Main QT application that sets up the window and the thread that listens
+    for instructions from the multiprocessing pipe.
+    '''
+    def __init__(self, pipe, parent=None):
+        super(self.__class__, self).__init__(parent)
+
+        self.window = Window()
+        self.pipe_loop = PipeLoop(pipe)
+        self.window.cancelClicked.connect(self.pipe_loop.cancel)
+        self.window.clickChoice.connect(self.pipe_loop.setChoice)
+        self.pipe_loop.setChoices.connect(self.window.setChoices)
+        self.pipe_loop.showWindow.connect(self.window.show)
+        self.pipe_loop.hideWindow.connect(self.window.hide)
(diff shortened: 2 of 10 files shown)


LexiconCode:
Looks like the beginnings of something great! I look forward to testing. Caster roots our for voice voding, but I want to continue to expand that to a general voice interface for PC computing. Which would include use cases like gaming and writing prose.

> Currently, I am initializing the multi-processing spawner in the corrections manager module. 

@dmakarov might have some thoughts on how to collaborate with GUI implementation.

> Is this something that you would want to include in Caster

Yes and I have no interest either in promoting python 2.7. If you wanted to it's trivial to convert the grammar caster to dragonfly.

LexiconCode:
>     * Are there better or additional ways to find related words?

It may not be comprehensive but is enough to provide whatever you do implement context to provide better alternatives as well? It looks like they've implemented as an engine call back which would make it easy to experiment see https://github.com/dictation-toolbox/dragonfly/pull/248

dusty-phillips:
Cool! I will try to write some documentation later this week to add to the PR. I want to experiment with the context stuff, but I think it should be in a separate pull request. In it will be quite a bit different in implementation, so it is probably better to reason about it separately.

dusty-phillips:
another feature that it would be cool to add is the ability for users to have their own commonly confused words to the library. I'm imagining a toml file in the settings directory that the user could customize with voice commands and a light GUI. But that should also go in a future pull request.

LexiconCode:
You might want to add `CorrectionsWindowRule` const.py `core` as it will enable the rule by default when settings are generated for the 1st time.
https://github.com/dictation-toolbox/Caster/blob/10327a53aa85826a2c8e03a641bbdd5a1d9623cc/castervoice/lib/const.py#L12

dusty-phillips:
Good call, I could also make the two rules companion rules maybe.

I usually get that error a couple times before it finds and focuses that window, but I have it retry every few tenths of a second, so it should get it eventually... If you manually click the remedy window into focus, do the choose commands work?

LexiconCode:
> If you manually click the remedy window into focus, do the choose commands work?

It does however I do have to manually `corrections window` by voice.

LexiconCode:
One other thought for a feature. Is if the remedy command is utilized without the dictation it would copy highlighted text. 

That would allow little more flexibility to scale with people's disabilities who can utilize mouse with limited interaction.

LexiconCode:
if you dictate `remedy window` the GUI pops up with empty choices


image

tieTYT:
I was asked to add this comment to this issue: I had a random thought about how to correct those tricky words
maybe you could make a command called "contraction". You could say "contraction they are"

dusty-phillips:
@tieTYT that's a good idea. A contraction command itself doesn't go with this PR, but I updated the similar-sounding-words list to include some common transactions. That would allow you to say, for example, "this sentence should have had a contraction in it" and get a box popping up that has "should've" is one of the choices.

dusty-phillips:
I made a small but notable change that both speeds up the remedy window and makes it more reliable:

In my original version, it showed the remedy window and then used a FocusWindow command to give the window focus. When the window had focus, the commands for "choose" were enabled because the window title matched the rule.

This led to weird race conditions when the window was closed, as the text didn't necessarily go where it was supposed to, either because the opened application didn't focus the text cursor where it was previously, or because the chosen text got "typed" into the remedy window instead.

In my new design, the corrections window rule is disabled by default, but is made global. Whenever the Remedy window is shown, it is shown as "always on top" but is not given focus. Additionally, the remedy window rule is programmatically enabled, so the choose commands are activated. They get deactivated again when the window closes.

There may be a couple timing issues as I haven't tested all scenarios yet, but I think this is a simpler design that will work correctly under more conditions.

LexiconCode:
As a last resort could utilize Nexus to enable and disable the grammar. However I just pushed the fix for FuncContext. FuncContext allows you to evaluate a function returning a bool to enable and disable a context.

```
from dragonfly import Function, Repeat, Choice, Dictation, MappingRule, Pause

from castervoice.lib.actions import Key, Text, Mouse

def testfunk():
    print ("Testfunk Evaluated")
    return True # Must return true or false Toggling on and off rule


# Testing for VS code
# I included extra imports above so you can easily switch between MappingRule and MergeRule
class TestFunkRule(MergeRule):
    pronunciation = "Visual Studio Test"

    mapping = {
        "test funk": R(Text("funk")), 
    }
    extras = [
    ]
    defaults = {}


def get_rule():
    return TestFunkRule, RuleDetails(executable="code", function_context = testfunk, ccrtype=CCRType.APP)
```

dusty-phillips:
I changed the command from "Remedy" to fix because the three syllables of "remedy" felt like too much to say for what I see as a common operation. Happy to change the command and set it up in my words.txt, though.

I'll take a look at funcContext later.

dusty-phillips:
Switched to function_contexts for enabling/disabling the grammar. I also rebased and squashed the whole chain of commits against latest master.

castervoice/asynch/corrections/manager.py
```diff
@@ -0,0 +1,103 @@
+"""
+Manager for the corrections window. This module runs in the 
+Caster process. Its primary purpose is to initialize, manage, and communicate
+with the process that is actively displaying the window.
+
```
dmakarov, line 5: Why not use the xmlrpc client/server mechanism that other caster asynch subprocesses use? This would make it easier to integrate remedy dialogue in a consolidated GUI process eventually, as opposed to having each dialogue/window running a separate GUI application.

dusty-phillips: I originally implemented it that way, but it was way too slow. XMLRPC + a TCP Server is much much slower than Python pickle v5 + a system pipe. Actually, I think I'm still on pickle v4 in this PR, because v5 has to be explicitly enabled, but v4 is super fast, too. v5 adds some interesting shared memory features, but I haven't researched how it works.

I had a similar idea to port the other GUI windows to a single application, but I recommend putting them behind multiprocessing. Under the hood, multiprocessing does the same thing as the other windows (spins up a separate Python process, at least on Windows; on Unix systems it uses a fork, which is much faster), but it manages all the weird edge cases for you. 

As described in the PR comment, another thing I did in this PR was to load the Window once, and keep it hidden in the background instead of constructing it from scratch on every command. This gave a modest speed boost, but for something like Rainbow Grid, it would be a huge improvement. On my 4K screens, Ranbow takes seconds to load and is therefore unusable. If the window was preloaded, it would snap up instantly.

dmakarov: > I originally implemented it that way, but it was way too slow. XMLRPC + a TCP Server is much much slower than Python pickle v5 + a system pipe. Actually, I think I'm still on pickle v4 in this PR, because v5 has to be explicitly enabled, but v4 is super fast, too. v5 adds some interesting shared memory features, but I haven't researched how it works.

Are other caster asynch apps also slow on your end? It is surprising to me, as I never observed noticeable delays by xmlrpc.

dusty-phillips: The various mouse navigation windows are unusable for me, although most of that is from rendering the windows, which takes between 1 and 5 seconds depending on which grid is being used; rainbow is longest.

In my profiling for the remedy window (I didn't keep numbers, sorry) the time delays were in this order:

 *  Time it takes to start up the interpreter
 *  Time it takes to render and display the window
 *  Time it takes for Dragon to recognize the command
 *  Time it takes to send commands

I optimized each in turn as I've described, other than the Dragon recognition which is outside my control (switching from Normal mode to Command mode helps, but that interfered with my dictation as I kept having to remember whether to insert an "ace" after each "say", "cop" or "slip"). The biggest win on sending commands was reducing the size and number of commands (calculating the choices in the subprocess instead of passing them across the pipe, for example) more than changing the transport, but since I'd already done that part I left it in. If you leave the other optimizations in place, switching it back to XMLRpc should only be a few minutes work as they are both just bidirectional pipes.

Something like the corrections window needs to be instantly responsive or else it interrupts concentration too much to be useful. It's still a little too slow for my taste, especially when it comes to hiding the window after selecting the choice. I *think* I've got it to the point where the highest percentage of the time is consumed by Dragon deciding whether it recognized a command, but it may be worth doing some finer grained profiling and compiling certain pieces of it with Cython.

The main drawback of multiprocessing is that it explicitly depends on Python. Your fastest option would be to write the GUI half of the applications in a low-level language like C++ or Rust and then use an IPC pipe to communicate between the two using a compact protocol such as CBOR. But I don't think that will give enough of an improvement to justify the maintenance overhead, and the time would probably be better spent profiling and optimizing natlink or finding some way to speed Dragon itself up.

Overall, I've been pretty disappointed in the corrections window experiment. It does a pretty good job of offering useful corrections, but even fully optimized, it interrupts my flow enough that I lose a minute of concentration for every mistake Dragon makes. This seems to be an unfixable function of the way my brain works, and I've been recently investing my time in ways to get my typing back instead. Spending a few hours coding the corrections window would probably have been much better spent doing a lot of hand and forearm yoga!

LexiconCode: Okay why not try it this way. As an option have the Remedy dialogue box stay in the forefront but not in focus. That way can sit in the corner of the screen and update as you work with dictation. That would that would reduce `render and display the window`.

dusty-phillips: That'd be brilliant; something like how the Serenade window works. I'm imagining that it normally shows just the alternate interpretations the underlying speech engine gives, perhaps with some different capitalization options. But if the user says "fix <phrase>" it would drop into something like this version.

I had poked a bit at the alternate phrasing diff on the dragonfly grammar, but I couldn't figure out how to patch it through. Definitely worth it, though.

LexiconCode: > I couldn't figure out how to patch it through

Patching it through in terms of updating a grammar during runtime or updating the GUI?

dusty-phillips: of where to access the grammar value from within a rule.

LexiconCode: Okay I think I know what you're saying. You have a variable that's defined in a grammar. It's only evaluated once during rule initialization. Therefore you can't access the updated value in the grammar because it has not been evaluated.

So rather than doing the standard `Function(myfunction)`use a lambda.
`Function(lambda n: say_number(n))`

kendonB:
@dusty-phillips I tried testing this and it didn't load. I'm using Kaldi and Python 3

LexiconCode:
> 
> 
> @dusty-phillips I tried testing this and it didn't load. I'm using Kaldi and Python 3

Did you install through PIP `similar-sounding-words`?

Issue #798, opened 2020-04-29
Various thoughts about function context


mpourmpoulis:
## Disclaimer

The is in continuation of the chat in the gitter channel. If I have misunderstood something, please feel free to correct me!

## Files Involved

the main files involved here are

- [ccrmerger2](https://github.com/dictation-toolbox/Caster/blob/master/castervoice/lib/merge/ccrmerging2/ccrmerger2.py)

- [context from dragonfly](https://github.com/dictation-toolbox/dragonfly/blob/master/dragonfly/grammar/context.py)

## Where Contexts Get Created

unlike previous versions of Custer, instead of creating the context of our selves, we instead provide rule details object and castor creates the context for us. 

for non-CCR grammars, this is handled by the mapping_rule_maker with the following bit of code

```python
        context = None
        if details.function_context is not None:
            context = FuncContext(function=details.function_context, executable=details.executable, title=details.title)
        else:
            if details.executable is not None or details.title is not None:
                context = AppContext(executable=details.executable, title=details.title)

```

For CCR grammars these happens right here

https://github.com/dictation-toolbox/Caster/blob/d8f696284cf3df7f5161a3ba4a64c6f12b55f608/castervoice/lib/merge/ccrmerging2/ccrmerger2.py#L136-L168

an important thing to note here is that all of these happens long after they rule details have been validated, so any restrictions imposed there will have no effect from now on!

## The Lost Executable

one of the first things to notice is that the function context constructor does not only receive the callable function we specified, but also the executable and title parameters we included in the rule details. At first glance these can have users thinking that by including an executable, the function context really evaluates to true only if both their function evaluates to true and the executable matches the foreground window. So for instance,@davidlehub shared the following example

```python
def get_rule():
    return notepad,RuleDetails( executable="notepad", function_context = f2, ccrtype=CCRType.APP)
```

which unfortunately does not work as expected! if we take a quick look at the part of the constructor

https://github.com/dictation-toolbox/dragonfly/blob/7b9118f66b667bc474c740ff4f0617f700ee97b1/dragonfly/grammar/context.py#L377-L391

we see that they are treated as double star keywords arguments, to be fed to our function as default values if it has a Keyword Parameter named `executable` or `title`. But they either way do not make their, as dragonfly overrides them with their respective values for the current foreground window(line 400)

[...]

PR #820, opened 2020-05-25
Fix dragonfly `FuncContext` Caster App integration

LexiconCode:
# Title
Fix FuncContext 

## Description

Changes how Caster creates contexts for CCR/Mapping for applications context. Note not global context.

## Related Issue

Background explanation with the bug #798
An amazing community! A big thank you goes to @mpourmpoulis with that right up and insights with this issue.

## Motivation and Context

This PR fixes FuncContext functions from being evaluated twice, outside of context, and context ignoring the exe/ttitle.

## How Has This Been Tested
This is been tested with app mapping and mergerules(CCR)
- Out of context - FuncContext is not evaluated and fails when set to False or True
- In context - FuncContext is evaluated once and Fails when set to False
- In context - FuncContext is evaluated once and Succeeds when set to True

Test Rule
```
from dragonfly import Function, Repeat, Choice, Dictation, MappingRule, Pause

from castervoice.lib.actions import Key, Text, Mouse

def testfunk():
    print ("Testfunk Evaluated")
    return True # Must return true or false


# Testing for VS code
# I included extra imports above so you can easily switch between MappingRule and MergeRule
class TestFunkRule(MergeRule):
    pronunciation = "Visual Studio Test"

    mapping = {
        "test funk": R(Text("funk")), 
    }
    extras = [
    ]
    defaults = {}


def get_rule():
    return TestFunkRule, RuleDetails(executable="code", function_context = testfunk, ccrtype=CCRType.APP) 

```

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/castervoice/lib/ctrl/mgr/rule_maker/mapping_rule_maker.py
+++ b/castervoice/lib/ctrl/mgr/rule_maker/mapping_rule_maker.py
@@ -25,7 +25,7 @@ class MappingRuleMaker(BaseRuleMaker):
 
         context = None
         if details.function_context is not None:
-            context = FuncContext(function=details.function_context, executable=details.executable, title=details.title)
+            context = AppContext(executable=details.executable, title=details.title) & FuncContext(function=details.function_context)
         else:
             if details.executable is not None or details.title is not None:
                 context = AppContext(executable=details.executable, title=details.title)
--- a/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
+++ b/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
@@ -150,11 +150,14 @@ class CCRMerger2(object):
         negation_context = None
         for cr in app_crs:
             details = rcns_to_details[cr.rule_class_name()]
+            context = AppContext(executable=details.executable, title=details.title)
             if details.function_context is not None:
-                context = FuncContext(function=details.function_context, executable=details.executable, title=details.title)
+                funkcontext = context
+                funkcontext &= FuncContext(function=details.function_context)
+                contexts.append(funkcontext)
             else:
-                context = AppContext(executable=details.executable, title=details.title)
-            contexts.append(context)
+                contexts.append(context)
+
             if negation_context is None:
                 negation_context = ~context
             else:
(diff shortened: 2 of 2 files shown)


LexiconCode:
Thanks for documenting the issue. To be clear about the replication procedure

1. To replicate have FunkContext Function return `false` 
2.  Enable grammar that shares the same context (Python or LatexMath that global context)
3. Disable (Python or LatexMath)
4. Observe that the global grammars (Like alphabet) context the inactive and not working as expected.

mpourmpoulis:
This seems to be working On my end with py2

```python
    @staticmethod
    def _create_contexts(app_crs, rcns_to_details):
        """
        Returns a list of contexts, AppContexts based on 'executable', FuncContext, one for each
        app rule, and if more than zero app rules, the negation context for the
        global ccr rule. (Global rule should be active when none of the other
        contexts are.) If there are zero app rules, [None] will be returned
        so the result can be zipped.

        :param app_crs: list of CompatibilityResult for app rules
        :param rcns_to_details: map of {rule class name: rule details}
        :return:
        """
        contexts = []
        context_evaluations = {} 

        def wrap_context(context):
            old_matches = context.matches
            context_evaluations[context] = (False,False)
            def matches(executable,title,handle):
                result,valid = context_evaluations[context]
                if valid:
                    context_evaluations[context] = (result,False)
                else:
                    result = old_matches(executable,title,handle)
                    context_evaluations[context] = (result,True)
                return result
            context.matches = matches
            return context

        for cr in app_crs:
            details = rcns_to_details[cr.rule_class_name()]
            context = AppContext(executable=details.executable, title=details.title)
            if details.function_context is not None:
                context &= FuncContext(function=details.function_context)
            contexts.append(wrap_context(context))

        negation_context = FuncContext(lambda **kw:[x for x in context_evaluations if x.matches(**kw)]==[])


        contexts.insert(0, negation_context)
        return contexts
```

The function context gets evaluated only once And no DOS

LexiconCode:
> 
> This seems to be working On my end with py2
> 

I will test this on Python 3. Can you verify the procedure to produce DOS in the above post?

mpourmpoulis:
yes that is the procedure

My only comment might be the step two and three are mostly for emphasis, You should get the same results if you boot castor With the rule disabled from the start

LexiconCode:
This seems to work on Python 3, do you want to open up a PR?

PR #822, opened 2020-06-04
Restructure ccr context creation to fix problems with function contexts without ignoring them from the computation of the negation context

mpourmpoulis:
# Title

## Description

This is essentially part two of #820 that's tried to address #798 And took care of most of the work needed but left a couple of details open, like excluding function contexts from the computation of the negation context, In order to avoid an edge case that could leave the user without global commands. This problem is now taken care off without sometimes leaving two merged CCR's active

This is achieved by tweaking the matches function of context created to cache It's result So it can be reused precisely one more time! as a consequence, the context gets evaluated Only once even though it is used both for its own grammar as well as for negation context. Furthermore, the implementation does not depend on the order in which the contexts are being processed

## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have added tests to cover my changes.

### DIFF
--- a/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
+++ b/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
@@ -1,3 +1,5 @@
+import traceback
+
 from dragonfly.grammar.elements import RuleRef, Alternative, Repetition
 from dragonfly.grammar.rule_compound import CompoundRule
 from dragonfly import FuncContext
@@ -157,7 +159,11 @@ class CCRMerger2(object):
                 if valid:
                     context_evaluations[context] = (result,False)
                 else:
-                    result = old_matches(executable,title,handle)
+                    try : 
+                        result = old_matches(executable,title,handle)
+                    except :
+                        results  = True
+                        traceback.print_exc()
                     context_evaluations[context] = (result,True)
                 return result
             context.matches = matches
--- a/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
+++ b/castervoice/lib/merge/ccrmerging2/ccrmerger2.py
@@ -147,21 +147,32 @@ class CCRMerger2(object):
         :return:
         """
         contexts = []
-        negation_context = None
+        context_evaluations = {} 
+
+        def wrap_context(context):
+            old_matches = context.matches
+            context_evaluations[context] = (False,False)
+            def matches(executable,title,handle):
+                result,valid = context_evaluations[context]
+                if valid:
+                    context_evaluations[context] = (result,False)
+                else:
+                    result = old_matches(executable,title,handle)
+                    context_evaluations[context] = (result,True)
+                return result
+            context.matches = matches
+            return context
+
         for cr in app_crs:
             details = rcns_to_details[cr.rule_class_name()]
             context = AppContext(executable=details.executable, title=details.title)
             if details.function_context is not None:
-                funkcontext = context
-                funkcontext &= FuncContext(function=details.function_context)
-                contexts.append(funkcontext)
-            else:
-                contexts.append(context)
-            if details.function_context is None:
-                if negation_context is None:
-                    negation_context = ~context
-                else:
-                    negation_context &= ~context
+                context &= FuncContext(function=details.function_context)
+            contexts.append(wrap_context(context))
+
+        negation_context = FuncContext(lambda **kw:[x for x in context_evaluations if x.matches(**kw)]==[])
+
+
         contexts.insert(0, negation_context)
         return contexts
 
--- a/tests/lib/merge/ccrmerging2/test_CCRMerger2.py
+++ b/tests/lib/merge/ccrmerging2/test_CCRMerger2.py
@@ -28,7 +28,7 @@ class TestCCRMerger2(SettingsEnabledTestCase):
     def _evaluate_context_in_every_permutation(self,contexts,expected_output,data = {}):
         for permutation in permutations(zip(contexts,expected_output)):
             expected = [x[1] for x in permutation]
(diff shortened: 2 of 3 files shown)


LexiconCode:
The following unit tests are failing. Primarily they are failing because they expect no context for global rules or LogicNotContext. In every failure FuncContext is received  instead of the original assertion. We need to validate the logic behind the unit tests in relation to code additions. 

- test_merge_nonempty: Merger successfully "merge" one CCR rule w/ no context. 

	The unit test fails because it expects `self.assertIsNone(context)` an None context but receives FuncContext.

-  test_merge_one_context_one_no_context:  One contexted CCR rule and one non-contexted mergerule should result in two RepeatRules, the contexted one with a Context aligned with it and the non-contexted one with a LogicNotContext aligned with it.

	The test fails because it expects LogicNotContext and receives FuncContext instead. 

- test_merge_two: Merger successfully merges two CCR rules w/ no context.

	The unit test fails because it expects `self.assertIsNone(context)` an None context but receives FuncContext.

- test_merge_two_app_ccr: Merger successfully merges one global + two CCR app rules.

	The test fails because it expects LogicNotContext and receives FuncContext instead.

mpourmpoulis:
Whoops :) I believe A quick solution so as to make those unit tests stop crashing is the simple  modification

```python
if context_evaluations:
    negation_context = ~FuncContext(lambda **kw:[x for x in context_evaluations if x.matches(**kw)]!=[])
else:
    negation_context = None

```

Or at least some of them, it probably will not fix the logic and unit test

But I guess the more important point is to clear out why a function context  is used in the first place Instead of simply negating all contexts we have wrapped with our function replacing their matches method in taking the logical and of them

The reason was logical and  and logical or contexts working a lazy manner

https://github.com/dictation-toolbox/dragonfly/blob/b3f82fe29f434bd8650a0e8e7566b62bbeed0b13/dragonfly/grammar/context.py#L181-L186

Which means that sometimes not all of the contexts Are going to get evaluated When computing the negation context ( for instance if you have C1=True,C2 = False,C3 = True When evaluating c1&&c2&&c3 The computation is going to stop at c2 because No matter what follows it the final result is going to be the same = False)

Now regardless of what implementation would choose, if we use caching we need to make sure that the precomputed values are only valid for the current utterance for obvious reasons! one way to achieve this is using recognition observers and register A call back that will invalidate the cache every time space is detected(on_process_begin if not I am not mistaken) But

- I feel this is too much of an invasive change In the code 

- I am not sure what is going to happen with  Mimic ( which also evaluates contexts)

So to avoid all that, In a way that avoids setting restrictions in the order it which the contexts  are going to get evaluated I have chosen for an implementation That guarantees that we will check the value of every context precisely 2 times for every utterance( the first time we compute it The second we use the precomputed value) And such an implementation is impossible with the laziness of the logical and context, hence a custom function: context is used

LexiconCode:
> Whoops :) I believe A quick solution so as to make those unit tests stop crashing is the simple modification
> 

Great that fixes everything but LogicNotContext which you expected.
  
> The reason was logical and and logical or contexts working a lazy manner
> 
> [dictation-toolbox/dragonfly:dragonfly/grammar/context.py@`b3f82fe`#L181-L186](https://github.com/dictation-toolbox/dragonfly/blob/b3f82fe29f434bd8650a0e8e7566b62bbeed0b13/dragonfly/grammar/context.py#L181-L186)


Would be an enhancement to Dragonfly by tweaking `def matches` to make the context managing more robust? I could get Danesprite's insights on this particular issue. On the other hand we could continue forward with the custom function.

mpourmpoulis:
in general change in the laziness might be unpreferable For several reasons

- There is probably already user code that depends on it and will break

- A great deal many of programming languages allow for  at least some degree of laziness when evaluating conditions (including python) So dragonfly users might Intuitively expect these to hold

- It is going to break little trick of using `app&fun` for function contexts that also need an executable, Because while with the current implementation and the laziness If the application in the foreground does not match the context the function will never get evaluated, without the laziness it would proceed to evaluate the function regardless of whether the application was What we wanted

Unless perhaps we could set a special flag For laziness that could default To the current behavior and we could flip it just for these particular use case But I'm not sure this is worth going all the trouble with messing up with dragonfly. End of the day we can do that by replacing the matches function of the negation context only, in a monitor similar to what we do for the other contexts.

LexiconCode:
> End of the day we can do that by replacing the matches function of the negation context only, in a monitor similar to what we do for the other contexts.

Those are some good points and clarifies moving forward with the custom function and updating the unit test to reflect the change.

mpourmpoulis:
A couple of notes

- Firstly, I will see to updating those unit tests But I would prefer to do more than simply patching them. In particular I would like to add some unit tests that are going to evaluate contexts instead of only checking their types in order to make sure their runime behavior  is as desired (For instance this will allow us to properly verify that the function contexts are only evaluated once) 

- I realize there is a flaw in my line of thought, as I have not taken care off the possibility of exceptions being raised. With plain dragonfly this causes the problem The grammars become all of a sudden global:) but with this implementation, if an exception is raised while trying to evaluate One of the contexts for the negation context, it is going to proceed no further And all the others after it are going to get called only once, Which cause the same problem as the laziness we discussed above! I need to add try-except blocks inside the replacement matches function To make sure this does not happen but also print out the error message to the user so that he is aware there is a problem

LexiconCode:
@mpourmpoulis do you think there be a possibility of resolving a few of these issues this coming week?

mpourmpoulis:
ok I took a look at it again and and I can confirm we are good to merge!

LexiconCode:
I've also been experimenting with it last couple days and haven't come across any issues. Amazing work! thank you!

PR #828, opened 2020-07-05
Add active python/pip info in install scripts

daanzu:
# Trivial change

## Description

Seems like this could be helpful for discovering/debugging python issues. Feel free to close if not desired.


LexiconCode:
Thank you for this! The added info is a good addition to the install process to help the user.

daanzu:
I had also wondered about adding similar output to the run scripts, but wasn't sure if it would be too noisy.

PR #852, opened 2020-09-27
add six to requirements (pretty trivial)

kendonB:
Found that this stopped caster working after a reinstall.

Current version is 1.15, do we want to specify version greater than that?

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -8,3 +8,4 @@ appdirs>=1.4.3
 scandir>=1.10.0
 pylint
 pyvda
+six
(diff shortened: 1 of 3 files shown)


LexiconCode:
> 
> 
> Found that this stopped caster working after a reinstall.
> 
> Current version is 1.15, do we want to specify version greater than that?

Yes I will work on land soon as I get a few things merged.

kendonB:
@LexiconCode so I don't actually think this fixes the bug as I saw it say it didn't have six after installing it.

The bug occurs because the pythonw setting in `settings.toml` can point to the installation of python 2.7. When you call the mouse grids, caster will then use this installation of python which then may not have the `six` package installed. So I think we have to try and ensure that the correct version of python gets put into pythonw setting in `settings.toml` in the first place.

LexiconCode:
Experiment with the following function within settings and see what returns.

```
`def _get_platform_information():`
```

Issue #797, opened 2020-04-28
add a kaldi rule (at least for sleep/wake)


kendonB:
**Is your feature request related to a problem? Please describe.**
Dragon has a bunch of built in commands that help with use. It would be nice for switchers to have a Kaldi rule loosely based on the functionality available in base Dragon.

**Describe the solution you'd like**
A Kaldi grammar with the following features:
 - [ ] Sleep/wake

Just the one for now but others may be appropriate especially as KAG adds functionality.

To add sleep/wake we need to translate the grammar in [this file](https://github.com/dictation-toolbox/dragonfly/blob/master/dragonfly/examples/kaldi_module_loader_plus.py) into caster.

@daanzu if you have any pointers or know of someone who has done this please let us know.

LexiconCode:
This potential isn't too hard to implement. This would be relevant to all engines not just Kaldi. How Kaldi  manages to make this work `set_exclusiveness()`. An exclusive grammar takes precedent over all other active grammars. A Rule/grammars that are exclusive are only available for recognition.

Note setting a grammar to exclusive overrides DNS's built-in sleep-wake function. When using DNS "natlink.setMicState("sleeping")" setting to match the state of the grammar.

The FunkContext with a mapping rule could be used when sleep state is enabled. The enabling command could be housed within `utility_rules\caster_rule.py`

lexxish:
> This potential isn't too hard to implement. This would be relevant to all engines not just Kaldi. How Kaldi manages to make this work `set_exclusiveness()`. An exclusive grammar takes precedent over all other active grammars. A Rule/grammars that are exclusive are only available for recognition.
> 
> Note setting a grammar to exclusive overrides DNS's built-in sleep-wake function. When using DNS "natlink.setMicState("sleeping")" setting to match the state of the grammar.
> 
> The FunkContext with a mapping rule.

Overriding the DNS default commands for sleep/wake would be nice because you can use shorter commands such as "snore" to put the mic to sleep.

Kaldi implementation is more important in my opinion but, to use this idea for Natlink it can be done with something like:
```
def snore():
    import natlink
    engine = get_engine()
    engine.speak("Sleeping")
    natlink.setMicState("sleeping")

def wake():
    import natlink
    natlink.setMicState("on")
    engine = get_engine()
    engine.speak("Awake")
```

LexiconCode:
> Only downside is I'm not sure if we could get the taskbar icon toggling from green to blue. 

[...]

PR #853, opened 2020-10-04
Implemented microphone modes

LexiconCode:
## Description

Implemented microphone modes to toggle microphone on, off, and sleep.

The new commands
`caster <mode>` on, sleep, off (DNS/DPI only)

These modes use grammar/rule exclusivity to block other grammars from being recognized. 
- Kaldi uses still recognize dictation but no action will be executed. The same is true for DNS/DPI but it's not printed out the window.
- DNS/DPI is also a bit special if you change the mic mode from the DNS/DPI GUI it will stay in sync with Caster mic modes
- DNS/DPI built-in commands for waking up and going to sleep are still functional as they rely on exclusivity internally. There is no method that I know of to disable these internal commands. 

These modes are implemented through 2 rules. The 1st rule `CasterMicRule` to execute the function to enable exclusivity. The 2nd rule `sleep_rule` leverages native dragonfly to become exclusive via `ExclusiveManager` overriding `CasterMicRule` and all other rules loaded into the engine. `sleep_rule` uses dragonfly `FuncContext` to toggle on and off. Both rules use `EngineModesManager` to manage the different modes based on the engine is running. There are some significant differences. DNS utilizes Natlink and `_sync_mode` to synchronize modes then uses ExclusiveManager. The other engines utilize just ExclusiveManager. 

**Note** because native dragonfly was used for `sleep_rule` simplified transformers cannot be used on `CasterMicRule` or `sleep_rule`.

Could this implementation be simplified?
 If we could create separate CCR merger that's for exclusive grammars only it would drastically simplify the implementation.

What about additional modes?
These are some thoughts for the future but not in this PR. I did not want to take longer to get the sleep mode out to users as it impacts the usability of Caster.
- Implement set_engine_mode exclusivity with mode rules. dictation, numbers, and spell for all engines
- Implement timer for sleep mode for all engines but DNS/DPI.
- Implement hotkey for microphone on-off

## Related Issue

https://github.com/dictation-toolbox/Caster/issues/797

## Motivation and Context

Motivation for this is to unify microphone modes command regardless of the speech recognition backend for users.

## How Has This Been Tested

I've tested this with DNS and Kaldi switching between microphone modes.

## Types of changes

## Checklist


castervoice/rules/core/engine_manager_rules/mic_rules.py
```diff
@@ -0,0 +1,24 @@
...
+from castervoice.lib.ctrl.mgr.engine_manager import EngineModesManager
+
+class CasterMicRule(MappingRule):
+    mapping = {
+        "caster <mic_mode>": Function(lambda mic_mode: EngineModesManager.set_mic_mode(mode=mic_mode)),     
+    }
+    extras = [
+        Choice(
+            "mic_mode", {
+                "off": "off",
```
kendonB, line 16: I get this when saying "caster off"

`Caster: 'off' is not valid. set_mic_state modes are: 'off' - DPI Only, 'on', 'sleeping'`

Is that deliberate? We should also have "caster wake" as an option.

kendonB: @LexiconCode thanks for doing this!

LexiconCode: Yes it is intentional ('off' - DPI Only). For other engines besides DPI/DNS to have an off state they need a method to turn the microphone on without voice activation like a keyboard shortcut/GUI toggle. DPI/DNS has that capability.

esc123:
Using DPI 15.3 on Windows 10, all three commands (on, off, sleep) work perfectly. Good stuff.

kendonB:
@LexiconCode is there an easy way to implement automatic sleeping functionality? I often forget that the system is on after using it for a few minutes and end up making it do a bunch of crazy things because I'm speaking around it or listening to something.

LexiconCode:
> @LexiconCode is there an easy way to implement automatic sleeping functionality? I often forget that the system is on after using it for a few minutes and end up making it do a bunch of crazy things because I'm speaking around it or listening to something.

Yes I think that shouldn't be too bad to do. It would involve creating a timer that is reset every time there's a successful recognition.

PR #859, opened 2020-10-17
Add new keyboard emulator commands 

kendonB:
# Add new keyboard emulator commands 

## Description

This PR implements a new grammar that allows the user to press any key, potentially with a modifier (control, alt, shift, windows/super). It also gives the user the ability to hold and release and key on the keyboard.

It also consolidates some of the commands in `nav.py` that are individual key presses (e.g. clear). It also removes what would be duplicate keyboard emulation functionality from `nav.py`.

The breaking changes: 
 - the F1-12, alphabet, and punctuation key presses are no longer repeatable. So you can no longer say "hit F3 7" to press F3 7 times. 
 - It also changes the spec for press to be "press" not "hit" to be consistent with Dragon.

## Related Issue

PR #850

## Motivation and Context

The ability to press many keys on the keyboard (including the modifier keys themselves) was missing from Caster (all are available natively in Dragon using the prefix "press"). Holding and releasing is sometimes desired for keys as well.

@LexiconCode I'm finding that the performance hasn't really changed, so I'm happy to push ahead with this as it is. It is probably worth someone testing on an older setup.

## How Has This Been Tested

Tested using Kaldi. It is likely that I haven't tested every single combination so some help in testing this further would be useful.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

@LexiconCode please review before I do the docs changes


kendonB:
I will need to work out what's breaking the tests

LexiconCode:
This is looking great! I will get around to testing this today and tomorrow.

2 things
Does this need to be CCR, this end-user really lose that much functionality?
Does this need to be enabled by default?

Out of those 2 questions I have a strong preference to leave it disabled by default for the following rationale.
The purpose of keyboard grammar is either to implement support for an application that does not have a grammar or
a grammar that is missing is command. Both of which suggest that updates to a grammar or the creation of grammar should be made for proper control. Speaking keyboard keys is really inefficient and that's the last thing I would like users to do on a day-to-day basis if we can help it. Users can then choose whether or not to have it enabled. 

Perhaps with user feedback we can decide later mapping versus CCR?
kendonB If you're finding yourself using this a lot where you're using it and what applications?

castervoice/rules/core/keyboard_rules/keyboard.py
```diff
@@ -0,0 +1,147 @@
...
+        "pause": "pause",
+        windows_spec: "win",
+        "(apps | popup)": "apps",
+        "print screen": "printscreen",
+        "scroll lock": "scrolllock",
+        "num lock": "numlock",
+        "caps lock": "capslock",
+        "(home | lease wally | latch)": "home",
+        "(end | ross wally | ratch)": "end",
+        # number pad numbers deliberately left off
```
LexiconCode, line 107: I definitely agree with these choices here. We should continue to elevate those things into domain specific use cases or like applications or high-level grammars.

castervoice/rules/core/keyboard_rules/keyboard.py
```diff
@@ -0,0 +1,147 @@
...
+    button_dictionary_1 = {
+        "(F{}".format(i) + " | function {})".format(i) : "f{}".format(i)
+        for i in range(1, 13)
+    }
+    button_dictionary_1.update(caster_alphabet())
+    button_dictionary_1.update(_tpd)
+    longhand_punctuation_names = {
+        "minus": "hyphen",
+        "hyphen": "hyphen",
+        "comma": "comma",
```
LexiconCode, line 55: In order for natlink to recognize comma symbol the following is needed
```
if (dragonfly.get_current_engine().name == 'natlink'):
    comma = "(comma | ,)"
else:
    comma = "comma"
```
`comma: ",",`

kendonB:
> Does this need to be CCR, this end-user really lose that much functionality?

This would definitely be low priority for CCR. Making it a `MappingRule` would mean another breaking change (I have edited the above to spell out the current breaking changes.) but I don't recall using this as CCR often.

Does this need to be enabled by default?

> I think yes. We're never going to be able to cover every app out of the box and I think that Caster should have as much command functionality as plain Dragon. I'd much rather keep it default and make it nonCCR than make it an extra grammar.

LexiconCode:
> 
> 
> > Does this need to be CCR, this end-user really lose that much functionality?
> 
> This would definitely be low priority for CCR. Making it a `MappingRule` would mean another breaking change (I have edited the above to spell out the current breaking changes.) but I don't recall using this as CCR often.
> 
> > Does this need to be enabled by default?
> 
> I vote yes. We're never going to be able to cover every app out of the box and I think that Caster should have as much command functionality as plain Dragon. I'd much rather keep it default and make it nonCCR than make it an extra grammar.

Okay let's keep it enabled by default ass non-CCR. We can switch to CCR in the future without breaking changes if you leave the imports for CCR in place in the grammar.

kendonB:
@LexiconCode I have removed those extra specs associated with a few of the punctuation keys. Those can be swapped using words.txt if users wish

kendonB:
@LexiconCode do you have insight into why the python 2.7 tests fail? I can replicate the failure locally but I'm not sure what's causing it

castervoice/rules/core/navigation_rules/navigation_support.py
```diff
@@ -16,8 +16,11 @@ def get_direction_choice(name):
 
  # Insurers comma is recognized consistently with DNS/Natlink
  # if/else statement workaround engines that do not expect punctuation symbol as a command
-if (dragonfly.get_current_engine().name == 'natlink'):
-    comma = "(comma | ,)"
+if hasattr(dragonfly.get_current_engine(), "name"):
```
LexiconCode, line 19: With hasattr don't all engines have the attribute name??

LexiconCode (commented):
> 
> 
> @LexiconCode do you have insight into why the python 2.7 tests fail? I can replicate the failure locally but I'm not sure what's causing it

Yes I will look into it and let you know what I find.

castervoice/rules/core/navigation_rules/navigation_support.py
kendonB: This change allowed me to step through that bit interactively (i.e. when there is no current engine). That first bit evaluates to False if interactive and True if being run with an engine

LexiconCode:
> 
> 
> @LexiconCode do you have insight into why the python 2.7 tests fail? I can replicate the failure locally but I'm not sure what's causing it


Python 3.3+ has [Implicit Namespace Packages](https://stackoverflow.com/a/21820156/7663566) that allow it to create a packages without an __init__.py file. So Python 2 requires `__init__.py` for each directory. Adding `__init__.py` to the keyboard resolves the issue for me.

LexiconCode:
The test failure is from a change yesterday ERROR: `Invalid requirement: u"'dragonfly2[kaldi]>=0.28.0'" (from line 1 of requirements-mac-linux.txt)` unrelated to your PR.

LexiconCode:
okay a re base and this should be ready to go if you consider it complete.

LexiconCode (approved):

MarkRx:
Any thoughts on putting this in the CasterQuickReference.pdf documentation?

kendonB:
> Any thoughts on putting this in the CasterQuickReference.pdf documentation?

@MarkRx The quick reference is on its way out but we do need to work on the documentation for the keyboard grammar on the readthedocs site.

If you'd like to work on it please go for it and submit a PR.

kendonB:
I forgot we had made this:

https://caster.readthedocs.io/en/latest/readthedocs/Caster_Commands/Keyboard/

Is that good enough @MarkRx ?

kendonB:
If you want to make it easy to get there, you can add it to bringme using "website to bring me as caster keyboard docs". Then activate with "bring me caster keyboard docs".

PR #871, opened 2021-01-02
Fix Grids could be called multiple times

tripfish:
# MouseGrids could be called multiple times

## Description
window_exists expects windowname in the first place. So that the second part always returned false.

## Types of changes


LexiconCode:
@tripfish, thank you! I will test this out and get merged today

PR #887, opened 2021-04-04
Pin pyvda to 0.0.8

mrob95:
The latest release of pyvda contains a major API rework which will break the usage in caster, and also remove support for Python 2.7.

I'm not sure what the status is of Caster re python 3 but for now I would suggest pinning to [0.0.8](https://pypi.org/project/pyvda/0.0.8/), which should work indefinitely.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,5 +7,5 @@ mock>=3.0.5
 appdirs>=1.4.3
 scandir>=1.10.0
 pylint
-pyvda
+pyvda==0.0.8
 six
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,5 +6,5 @@ future>=0.18.2
 mock>=3.0.5
 appdirs>=1.4.3
 scandir>=1.10.0
-pyvda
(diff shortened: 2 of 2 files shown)


LexiconCode:
> 
> 
> The latest release of pyvda contains a major API rework which will break the usage in caster, and also remove support for Python 2.7.
> 
> I'm not sure what the status is of Caster re python 3 but for now I would suggest pinning to [0.0.8](https://pypi.org/project/pyvda/0.0.8/), which should work indefinitely.

Thank you for the heads up! Caster is Python 3 ready and many people do use it. However Natlink has yet to make an official release to Python 3. Until that time I'm stuck supporting Python 3 and 2. If Ryan were to release draconity with the configuration files for Dragon versions that would be kind to the rest the voice coding community. The community could migrate Python 3 64-bit.

mrob95:
No worries!

Draconity is open source and I think we have a working dragonfly backend for it. The difficulty is in packaging and distributing draconity plus a 32 bit DLL injector - I'm not convinced it would be easier than packaging natlink

PR #897, opened 2021-06-01
Enhanced grids linux + complete Sudoku

kendonB:
# Get grids working on linux + complete Sudoku

## Description

 - Made dragging in Sudoku work like it does in the other two grids.
 - Did what's required to get the grids to work in Ubuntu
   - Use `pyscreenshot as ImageGrab`
   - Remove `self.hide()` in Rainbow. This was causing the grid to draw on the wrong monitor in a two monitor setup. **This is what I'm most uncertain about. I don't know what function it served previously. We'll need to test on multiple systems.**
   - Moved to monitoring the process to detect if the grid Context is there **for all systems**. This is probably better than the existing way of just looking at the window title.
   - Changed the number-action mapping so that they're consistent. A simple change but it should get another set of eyes to make sure I didn't mess up.
 
## Related Issue

None.

## Motivation and Context

Mouse grids don't work on Linux.

## How Has This Been Tested

Verbally trying out the commands for Douglas/Rainbow/Sudoku on my Ubuntu machine with two monitors. I also tried Rainbow on my Windows machine but someone else should test these still work on Windows.

## Types of changes

## Checklist

- [x] Test existing functionality on Windows.

### DIFF
--- a/castervoice/rules/apps/mouse_grids/griddouglas.py
+++ b/castervoice/rules/apps/mouse_grids/griddouglas.py
@@ -16,12 +16,14 @@ def send_input(x, y, action):
     s = control.nexus().comm.get_com("grids")
     s.move_mouse(int(x), int(y))
     int_a = int(action)
-    if (int_a == 0) | (int_a == 1) | (int_a == -1):
+    if (int_a == 0) | (int_a == 1) | (int_a == 2) | (int_a == -1):
         s.kill()
         navigation.wait_for_grid_exit()
     if int_a == 0:
         Mouse("left").execute()
-    elif int_a == 1:
+    if int_a == 1:
+        Mouse("left:2").execute()
+    elif int_a == 2:
         Mouse("right").execute()
 
 
@@ -93,12 +95,9 @@ class DouglasGridRule(MappingRule):
         ShortIntegerRef("y2", 0, 300),
         Choice("action", {
             "kick": 0,
-            "psychic": 1,
-            "move": 2,
-        }),
-        Choice("point", {
-            "one": 1,
-            "two": 2,
+            "kick (double | 2)": 1,
+            "psychic": 2,
+            "move": 3,
         }),
     ]
     defaults = {
--- a/castervoice/rules/apps/mouse_grids/griddouglas.py
+++ b/castervoice/rules/apps/mouse_grids/griddouglas.py
@@ -1,4 +1,4 @@
-import time
+import time, psutil
 from dragonfly import Function, Choice, MappingRule, ShortIntegerRef
 from dragonfly.actions.mouse import get_cursor_position
 from castervoice.lib import control, navigation
@@ -104,6 +104,19 @@ class DouglasGridRule(MappingRule):
         "action": -1,
     }
 
+def is_douglas_on():
+    for proc in psutil.process_iter():
+        try:
+            # Get process name & pid from process object.
+            if proc.name().startswith("python"):
+                if len(proc.cmdline()) > 2:
+                    if proc.cmdline()[1].endswith("grids.py"):
+                        if proc.cmdline()[3] == "r":
+                            return True
+        except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
+            pass
+    return False
 
 def get_rule():
-    return DouglasGridRule, RuleDetails(name="douglas grid rule", title="douglasgrid")
+    Details = RuleDetails(name="douglas grid", function_context=is_douglas_on)
+    return DouglasGridRule, Details
--- a/castervoice/rules/apps/mouse_grids/griddouglas.py
+++ b/castervoice/rules/apps/mouse_grids/griddouglas.py
@@ -79,11 +79,11 @@ class DouglasGridRule(MappingRule):
             R(Function(send_input_select)),
         "<x1> [by] <y1> (grab | select) <x2>":
             R(Function(send_input_select_short)),
-        "squat":
+        "squat {weight=2}":
             R(Function(store_first_point)),
-        "bench":
+        "bench {weight=2}":
             R(Function(select_text)),
-        SymbolSpecs.CANCEL:
+        SymbolSpecs.CANCEL + "{weight=2}":
             R(Function(kill)),
     }
     extras = [
@@ -111,7 +111,7 @@ def is_douglas_on():
             if proc.name().startswith("python"):
                 if len(proc.cmdline()) > 2:
                     if proc.cmdline()[1].endswith("grids.py"):
-                        if proc.cmdline()[3] == "r":
+                        if proc.cmdline()[3] == "d":
                             return True
         except (psutil.NoSuchProcess, psutil.AccessDenied, psutil.ZombieProcess):
             pass
--- a/castervoice/rules/apps/mouse_grids/gridrainbow.py
+++ b/castervoice/rules/apps/mouse_grids/gridrainbow.py
@@ -1,4 +1,4 @@
-import time
+import time, psutil
 from dragonfly import Function, Choice, MappingRule, ShortIntegerRef
 from dragonfly.actions.mouse import get_cursor_position
 from castervoice.lib import control, navigation
@@ -6,7 +6,6 @@ from castervoice.lib.actions import Mouse
 from castervoice.lib.ctrl.mgr.rule_details import RuleDetails
 from castervoice.lib.merge.state.short import R
 from castervoice.rules.ccr.standard import SymbolSpecs
-
 
 def kill():
     control.nexus().comm.get_com("grids").kill()
@@ -16,13 +15,15 @@ def send_input(pre, color, n, action):
     s = control.nexus().comm.get_com("grids")
     s.move_mouse(int(pre), int(color), int(n))
     int_a = int(action)
-    if (int_a == 0) | (int_a == 1) | (int_a == -1):
+    if (int_a == 0) | (int_a == 1) | (int_a == 2) | (int_a == -1):
         s.kill()
         navigation.wait_for_grid_exit()
         time.sleep(0.1)
     if int_a == 0:
         Mouse("left").execute()
-    elif int_a == 1:
+    if int_a == 1:
+        Mouse("left:2").execute()
+    elif int_a == 2:
         Mouse("right").execute()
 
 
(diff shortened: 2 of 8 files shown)


kendonB:
@BenChand would you like to review this PR?

LexiconCode:
Great minds think alike as I was looking at making the grid use function context last night.
Something like the following but I think achieves the same goal without another dependency. Probably could be made a little cleaner wrapping it up in a class.

```python
import six
import subprocess
import time
from dragonfly import get_engine, monitors, FuncContext

from castervoice.lib import control, settings

GRID_PROCESS = None
MODE = None

def get_grid_mode_context(grid):
    return FuncContext(lambda: get_grid_mode(grid))

def get_grid_mode(grid):
    global MODE
    print("get_grid_mode")
    if MODE is None:
        return False
    elif MODE and grid == "legion":
        return True    
    elif MODE and grid == "rainbow":
        return True 
    elif MODE and grid == "sudoku":
        return True
    elif MODE and grid == "douglas":
        return True

def mouse_alternates(mode, monitor=1, rough=True):
    global GRID_PROCESS, MODE
    args = []

    if mode == "legion":
        from castervoice.asynch.mouse.legion import LegionScanner
        r = monitors[int(monitor) - 1].rectangle
        bbox = [
            int(r.x),
            int(r.y),
            int(r.x) + int(r.dx) - 1,
            int(r.y) + int(r.dy) - 1
        ]
        ls = LegionScanner()
        ls.scan(bbox, rough)
        tscan = ls.get_update()
        args = [
            settings.settings(["paths", "PYTHONW"]),
            settings.settings(["paths", "LEGION_PATH"]), "-t", tscan[0], "-m",
            str(monitor)
        ]
    elif mode == "rainbow":
        args = [
            settings.settings(["paths", "PYTHONW"]),
            settings.settings(["paths", "RAINBOW_PATH"]), "-g", "r", "-m",
            str(monitor)
        ]
    elif mode == "douglas":
        args = [
            settings.settings(["paths", "PYTHONW"]),
            settings.settings(["paths", "DOUGLAS_PATH"]), "-g", "d", "-m",
            str(monitor)
        ]
    elif mode == "sudoku":
        args = [
            settings.settings(["paths", "PYTHONW"]),
            settings.settings(["paths", "SUDOKU_PATH"]), "-g", "s", "-m",
            str(monitor)
        ]
    MODE=mode
    GRID_PROCESS = subprocess.Popen(args) if args else None
    get_engine().process_grammars_context()


def wait_for_grid_exit(timeout=5):
    global GRID_PROCESS, MODE
    if GRID_PROCESS:
        # TODO Remove if-part after fully migrating to Python3
        if six.PY2:
            t = 0.0
            inc = 0.1
            while t < timeout:
                GRID_PROCESS.poll()
                if GRID_PROCESS.returncode is not None:
                    break
                t += inc
                time.sleep(inc)
            if t >= timeout:
                GRID_PROCESS.kill()
        else:
            try:
                GRID_PROCESS.wait(timeout)
            except subprocess.TimeoutExpired:  # pylint: disable=no-member
                GRID_PROCESS.kill()
    MODE = None
    GRID_PROCESS = None

def kill():
    global GRID_PROCESS, MODE
    control.nexus().comm.get_com("grids").kill()
    MODE = None
    GRID_PROCESS = None
```

castervoice/rules/apps/mouse_grids/gridrainbow.py
```diff
@@ -138,5 +136,19 @@ class RainbowGridRule(MappingRule):
     }
 
 
+def is_rainbow_on():
```
LexiconCode, line 139: This is checked every utterance so we have to be careful of for performance reasons. In the event that it takes longer to process the utterance it delays recognition.

LexiconCode:
There's an issue with how we're checking for the grid with BenChand addition. Basically it doesn't account for canceling the grid and it thinks grid system still running.

LexiconCode:
`self.hide()` is meant to hide the overlay causing the window to become invisible until it's needed to be redrawn. Worst case scenario if this is needed on others OS we can put it behind the platform check.

LexiconCode:
Would it make sense to separate the PR into 2 2 separate pull requests. One for function contacts and one for Linux support?

castervoice/rules/apps/mouse_grids/gridrainbow.py
kendonB: mm good point - the question is whether it takes longer than checking the window name though right? Though I presume it's still going to be slower. We could also make this linux only

LexiconCode: It depends on a number of factors. As I was working on this a couple nights ago here's how I worked out the issue hopefully in a cross-platform wat.

I eliminated the performance concerns completely by unifying the state management with the grids in `navigation`. a few things were needed:

- one function context function to call for all grids. ` function_context=lambda: navigation.is_grid_active("legion")` using a lambda I was able to pass a parameter.

-  re-factored `def kill():` from the individual grids since to  `navigation`

It's important that `mouse_alternates` and the methods that kill the grid are the same file. Principally when `mouse_alternates` stores the process `GRID_PROCESS` and grid mode `MODE` they are stored as global states. if the grid is asked to kill or directly canceled `Kill` is issued with following `wait_for_grid_exit` which polls the process checking to make sure it's terminated. When the process is successfully terminated both `GRID_PROCESS` and `MODE` are set to `None` updating global variables.

By keeping track of the grid process and the mode that requested `GRID_PROCESS` and `MODE` we can use function context to call `is_grid_active` there's no performance penalty here.

`
def is_grid_active(grid):
    global MODE
    if MODE is None:
        return False
    elif MODE == grid:
        return True 
`

[is_grid_active](https://github.com/LexiconCode/Caster/blob/8b2af62fdbc961a425a426c681ef0cec2d8bda6a/castervoice/lib/navigation.py#L27)

kendonB:
I support your approach for detecting the grid context. It is cleaner than mine. I haven't noticed any performance difference between the two branches though - on ubuntu.

@LexiconCode If you make a PR let's merge yours first then I'll add my linux and sudoku changes.

kendonB:
need to remember to add to requirements as well so the checks pass

castervoice/rules/apps/mouse_grids/griddouglas.py
```diff
@@ -1,4 +1,4 @@
-import time
+import time, psutil
```
LexiconCode, line 1: Once `psutil` is removed the Python 3 tests will pass.

castervoice/rules/apps/mouse_grids/gridrainbow.py
```diff
@@ -1,4 +1,4 @@
-import time
+import time, psutil
```
LexiconCode, line 1: Same as above with `psutil`

LexiconCode:
What's your rationale for `pyscreenshot` over `PIL ImageGrab`? They both should be cross-platform however I definitely could be missing something.

LexiconCode:
>     * Remove `self.hide()` in Rainbow. This was causing the grid to draw on the wrong monitor in a two monitor setup. **This is what I'm most uncertain about. I don't know what function it served previously. We'll need to test on multiple systems.**

With Windows the grid appears blank until the rainbow grid is rendered. `self.hide()` hides the blank grid until it's rendered and then the grid appears.

kendonB:
> What's your rationale for `pyscreenshot` over `PIL ImageGrab`? They both should be cross-platform however I definitely could be missing something.

ImageGrab is not cross platform: https://stackoverflow.com/a/43636492/2726309

LexiconCode:
@kendonB The only reason I asked is because the docs make reference to Mac and Linux [ImageGrab](https://pillow.readthedocs.io/en/stable/reference/ImageGrab.html). However we'll go with pyscreenshot.

kendonB:
Hmm the stackoverflow may be outdated. Though I definitely only would have switched if I was getting an error with PIL. I will have another look later

kendonB:
@LexiconCode my pillow was too old - I have reverted that bit and updated reqs to be pillow>=8.0.0

kendonB:
recommend squash and merge for this one as well :)

LexiconCode:
@kendonB 

did you add the conditional check for windows 'self.hide()'? Does that need to be for each print system as well? This PR is looking great thank you!

It'll be a day or so before I can do the final check also testing on Python 2.7.

LexiconCode:
Depending on how you tested I bet the issue isn't `hide` but `unhide`.

These different ways to set focus came about because Windows is notoriously unreliable about ensuring focus of the window. I would be curious if you commented out derivatives of focus and see if the behavior is the same. I thought about taking this out completely as we no longer rely on focus but we do need to ensure the windows in the foreground.

```
    def unhide(self):
        ''''''
        self.deiconify()
        self.lift()
        time.sleep(0.1)
        self.focus_force()
        self.focus_set()
        self.focus()
```

kendonB:
doing this 
```
    def unhide(self):
        ''''''
        self.deiconify()
        self.lift()
        time.sleep(0.1)
        #self.focus_force()
        #self.focus_set()
        #self.focus()
```
but leaving `self.hide()` in also incorrectly prints the rainbow grid on the first monitor.

But if we're not actually using hide or unhide for anything then let's just get rid of it here in this PR I say.

kendonB:
Part of the function of hide/unhide on rainbow might be because it draws the background image inside the PIL window. I don't know why it's set up this way because the PIL image is transparent. So you see the actual background image as well as the old captured one. You see both if something changed

LexiconCode:
> 
> 
> Part of the function of hide/unhide on rainbow might be because it draws the background image inside the PIL window. I don't know why it's set up this way because the PIL image is transparent. So you see the actual background image as well as the old captured one. You see both if something changed

Yes I am seeing a difference on Windows. Basically a blank gray window appears the size of the monitor before the rest of the window is drawn. `hide` successfully hides that until it's ready to be drawn. 

Not a big issue it's something we can do a platform check to execute on other platforms and not Linux. I can push that to the PR as I have it done and removed the focus bits as are not needed anymore.

LexiconCode:
> 
> 
> doing this
> 
> ```
>     def unhide(self):
>         ''''''
>         self.deiconify()
>         self.lift()
>         time.sleep(0.1)
>         #self.focus_force()
>         #self.focus_set()
>         #self.focus()
> ```
> 
> but leaving `self.hide()` in also incorrectly prints the rainbow grid on the first monitor.
> 
> But if we're not actually using hide or unhide for anything then let's just get rid of it here in this PR I say.

Yes I am seeing a difference on Windows. Basically a blank gray window appears the size of the monitor before the rest of the window is drawn. `hide` successfully hides that until it's ready to be drawn. 

Not a big issue it's something we can do a platform check to execute on other platforms and not Linux. I can push that to the PR as I have it done and removed the focus bits as are not needed anymore.

LexiconCode:
Tests out on Python 2.7.

PR #916, opened 2022-01-20
Add environment variable expansion into path reading

kendonB:
# Add environment variable expansion into path reading

## Description

Changing a few instances where paths from `settings.toml` get read without expanding environment variables. 

## Related Issue
None.

## Motivation and Context

This allows users to reference environment variables in settings.toml for some paths. I wanted to use this to share a settings folder between two machines with different file systems via dropbox. 

## How Has This Been Tested

Ran this on my Ubuntu system and it started up fine.

## Types of changes

## Checklist


kendonB:
This seems to break user rules  - @LexiconCode can you think of why that might be?

kendonB:
Ok looks like my original approach wasn't right - simpler just to edit `settings.settings` as it gets read in. Importantly, after it rewrites any new defaults onto disk.

castervoice/asynch/hmc_rules/hmc_launch_rule.py
```diff
@@ -9,7 +9,7 @@
 
 def receive_settings(data):
     settings.SETTINGS = data
-    settings.save_config()
+    settings.save_config(paths = True)
```
kendonB, line 12: This one keeps the code the same for hmc. hmc is not really used as far as I know so this is just a conservative approach

castervoice/lib/settings.py
```diff
@@ -449,11 +451,31 @@ def settings(key_path, default_value=None):
...
+            printer.out("\n\n {} while loading settings file: {} \nAttempting to recover...\n\n".format(repr(e), _SETTINGS_PATH))
+        SETTINGS_tmp = copy.deepcopy(SETTINGS)
+        if "paths" in result:
+            SETTINGS_tmp['paths'] = result['paths']
+        _save(SETTINGS_tmp, _SETTINGS_PATH)
+    else:
+        _save(SETTINGS, _SETTINGS_PATH)
+
+     
+
```
kendonB, line 478: Right now, paths only get written onto settings.toml if they are brand new defaults. 

I think that probably we don't have a use case for overwriting settings paths from within caster? So we could remove the if statement here and always run the first chunk of code?

kendonB:
@LexiconCode I believe this one is almost ready. 

I have a couple of comments in the code.

Recommend squash and merge since the commits go back and forth.

The current approach reads in all paths from settings.toml expanding env variables. Then, whenever it saves, it rereads the paths from disk without expanding. This is fine because caster never overwrites paths from within caster.

The thing I'm not sure of is if we ever want to overwrite paths with what is in memory (the current approach); probably not because that would cause it to overwrite expanded environment variables which would be a bug. Any future overwrite code should be bespoke, read in from disk, edit, then write. So I recommend we just get the code to always read paths from disk before saving using `save_config`.

kendonB:
> Can you rebase on DropPy2?

Done now @LexiconCode

kendonB:
@comodoro are you able to review this PR?

comodoro:
> @comodoro are you able to review this PR?

I am not sure, because I don't know the purpose of the code. Why can env variable expansion not happen only once loading settings?

PR #965, opened 2026-05-16
fix: add retry logic for HUD startup to handle race conditions

### LINKED ISSUE #920: HUD doesn't load on computer startup.
**Describe the bug**
When automatically running Caster on computer startup. The HUD fails to connect.
It's most likely, because the HUD's GUI doesn't load quickly enough on startup. Manually loading it a bit after startup loads it up fine. As does closing Caster and the HUD and restarting.

**To Reproduce**
1. Make Caster run on startup (on windows one way to do that, is adding a link to Run_Caster_Kaldi.bat into %AppData%\Microsoft\Windows\Start Menu\Programs\Startup)
2. Boot / Reboot computer
3. Get error 

>Hud not available.
[WinError 10061] No connection could be made because the target machine actively refused it

**Expected behavior**
HUD loads on startup without having to restart caster or delaying starting Caster.

**System:**
 - OS: Windows 10
 - Caster Version: DropPy2 branch, 06ec0ae commit

amirf147:
# Fix HUD connection race condition on cold boot

## Description

I've updated `HudPrintMessageHandler` to include a retry loop when establishing the initial connection to the HUD process. Instead of failing instantly on the first ping, the handler will now poll for a connection up to 10 times with a 0.5-second delay. The hope is that this gives the OS enough time to spin up the background process and bind the port before Caster gives up and flags the HUD as inactive.

## Related Issue

Fixes #920

## Motivation and Context

After running into this issue and discussing it with an LLM, I suspect this might be a race condition occurring on a cold startup. It seems like when running Caster for the first time after a boot, `start_hud` launches the HUD via `subprocess.Popen`, but the main Caster thread continues executing and attempts to ping the HUD before the OS has fully opened the port. 

This appears to result in the `[WinError 10061] Connection Refused` error, leaving a blank white HUD window. Restarting Caster usually bypasses the error, likely due to OS caching making the process load much faster the second time around. This fix is an attempt to ensure Caster gracefully waits for the HUD to wake up before throwing the error.

hud-error

## How Has This Been Tested

Tested by observing Caster's initial startup immediately following a full system reboot. It appears that the new polling loop successfully catches the delay, waits for the port to open, and connects to the HUD GUI, avoiding the instant `[WinError 10061]` failure I was seeing previously.

## Types of changes

## Checklist

### DIFF
--- a/castervoice/asynch/hud_support.py
+++ b/castervoice/asynch/hud_support.py
@@ -1,4 +1,4 @@
-import sys, subprocess, json
+import sys, subprocess, json, time
 
 from dragonfly import CompoundRule, MappingRule, get_current_engine, Function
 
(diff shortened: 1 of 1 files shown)


LexiconCode (approved):
Thank you. I really appreciate these pull requests! Every bit helps.

